PR #149, opened 2023-06-02
[UX] settings close floating button, fix tokens position auto suggest prompt, AutoSuggest minor bugs

ljh5650:
# PR Checklist
- [ - ] Did you check if it works normally in all models? *ignore this when it dosen't uses models*
- [ O ] Did you check if it works normally in all of web, local and node hosted versions? if it dosen't, did you blocked it in those versions?
- [ O ] Did you added a type def?

# Description
There are many users who can't find the character deletion, so it has been changed so that it is clearly visible in the header.
스크린샷 2023-06-03 004117
스크린샷 2023-06-03 004122

It is inconvenient to scroll up and close the long settings window when scrolling, so it is fixed at the top.
스크린샷 2023-06-03 002151

Modify tokens count position not moved together auto suggest prompt.
스크린샷 2023-06-03 002350

Fixed the problem that auto-suggestion does not work for firstMessage
Fixed the problem that auto-suggestion is not saved

### DIFF
--- a/src/lib/ChatScreens/Suggestion.svelte
+++ b/src/lib/ChatScreens/Suggestion.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-    import { requestChatData } from "src/ts/process/request";
+	import { requestChatData } from "src/ts/process/request";
     import { doingChat, type OpenAIChat } from "../../ts/process/index";
-    import { DataBase, type character } from "../../ts/storage/database";
+    import { DataBase, setDatabase, type character, type Message, type groupChat, type Database } from "../../ts/storage/database";
     import { selectedCharID } from "../../ts/stores";
     import { translate } from "src/ts/translator/translator";
     import { CopyIcon, LanguagesIcon, RefreshCcwIcon } from "lucide-svelte";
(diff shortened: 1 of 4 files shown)


kwaroran:
Even if it's hard to find the delete button, I don't think it's good to add it to the header,
There are already too many buttons in the header, and it becomes more complicated.

Alwaysfindtheway:
This has caused quite a few people to become confused and unable to find the Delete Character button.

ljh5650:
> Could you revert the deletion menu?
> 
> Even if it's hard to find the delete button, I don't think it's good to add it to the header, There are already too many buttons in the header, and it becomes more complicated.

@kwaroran 
Deletion ui separation commit reverted.

How about improving the Edit Order feature with the ability to display a deleteable x button in the sidebar?
It seems easier to find than at the bottom of the Advanced Settings.
스크린샷 2023-06-03 172219

drPpZero:
> > Could you revert the deletion menu?
> > Even if it's hard to find the delete button, I don't think it's good to add it to the header, There are already too many buttons in the header, and it becomes more complicated.
> 
> @kwaroran Deletion ui separation commit reverted.
> 
> How about improving the Edit Order feature with the ability to display a deleteable x button in the sidebar? It seems easier to find than at the bottom of the Advanced Settings.

I don't think this is a good idea. The delete button should be in the character's sidebar.

PR #227, opened 2023-08-04
modify chat layout

mauar82:
My changes do not use models.
I tested it by hosting on desktop and mobile web, but I couldn’t test it locally because I don’t know how to build.
Type definition has not been added.

Developed to reduce space waste on mobile. Although the existing layout is more suitable for the essence of chat, it is thought that saving space is more important because a large number of messages are received in practice. However, I leave the decision to apply to you.

### DIFF
--- a/src/lib/ChatScreens/Chat.svelte
+++ b/src/lib/ChatScreens/Chat.svelte
@@ -75,17 +75,17 @@
     $: displaya(message)
 </script>
 <div class="flex max-w-full justify-center risu-chat" class:bgc={isLastMemory}>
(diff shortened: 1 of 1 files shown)


kwaroran:
Thanks for the PR, However, since Risu is for multi-platform, and has a lot of desktop and tablet users,I think the current UI suits more.

mauar82:
> Thanks for the PR, However, since Risu is for multi-platform, and has a lot of desktop and tablet users,I think the current UI suits more.



> Thanks for the PR, However, since Risu is for multi-platform, and has a lot of desktop and tablet users,I think the current UI suits more.

I think so, too. but how about adding it as a display option?

PR #234, opened 2023-08-19
[feat] apply request number of chat.

Yanggum:
# PR Checklist

# Description
Reduced context consumption by allowing you to limit the number of chats from last to send requests to. We modified the program to send a request for all chat history if it's 0, but a specific number of chats if it's 1 or higher.


kwaroran:
This can already be archived by prompt template feature.

PR #245, opened 2023-09-10
Some illegal patches

SeA-luby:
# PR Checklist

# Description
Surprisingly, it does not satisfy all 3 of the PR checklists.

This is an illegally modified version of node that I was personally using. I had to look at your entire code very carefully and the more I dug, the more I found it to be extremely complex. Unbelievable for a one-man development. I applaud you for that.

Anyway, this version is very badly programmatically designed because I failed to fully grasp its very complex function flow, and none of these files are intended for actual pulls. I heard about your recent delay in updating due to health issues, and since my fixes at least clearly point to the files with the functions that need to be fixed anyway, I thought this might help you a little bit, so I created this request.

However, it seems I'm a step behind, as the thought of this request was last Friday and you seem to have already covered most of the issues in the meantime XD It's not even compatible with the current version, as it was based on the old v1.50.1. So if you compare v1.50.1 to this branch, you can clearly see the 5 small but painful fixes I made.

Still, I hope this lets you know that there's someone watching and cheering you on and keeping an eye on this project.

### DIFF
--- a/server.bat
+++ b/server.bat
@@ -1,3 +1,4 @@
 call npm install
 call npm run build
 call npm run runserver
+pause
--- a/server/node/server.cjs
+++ b/server/node/server.cjs
@@ -41,8 +41,51 @@ app.get('/', async (req, res, next) => {
 })
 
 const proxyFunc = async (req, res, next) => {
+<<<<<<< Updated upstream
+    try{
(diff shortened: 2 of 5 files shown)


SeA-luby:
I just checked out the new patch and while the patch notes say there are a lot of improvements around SupaMemory and HypaMemory, the auxiliary model will still try to force a SupaMemory load and throw an error regardless of the HypaMemory toggle.

PR #287, opened 2024-01-26
Add persona portraits

bangonicdd2:
# PR Checklist

# Description
image
image


kwaroran:
This is a good commit, but I think it would be better to have it apply to each persona rather than modifying $DataBase.

bangonicdd2:
Understood, I'll reopen this after fix them.

PR #304, opened 2024-03-08
Add claude-3 stop_sequences

bangonicdd:
# PR Checklist

# Description

I would suggest adding stop sequences as the community is constantly reporting cases of direct user impersonation.


kwaroran:
This may solve some impersonation problems, but it may cause unexpected behavior in some cases; instead, it would be better to include the stop sequence customization feature to Claude-3 that currently exists for other models in RisuAI.

PR #360, opened 2024-04-11
Add exception on logit_bias for gpt-4-turbo 

IHaBiS02:
# PR Checklist

# Description
As gpt-4-turbo model can both receive logit_bias and inlay image, I make some exceptions from removing bias on the request.  
I check it run well with proxy, but not tested with API key, so testing with official API is needed before merge.


justpain02:
@kwaroran can you check this code? I think it is necessary to use bias on gpt4 turbo 0409.

PR #391, opened 2024-04-27
Add support for custom embedding URL in Playground

sub-hub:
# PR Checklist

# Description

Add support Openai-compatible custom embedding server on the playground.

Openai-compatible embedding servers:
https://github.com/toshsan/embedding-server
https://github.com/limcheekin/open-text-embeddings
https://github.com/michaelfeil/infinity

I only tested it with infinity. 
there is also a feature that If url does not end in /embeddings, automatically adds /embeddings.
I think I need to test it with more local embedding servers, but I think it'll be okay since it's only Playground

### DIFF
--- a/src/lib/Playground/PlaygroundEmbedding.svelte
+++ b/src/lib/Playground/PlaygroundEmbedding.svelte
@@ -8,6 +8,7 @@
 
     let query = "";
     let model = "MiniLM";
+    let customEmbeddingUrl = "";
     let data:string[] = [];
     let dataresult:[string, number][] = [];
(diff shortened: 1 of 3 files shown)


sub-hub:
Please look at the code about feature add /embeddings in url,  I think it's little wrong but I can't fix it now because I'm outside

sub-hub:
Add a new function `appendLastPath` that can be used by features such as `autofillRequestUrl`.
maybe also can be used in the `request.ts`, but I didn't change `request.ts` because it goes far from the purpose of pr

PR #405, opened 2024-05-05
Add: New Option Combine Translation

sub-hub:
# PR Checklist

# Description
Basically, this feature was created to combine a sentence together and translate it when a display edit script splits one sentence into multiple HTML tags.
I hope you confirm as this option will significantly improve translation performance without modifying existing scripts! `(e.g. Automark)`
I also made `translateHTML` accept a `chatID` and optimized the existing code to get the script from `charArg`.


sub-hub:
I've debugged it in a lot of situations, but I still think it needs a code review from someone else as this option manipulates HTML a lot. 
Please Look especially at the translateNodeText and translateHTML functions

kwaroran:
Since we cannot make code for all cases, I think this is good enough 👍

PR #411, opened 2024-05-07
Fix: handling more markdown in getNodetextToSentence

sub-hub:
# PR Checklist

# Description
Sorry but there were some markdown tags that I missed, But I think It's not an urgent patch like last time.
So If there are any tags I missed, I will continue commit to this pr until the next update.


sub-hub:
Maybe the markdown is working properly now.

PR #413, opened 2024-05-10
Add new option  Exclude non-existent assets

sub-hub:
# PR Checklist

# Description
I created a new option to remove non-existent assets from the editprocess step.
> If enabled, excludes assets that do not exist on the character when sending chat to the AI.
In the long chat, This will reduces the possibility of the AI mentioning assets that don't exist on the character.

I made the default value to `true` because I think it's a code with no side effects.
But still, please look at the code.


kwaroran:
Since it can already be implemented as a module, I wonder if there is a need to add it as a feature.

sub-hub:
My feature dynamically gets the list of assets a character has, so that neither the character creator nor the user has to do anything extra like regular expressions.
But Is it possible to create a feature like this with module? I know there are different types of module, but isn't it hard to get a character's assets and use them for operations like this?
(I realized `{{assetlist}}` existed just before I created this feature, but even if it did, I didn't think it would be easy to implement as a module)

It would be cool if you could show how you would implement this feature in a RisuAI module.
I've always thought modules are an interesting feature, but I haven't seen how official developers use them.

sub-hub:
If you're not going to merge pr, I'd like you to upload the module you made. Because this was a feature I need.
And I requested it from the main branch of my repo, so this has to close to do something else.
Or please merge this pr. There will be no bugs, I don't think there will be any problems.

PR #451, opened 2024-05-27
Add new CBS CotToggled

sub-hub:
# PR Checklist

# Description


kwaroran:
Now that the custom toggle has been added, there are no plans to add those.

PR #476, opened 2024-06-01
Fix Local Backup Loading Failure Due to Incorrect Backup Data Saving.

asm6788:
# PR Checklist

# Description
Fix https://github.com/kwaroran/RisuAI/issues/475


kwaroran:
👍

PR #572, opened 2024-07-15
Fix lua lowlevel access key not added

sunho:
Seems like a coding mistake here. Even if low level access is turned on, another random access key was added making low level access feature not working.

### DIFF
--- a/src/ts/process/lua.ts
+++ b/src/ts/process/lua.ts
@@ -389,7 +389,7 @@ export async function runLua(code:string, arg:{
     else{
         LuaSafeIds.add(accessKey)
         if(lowLevelAccess){
-            LuaLowLevelIds.add(v4())
+            LuaLowLevelIds.add(accessKey)
         }
     }
     let res:any


sunho:
Also noticed low level access is just turned off for lua trigger in general. Is this intended?

sunho:
Actually, it's turned off on runLua argument too. Is this intentional?

kwaroran:
The Lua accessId part is a mistake, but others are intentional.

sunho:
Oh so I guess low level access for lua is not allowed right now?

sunho:
There are so many cool applications with these things. (one thing i'm working on is fixing wrong asset image using vector embedding similarity search from hypa processor) We should allow it while warning users of its danger properly.

kwaroran:
No, it has not been completely not allowed. only the edit part is blocked, the rest works.

sunho:
Ah that make sense. I'll remove the change in the other part.

sunho:
I combined everything into https://github.com/kwaroran/RisuAI/pull/575

PR #575, opened 2024-07-15
Fix various lua low level access bugs

sunho:
### 1 Access key was not added properly
First commit fixes this 

### 2 Lua engine was getting killed when it waits for async function
runLua function gets run in two different paths (i.e. runLua for lua edit triggers and another one for onOuput handling) However, when one of invocations waits for another promise from js side (i.e. low level function) to finish, it gives chance for next trigger mode to run which would call `if(luaEngine){ luaEngine.global.close() }` that will close lua engine for the invocation that was waiting for promise which is problematic. 

Second commit fixes this by having a pool of engine for each trigger mode having single mutex that prevents running same engine at the same time while allowing different tigger modes to run parallely

### 3 Break statements were missing making every trigger to run for input event.
This was causing insane lagging and buggy behaviours. Third commit fixes this.

### DIFF
--- a/src/ts/mutex.ts
+++ b/src/ts/mutex.ts
@@ -1,0 +1,88 @@
+/**
+ * A lock for synchronizing async operations.
+ * Use this to protect a critical section
+ * from getting modified by multiple async operations
+ * at the same time.
+ */
+export class Mutex {
+    /**
+     * When multiple operations attempt to acquire the lock,
+     * this queue remembers the order of operations.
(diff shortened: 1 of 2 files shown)


sunho:
Actually it might make sense to have a pool of lua engines. Let me think more.

sunho:
I guess we can just use this PR

sunho:
This is ready to be merged now.

PR #576, opened 2024-07-15
fix: finer granularity cache key for dynamic assets

sunho:
Asset list can be different depending on character. Technically, we should also invalidate cache when asset list is changed but I doubt people will hit this case often. (bot maker somewhat hit it though so maybe worth taking care of)


sunho:
Or even better make it default enabled :) It runs pretty fast on my computer.

PR #598, opened 2024-08-22
Add Show Masked Character option

sub-hub:
# PR Checklist

# Description


kwaroran:
This implementation exposes not only API keys, but also that need to be masked at all time (like passwords). A commit that also fixes this has been commited at 628b17510d038902a1a792c3ee9792bc20d65e93.

sub-hub:
I've seen all the code where hideText is used, and I think the hideText in BotSettings.svelte can be turned off.
I don't see any input that is always hideText in your commits either, can you tell me what you mean by need to be masked at all time?
Is there some code that is not visible to other developers but you?

PR #635, opened 2024-09-28
Add support for character-specific translation prompts with Ax. model

poroyo:
# PR Checklist
- [v] Did you check if it works normally in all models? *ignore this when it dosen't uses models*
- [v] Did you check if it works normally in all of web, local and node hosted versions? if it dosen't, did you blocked it in those versions?
- [v] Did you added a type def?

# Description
I have added the ability to insert a character-specific prompt when using the Ax. model. This feature is called **Translator’s Note** and can be found in the advanced settings of each character section. To enable it, simply include `{{slot::tnote}}` in the translation prompt within the language settings. Due to certain limitations, this feature is not available in group chats.

This feature allows for providing separate translation glossaries for each character and requesting additional translations for narrative tone, character-specific titles, and forms of address.


kwaroran:
Won't this hurt the UX for newcomers? I think the basic character config UI should not become any more complicated.

PR #655, opened 2024-11-01
Log Generator Pro 1.1 update

dootaang:
I'm sending this because version 1.1 update is out. The main points are a few fixes for text issues and enhancements for convenience. Just check the log_generator_pro.py file.


kwaroran:
I will transplant it separately. Thank you.

PR #680, opened 2024-12-09
HypaV2 context deletion safety

LightningHyperBlaze45654:
# PR Checklist
- [] Did you check if it works normally in all of web, local and node hosted versions? if it dosen't, did you blocked it in those versions? << Checked on browser only, works fine

# Description
HypaV2 data's large type definition update.
```ts
    mainChunks: { // summary itself
        id: number;
        text: string;
        chatMemos: Set<string>; // UUIDs of summarized chats
        lastChatMemo: string;
    }[];
    chunks: { // split mainChunks for retrieval or something. Although quite uncomfortable logic, so maybe I will delete it soon.
        mainChunkID: number;
        text:string;
    }[];
```
With this, ensure that mainChunks is relevant on chat context change by deletion

If there is no UUID in the chat context, but it exists on chatMemos on certain mainChunk, removes it.

Changed index.svelte.ts to update args on each call to ensure hypav2 to stay adaptive on this change without refreshing the page

Also changed mainChunks to be pushed instead of unshifted


kwaroran:
Is this compatible with existing saves? or does it have a conversion process?

LightningHyperBlaze45654:
> Is this compatible with existing saves? or does it have a conversion process?

Since there is a data change, existing chat's hypav2 would be redundant(invalid)

It requires conversion, however I do not think it will be achieved easily.

LightningHyperBlaze45654:
Current interface of HypaV2 looks like this:
```
export interface HypaV2Data {
    chunks: {
        text: string;
        targetId: string;
    }[];
    mainChunks: {
        text: string;
        targetId: string;
    }[];
}
```
And considering the targetId is the last chat's memo of summary, conversion function can be made.

kwaroran:
It will need conversion or separation because it might break existing chats, which leads to a bad UX  in my opinion.

LightningHyperBlaze45654:
I'll revise the code to check if the type is previous one. And convert it to newly upgraded version.

If memo does not exist, it will also sanitize itself

LightningHyperBlaze45654:
Re-opened the brach as conversion function has been added

It works on the init of HypaV2

LightningHyperBlaze45654:
@kwaroran does it check out? Is it okay?

PR #688, opened 2024-12-23
Use elyn embedding

StableFluffy:
# PR Checklist

# Description


StableFluffy:
I mistakenly pressed the submit PR button. Sorry for the inconvenience.

PR #697, opened 2024-12-28
Add Custom Character Folder Image Feature Using Global Module Assets

poroyo:
# PR Checklist

# Description
This PR adds a new feature that allows users to add custom images to character folders. Instead of uploading images directly, this feature utilizes the existing image pool from **the global module**. Users can either select from existing images in the global module or upload new images to the module.

The users can use a context menu on a folder to update its image.

## Why Use the Global Module?

The global module was chosen for several reasons:  
1. It allows the reuse of existing functions and code, minimizing the need for new implementations.  
2. Users can easily select existing images without re-uploading them.  
3. The global module provides reliable support for image upload and deletion.  
4. Images uploaded via the global module are accessible across all platforms where the shared data is available.

## Note

To retrieve the list of assets from the global module, I used the following code:

```typescript
let assetPaths: { [key: string]: { path: string } } = {};

const moduleAssets = getModuleAssets();
if (moduleAssets.length > 0) {
  for (const asset of moduleAssets) {
    const assetPath = await getFileSrc(asset[1]);
    assetPaths[asset[0].toLocaleLowerCase()] = {
      path: assetPath,
    };
  }
}
```

I reused the `parseAdditionalAssets` code from `/src/ts/parser.svelte.ts` because it seemed simple and practical for this feature. I thought about refactoring it into a shared function but decided to keep it simple for now. I’m not sure if this is the best approach, but it can always be changed later if needed.

---

If you are already working on a similar feature, find any issues with this code, or feel this PR does not align with the project's direction, I fully understand if this PR is not accepted.

Thank you!

---

One last note: With *transparent background images* and the existing folder color functionality, users can differentiate categories even when using the same image by applying different colors!

### DIFF
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -490,6 +490,7 @@ export const languageEnglish = {
     cancel: "Cancel",
     renameFolder: "Rename Folder",
     changeFolderColor: "Change Folder Color",
+    changeFolderImage: "Change Folder Image",
     fullWordMatching: "Full Word Matching",
     botSettingAtStart: "Bot Menu when Launch",
     triggerStart: "On chat Send",
--- a/src/lang/ko.ts
+++ b/src/lang/ko.ts
@@ -441,6 +441,7 @@ export const languageKorean = {
     "cancel": "취소",
     "renameFolder": "폴더 이름 변경하기",
     "changeFolderColor": "폴더 색상 변경하기",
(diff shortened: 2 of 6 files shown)


kwaroran:
In my opinion, this is too unintuitive and goes off the module's use cases. Uploading and saving separately instead of in modules would be better.

PR #701, opened 2025-01-02
fix: HypaV2 serialization and token calculation

Bo26fhmC5M:
# PR Checklist

# Description
This PR addresses two issues in HypaV2:

1. Fixes JSON serialization of chatMemos in HypaV2Data by converting Set to string array
2. Corrects token calculation by subtracting tokens from previously summarized chats

Changes:
- Added SerializableHypaV2Data interface with string[] instead of Set<string>
- Added conversion functions between HypaV2Data and SerializableHypaV2Data
- Fixed token calculation by properly accounting for removed chats

Note:
I'm not entirely confident about the token calculation changes. A thorough review would be appreciated, particularly regarding:
- Whether token subtraction for removed chats is the correct approach
- Potential edge cases in the current implementation 

Please review these changes, especially the token calculation logic, to ensure it aligns with the intended behavior of the memory system.

### DIFF
--- a/src/ts/process/memory/hypav2.ts
+++ b/src/ts/process/memory/hypav2.ts
@@ -381,6 +381,12 @@ export async function hypaMemoryV2(
         const lastChatIndex = chats.findIndex(chat => chat.memo === lastChatMemo);
         if (lastChatIndex !== -1) {
             idx = lastChatIndex + 1;
+
+            // Subtract tokens of removed chats
+            const removedChats = chats.slice(0, lastChatIndex + 1);
+            for (const chat of removedChats) {
+                currentTokens -= await tokenizer.tokenizeChat(chat);
+            }
         }
     }
     // Starting chat index of new mainChunk to be generated


LightningHyperBlaze45654:
```ts
const fullResult = `<Past Events Summary>${mainPrompt}</Past Events Summary>\n<Past Events Details>${chunkResultPrompts}</Past Events Details>`;
    const fullResultTokens = await tokenizer.tokenizeChat({
        role: "system",
        content: fullResult,
    });
    currentTokens += fullResultTokens;
```
remove this part as this is reduandant and later on create similar issues over chat

src/ts/process/memory/hypav2.ts
LightningHyperBlaze45654, line 1: Seems required when HypaMemory is already called, and the entire page is refreshed. This case, entire chat will be counted even though it does not goes into the context.

Reference on index.svelte.ts.

LightningHyperBlaze45654 (approved):
Although may feel unrequired if tested without refreshing the chat, it is needed on when a chat that has used HypaV2Memory before.

Good, and seems logical. Please check on the comment on this PR though. That is also a problematic part.

Bo26fhmC5M:
> ```ts
> const fullResult = `<Past Events Summary>${mainPrompt}</Past Events Summary>\n<Past Events Details>${chunkResultPrompts}</Past Events Details>`;
>     const fullResultTokens = await tokenizer.tokenizeChat({
>         role: "system",
>         content: fullResult,
>     });
>     currentTokens += fullResultTokens;
> ```
> 
> remove this part as this is reduandant and later on create similar issues over chat

I discovered something interesting in index.svelte.ts. Initially, the hypaMemoryV2 function receives currentTokens as a parameter, which includes an addition of maxResponse. However, the currentTokens returned by this function are not reused, and the input tokens are recalculated instead.

Therefore, I believe that the returned value of currentTokens is not particularly important. Additionally, since maxResponse is not the same as allocatedTokens, this could potentially cause issues when deciding whether to perform summarization.

// https://github.com/kwaroran/RisuAI/blob/821ba1812f63453bd7c1c2ef386dec85077d598f/src/ts/process/index.svelte.ts#L456
// Awaiting tokenize current
let currentTokens = DBState.db.maxResponse;

// https://github.com/kwaroran/RisuAI/blob/821ba1812f63453bd7c1c2ef386dec85077d598f/src/ts/process/index.svelte.ts#L811
const sp = await hypaMemoryV2(chats, currentTokens, maxContextTokens, currentChat, nowChatroom, tokenizer);
if (sp.error) {
    console.log(sp);
    alertError(sp.error);
    return false;
}
chats = sp.chats;
currentTokens = sp.currentTokens;

// https://github.com/kwaroran/RisuAI/blob/821ba1812f63453bd7c1c2ef386dec85077d598f/src/ts/process/index.svelte.ts#L1122
// Token rechecking
let inputTokens = 0;

for (const chat of formated) {
    inputTokens += await tokenizer.tokenizeChat(chat);
}

Bo26fhmC5M:
I'll test if adding this line solves the issue where summarization keeps occurring, which happens when the maximum response size is very large (e.g., 8192).

currentTokens -= db.maxResponse;

LightningHyperBlaze45654:
Will revise it tommorow, but I do not think the returned tokens are not reused.

Bo26fhmC5M:
## Experiment 1
To verify that the currentTokens returned by the hypaMemoryV2 function are not being reused, I set currentTokens to a fixed large number:
```javascript
return {
    currentTokens: 9999999,
    chats: unsummarizedChats,
    memory: toSerializableHypaV2Data(data),
};
```

### First attempt
[HypaV2] Model being used: ada subModel
Current session tokens: 13578

### Second attempt
[HypaV2] Model being used: ada subModel
Current session tokens: 14426

### Conclusion
In the second attempt, the input tokens were calculated correctly, indicating that the returned currentTokens value is NOT being reused.

## Experiment 2
To verify that the currentTokens parameter in the hypaMemoryV2 function includes unnecessary maximum response size, tests were conducted with maximum response sizes of 4096 and 8192.

### maxResponse = 4096
currentTokens (initial): 78210

[HypaV2] Model being used: ada subModel
Current session tokens: 14674

### maxResponse = 8192
currentTokens (initial): 82306

[HypaV2] Starting summarization iteration:
Current Tokens (before): 21770
Max Context Tokens: 20000
StartIdx: 138
chunkSize: 2000

[HypaV2] Model being used: ada subModel
Current session tokens: 16858

### Conclusion
With a maximum response size of 4096, the initial currentTokens is 78210.
With a maximum response size of 8192, summarization was triggered due to exceeding the maximum context, and the initial currentTokens was 82306, which is exactly 4096 (the difference in maximum response size) more than the previous case (78210 + 4096 = 82306).

Bo26fhmC5M:
@kwaroran 
There is an issue in Hypa V2 where token calculation unnecessarily includes **maxResponse**, causing unintended summarization to occur. I'd like to ask if it's okay to modify this part of the code:

https://github.com/kwaroran/RisuAI/blob/821ba1812f63453bd7c1c2ef386dec85077d598f/src/ts/process/index.svelte.ts#L456
```javascript
let currentTokens = 0
```
If there are concerns about impacts on other code, I can address this issue by adding the following line in hypav2.ts:

```javascript
currentTokens -= db.maxResponse;
```

LightningHyperBlaze45654:
It is quite concerning that there was also an issue. I will keep on looking through any issues and alert on this PR for QA purposes.
👍

PR #707, opened 2025-01-06
fix: resolve state_unsafe_mutation in getChatVar()

bangonicdd2:
# PR Checklist

# Description
While investigating an issue where variable calls were not possible in background embedding, I encountered a state_unsafe_mutation error. It appears that modifying the null check operator on the relevant line can resolve this issue. Thank for your time.


bangonicdd2:
I also found the issue where the order flag in the regex script was not working as intended and included a fix in an additional commit.

PR #711, opened 2025-01-07
fix: Fix message handling and summarization in HypaV2

Bo26fhmC5M:
# PR Checklist

# Description
This PR addresses several issues in the HypaV2 memory system:

1. Improved Summarization Error Handling
- Before: When unable to summarize messages while exceeding context size, HypaV2 would continue and make API requests with unintentionally truncated context
- After: HypaV2 now provides clear error messages asking users to adjust their settings when:
  - Unable to summarize last 4 messages
  - A single message exceeds chunk size

2. Message Order Preservation
- Before: lastTwoChats logic was causing the first message (index 1) and user input (index 2) to appear in the wrong order
- After: Fixed by adjusting the initial index to preserve the original message sequence

3. Message Filtering
- Added logic to skip unnecessary messages during summarization:
  - Excludes [Start a new chat] message at index 0
  - Skips empty messages
<br>
Requesting review from @LightningHyperBlaze45654

### DIFF
--- a/src/ts/process/memory/hypav2.ts
+++ b/src/ts/process/memory/hypav2.ts
@@ -346,12 +346,16 @@ export async function hypaMemoryV2(
     memory?: SerializableHypaV2Data;
 }> {
     const db = getDatabase();
-    currentTokens -= db.maxResponse
     let data: HypaV2Data = {
         lastMainChunkID: 0,
         chunks: [],
         mainChunks: [],
     };
+
+    // Subtract maxResponse from currentTokens to fix the token calculation issue.
+    // This is not a fundamental solution but rather a temporary fix.
+    // It is designed to minimize impact on other code.
+    currentTokens -= db.maxResponse;
 
     if (room.hypaV2Data) {
         if (isOldHypaV2Data(room.hypaV2Data)) {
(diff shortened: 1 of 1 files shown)


Bo26fhmC5M:
@kwaroran WIP, Please do not merge this PR yet.

LightningHyperBlaze45654:
Is [Start a new chat] removed? Then, there is no problem using this.

Previously, RisuAI always unshifted this on the chat regardless of any case.

So the index 0 was that placeholder, while 1st index having the memory prompt 'technically'.

Bo26fhmC5M:
Checking the HypaMemoryV2 logs - [Start a new chat] is still present at index 0, with the first message at index 1.

The PR code maintains the exclusion of [Start a new chat] from summaries, and when there's a first message present, summaries will begin from index 1.

Bo26fhmC5M:
Now, this PR is ready for review. I have tested these changes locally and verified that all the fixes are working as expected.

LightningHyperBlaze45654:
Seems like previous version's summary xml tag capsulated hypav2 retrieved data has been fixed to go into the prompt and not in the chat context. LGTM, +1

PR #738, opened 2025-02-01
add custom flags for deepseek proxy

bangonicdd2:
# PR Checklist

# Description
This PR simply binds the already implemented flags to the UI.


kwaroran:
It is already commited in 25aa83877b42d917f6c61aa2f9d04b1e71274b81

PR #745, opened 2025-02-06
refactor(parser): import only the required asset when in a node environment

ThePotioner:
# PR Checklist
- [X] Have you checked if it works normally in all models? *Ignore this if it doesn't use models.*
- [X] Have you checked if it works normally in all web, local, and node hosted versions? If it doesn't, have you blocked it in those versions?

# Description
This PR contains modifications to import only the required assets when talking to Asset Bot.

- refactor parseAdditionalAssets()
- Modified the getClostMatch() to work well with Object<key, value> to which it is dynamically added.

### DIFF
--- a/src/ts/parser.svelte.ts
+++ b/src/ts/parser.svelte.ts
@@ -298,21 +298,21 @@ export const assetRegex = /{{(raw|path|img|image|video|audio|bg|emotion|asset|vi
 async function replaceAsync(string, regexp, replacerFunction) {
     const replacements = await Promise.all(
         Array.from(string.matchAll(regexp),
-            match => replacerFunction(...match as any)));
+            match => replacerFunction(...match as any)))
     let i = 0;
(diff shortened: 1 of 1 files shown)


ThePotioner:
Personally, I believe that if additionalAssets in the character interface were a unique hashmap of the form Object<key, value>, other parts of the code could achieve better performance with fewer modifications.

However, modifying a definition within an interface can have a significant impact on other code, so I'm bringing this up cautiously.

ThePotioner:
I've confirmed that the code works correctly on both mobile and desktop environments, and there are significant performance improvements. The complete code is ready, and I'm awaiting a code review before merging.

ThePotioner:
I've confirmed that the code works correctly on both mobile and desktop environments, and there are significant performance improvements. The complete code is ready, and I'm awaiting a code review before merging.

PR #755, opened 2025-02-09
feat: reintroduce module integration

bangonicdd2:
# PR Checklist

# Description
This PR reintroduces the module integration feature and includes the following changes:  

1. Modules activated via integration, rather than global activation from the module menu, are now visually distinguished with a separate color.
image

2. When a module is globally activated after being enabled through integration, it will no longer be applied twice.  
3. Added a button in the developer tools to view a list of all currently active modules.  
image 
image

4. Fixed an issue where the `module_enabled` cbs was operating based on module names instead of namespaces, contrary to the documentation. It now also detects modules activated through integration.  
5. Fixed a bug where a module would fail to activate properly if it was first integrated when not present and later imported.  

Let me know if any further adjustments are needed!

### DIFF
--- a/src/lib/Setting/Pages/BotSettings.svelte
+++ b/src/lib/Setting/Pages/BotSettings.svelte
@@ -643,6 +643,10 @@
         {/if}
     </Arcodion>
 
+    <Arcodion styled name={language.moduleIntergration} help="moduleIntergration">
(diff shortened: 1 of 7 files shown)


bangonicdd2:
6. Fixed an issue where changing the namespace of a previously disabled module to an already integrated but nonexistent namespace would prevent the module from activating properly.

bangonicdd2:
7. Fixed an issue where a module activated through integration would remain active until the app was restarted when it was removed.

PR #756, opened 2025-02-10
feat: reintroduce module integration

bangonicdd2:
# PR Checklist

# Description
This PR reintroduces the module integration feature and includes the following changes:  

1. Modules activated via integration, rather than global activation from the module menu, are now visually distinguished with a separate color.
image

2. When a module is globally activated after being enabled through integration, it will no longer be applied twice.  
3. Added a button in the developer tools to view a list of all currently active modules.  
image 
image

4. Fixed an issue where the `module_enabled` cbs was operating based on module names instead of namespaces, contrary to the documentation. It now also detects modules activated through integration.  
5. Fixed issues where module integration did not properly sync with module additions, modifications, and deletions.

Let me know if any further adjustments are needed!

### DIFF
--- a/src/lib/Setting/Pages/BotSettings.svelte
+++ b/src/lib/Setting/Pages/BotSettings.svelte
@@ -643,6 +643,10 @@
         {/if}
     </Arcodion>
 
+    <Arcodion styled name={language.moduleIntergration} help="moduleIntergration">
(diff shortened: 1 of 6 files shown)


bangonicdd2:
A community post reported a module-related bug, but I wasn’t able to identify or reproduce the issue in my environment. For now, this PR is ready to be merged.

PR #760, opened 2025-02-15
fix: issue with asset names case-insensitive

bangonicdd2:
# PR Checklist

# Description
This PR addresses the issue where asset names were only recognized when entered in lowercase, even if they were originally uppercase. It ensures that asset names are now case-insensitive, aligning with the intended behavior.


bangonicdd2:
After reviewing the context of the previous PRs, I believe this adjustment best preserves the intent of the original.

PR #765, opened 2025-02-19
feat: add local activation for lorebook

bangonicdd2:
# PR Checklist

# Description
This PR adds a toggle to enable lore books specifically for the current chat session. This feature is expected to be particularly useful for simulator-style chatbots as it allows users to select and activate specific lore books for each chat from a pre-configured set. Please note that this is still an early implementation, and I'm open to any feedback or suggestions.

image

### DIFF
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -1056,4 +1056,5 @@ export const languageEnglish = {
     copy: "Copy",
     paste: "Paste",
     depth: "Depth",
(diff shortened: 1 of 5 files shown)


bangonicdd2:
After some consideration, the current fake-sync concept might confuse users who want to constantly edit their lore books.

I'll refine this and work on implementing a proper global/local lore book sync feature.

Issue #735, opened 2025-01-30
Support for custom embedding API for long-term memory


funnycups:
It seems that custom embedding API for long-term memory is not supported yet. Would that be possible to add an OpenAI-compatible custom API settings in "other bots -> long-term memory" setting?

cubicj:
Resolved in PR #776 (2025-03) — OpenAI-compatible custom embedding API is now supported in long-term memory settings.

PR #776, opened 2025-02-27
feat: add support for OpenAI-compatible embedding API

Bo26fhmC5M:
# PR Checklist

# Preview
preview

# Description
This PR introduces support for OpenAI-compatible embedding API


Bo26fhmC5M:
Resolves #735.

PR #780, opened 2025-03-07
Update providers list in OpenrouterSettings.svelte

cloak1505:
# PR Checklist
- [X] Have you checked if it works normally in all models? *Ignore this if it doesn't use models.*
- [(X)] Have you checked if it works normally in all web, local, and node hosted versions? If it doesn't, have you blocked it in those versions?

~~I'm an idiot but I see no reasons why this wouldn't work, it's just a list update. (｡•̀ᴗ-)✧~~
Edit: Yes, it does work in local Docker copy.

# Description
Copied directly from https://openrouter.ai/docs/features/provider-routing


cloak1505:
image
What the heck is this? It wasn't showing on my test branch.

PR #784, opened 2025-03-10
fix: treating unary minus in calcString

bangonicdd2:
# PR Checklist

# Description
It has been reported that calcString incorrectly calculates expressions when a negative number appears on the right-hand side. This pull request addresses this issue.

### DIFF
--- a/src/ts/process/infunctions.ts
+++ b/src/ts/process/infunctions.ts
@@ -17,6 +17,7 @@ function toRPN(expression:string) {
         '≤': {precedence: 1, associativity: 'Left'},
         '≥': {precedence: 1, associativity: 'Left'},
         '=': {precedence: 1, associativity: 'Left'},
+        '≠': {precedence: 1, associativity: 'Left'},
         '!': {precedence: 5, associativity: 'Right'},
     };
     const operatorsKeys = Object.keys(operators);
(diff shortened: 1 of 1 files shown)


bangonicdd2:
I also found a report that the '!=' symbol exists in the docs but doesn't actually work, so I added another commit. Thanks for your time.

kwaroran:
Wouldn't these break the existing bots?

sirius422:
Indeed, I also noticed the abnormal behavior of != in numerical calculations today.

| expression (calcString) | result | expression (basicMatcher) | result |
| ----------- | ----------- | ----------- | ----------- |
|{{? 2 != 0}} | 0 |  not_equal(2,0)| 1|
|{{? 2 != 1}} | 1 |  not_equal(2,1)| 1|
|{{? 2 != 2}} | 0 |  not_equal(2,2)| 0|
|{{? 2 != 3}} | 0 |  not_equal(2,3)| 1|
|{{? 2 != 4}} | 0 |  not_equal(2,4)| 1|


For example, `{{? 2 != 3}}`, this occurs because `!=` is not defined in function `toRPN`, so when converting it from Infix notation to RPN, it's treated as `2 !0 = 3`, and 2 was also discarded in the stack calculation.
`{{? 2 != 3}}`  --parsed inside `toRPN`--> `[ "2", "!", "0", "=", "3"]` --`toRPN` returned--> `"2 0 ! 3 = "`

I was about to make a pull request when I noticed your work. I'm glad to see that commit 681846c resolved this issue.

PR #803, opened 2025-04-02
refactor: regex caching

bangonicdd2:
# PR Checklist

# Description

### Since this PR affects both editRequest mode and editDisplay mode, I recommend reviewing #812 first for context.

This PR aims to:

1.  Fix an issue where the current regex caching is incompatible with modules.
2.  Slightly modify the hashing structure to handle non-deterministic regex with cbs. Verified that this correctly handles expressions like `{{#if A}}`.

Thanks for your time.

### DIFF
--- a/src/ts/process/scripts.ts
+++ b/src/ts/process/scripts.ts
@@ -68,12 +68,20 @@ export async function importRegex(o?:customscript[]):Promise<customscript[]>{
 let bestMatchCache = new Map<string, string>()
 let processScriptCache = new Map<string, string>()
 
-function cacheScript(scripts:customscript[], data:string, result:string, mode:ScriptMode){
-    let hash = data + '|||' + mode + '|||'
-    for(const script of scripts){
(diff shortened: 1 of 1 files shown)


sub-hub:
Thank you! While testing again, I confirmed that it's difficult to exceed a caching size of 1000 in one chat session.
1000 seems to be both a small capacity and an appropriate size.

bangonicdd2:
Now that I've fixed the lua display issue, I'd like to do some more testing to see what impact this commit has on display regexes. This may take some time due to my personal schedule, so I apologize to anyone who might have to wait.

bangonicdd2:
As expected, there was an issue where the regex cache returned stale values even after the Lua trigger was updated.

Last Commit Changes:
- Regex caching now runs *after* Lua triggers (and plugin).
- Allows Lua triggers to correctly utilize CBS syntax by additional using of risuChatParser. While further optimization might be possible, I've prioritized minimizing changes to the existing codebase.
- Skip regex scripts with an empty `IN` to avoid app freezes.

PR #805, opened 2025-04-02
Refactor: Caching tokenize result

sub-hub:
# PR Checklist

# Improved performance
image
The picture above was taken when I sent a chat.
As you can see, the encode function occupies **0.68 seconds** out of 1.8 seconds.
This is **36.5%** of sending time
image
This is a picture taken after adding a caching function.
encode time Reduced from **0.68 seconds to 0.12 seconds, an improvement of 82%**

Since configuration values are also used as key values in the hash function,
it is guaranteed that **different results will be produced when the configuration changes**.
I tested this myself by changing the bot settings

# Implementation
I added the `mnemonist` library to implement this. At first, I used lru-cache, but that library recommended using this one if performance is important.
It actually performed better too.

# Integrity Check
image
Lastly,
the encoding result calculated without caching when sending the first chat,
the newly calculated value when sending the second chat, and
the result cached when sending the first chat were all the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "lucide-svelte": "^0.292.0",
     "markdown-it": "^14.1.0",
     "ml-distance": "^4.0.1",
+    "mnemonist": "^0.40.3",
     "mobile-drag-drop": "3.0.0-rc.0",
     "msgpackr": "1.10.1",
     "node-html-parser": "^6.1.12",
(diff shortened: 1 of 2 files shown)


sub-hub:
Sorry. After testing again, it seems the md5 function doesn't help at all according to js hash map theory.
But it's actually better, it reduced the more hashing time by 0.1 seconds.

PR #807, opened 2025-04-03
refactor: luaFactory threadsafe

bangonicdd2:
# PR Checklist

# Description
This PR addresses an issue where luaFactory fails to initialize correctly in 'editDisplay' mode under specific conditions.

## Problem Reproduction:
The initialization failure can be reliably reproduced under the following conditions:
1. Disable all modules.
2. Select any chatbot with 'editDisplay' in trigger lua.
3. Select or make chat with length of 2 or more.
4. Restart Risu and revisit the chatbot

## Initial Approach and Findings:
The first commit in this PR attempts to resolve the immediate problem by making the luaFactory initialization thread-safe.

However, further testing with longer chat (length 120+) revealed different issue: up to 30 luaEngine instances for 'editDisplay' mode were being created concurrently. While the Garbage Collector eventually cleans these up, I believe this behavior is not efficient or intended.

## Proposed Solution:
To address the underlying problem of excessive instance creation, I plan to implement an luaEngine pool specifically for 'editDisplay' mode.

### DIFF
--- a/src/ts/process/lua.ts
+++ b/src/ts/process/lua.ts
@@ -20,36 +20,18 @@ let LuaEditDisplayIds = new Set<string>()
 let LuaLowLevelIds = new Set<string>()
 
 interface LuaEngineState {
-    code: string;
-    engine: LuaEngine;
+    code?: string;
+    engine?: LuaEngine;
     mutex: Mutex;
-    chat: Chat;
-    setVar: (key:string, value:string) => void,
-    getVar: (key:string) => string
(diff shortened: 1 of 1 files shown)


bangonicdd2:
After testing, it looks like implementing a pool might be unnecessary here; Standard Promises seem sufficient.
I'll perform a final check for compatibility with the existing code before committing.

bangonicdd2:
I will make new PR with resolving conflicts.

PR #809, opened 2025-04-06
Add fixed chat textarea option to AccessibilitySettings

kyh11199:
# PR Checklist

# Description
This pull request introduces the following changes:
- Adds a new "fixedChatTextarea" option to control the positioning of the chat window text area.
- Modifies Shift+Enter functionality to align with updated input handling requirements.

To be honest, I'm not entirely sure if this approach is correct. Since I've developed with React and Next.js, I'm still a beginner when it comes to Vite. But as a PC user, I think this method will be convenient to use.


bangonicdd2:
For second change, since we already have 'Send Chat' option within the hotkey settings, I'm worried that this change might be redundant and could potentially cause confusion or conflicts for users who have already set up their hotkeys.

Would you consider reverting this part of the change via a new PR? Alternatively, I can handle the revert if that's easier for you.

Thanks for your time.

PR #821, opened 2025-04-19
feat: add persona sort

bangonicdd2:
# PR Checklist

# Description
This PR introduces the persona sorting feature.

**Changes:**

1.  Uses the `sortable` library, consistent with other sorting functionalities.
2.  To prevent desynchronization between the selected persona and the UI value during sorting:
    *   `OnStart`: Saves the currently selected persona.
    *   `OnEnd`: Switches selection to the first persona without saving the initial selection.

I believe I've tested this sufficiently, but please feel free to point out any potential issues or edge cases I might have missed.

Thank you for your time and review!


bangonicdd2:
I realized that resetting the selected persona after every sort was inconvenient. I've updated it to maintain the currently selected persona after sorting.

PR #826, opened 2025-04-21
Add: Caching tokenize

sub-hub:
# PR Checklist

# Description
Previous PR #805 had a bug due to incorrect flow of if statements. I apologize for this.
In this PR, I have improved the flow of if statement additionally from the modification I made before.
Please refer [Fix: Correct tokenize flow in tokenizer encode function](https://github.com/kwaroran/RisuAI/commit/09228f3f867a2f05b24e5bf1163ab743e3913894) To see the how I changed the flow.

The biggest change of this commit is that it resolves the issue of duplicate tokenization that was occurring
By add a `Line 122: if (result === undefined) {`
I think this change corrected the error even if people didn't turn on the caching option.

And there is an error in the tikJS function, independent of this wrong flow.
https://github.com/dqbd/tiktoken/issues/35

### DIFF
--- a/src/ts/tokenizer.ts
+++ b/src/ts/tokenizer.ts
@@ -209,6 +209,7 @@ async function gemmaTokenize(text:string) {
 
 async function tikJS(text:string, model='cl100k_base') {
     if(!tikParser || lastTikModel !== model){
+        tikParser?.free()
         if(model === 'cl100k_base'){
             const {Tiktoken} = await import('@dqbd/tiktoken')
             const cl100k_base = await import("@dqbd/tiktoken/encoders/cl100k_base.json");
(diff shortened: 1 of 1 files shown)


sub-hub:
I checked the impact of new Tiktoken().free() with a simple test.
This is a test code
```js
for (let index = 0; index < 9999999; index++) {
    console.log(`${index} iteration - without free`)
    console.time("new parser")
    tikParser = new Tiktoken(
        o200k_base.bpe_ranks,
        o200k_base.special_tokens,
        o200k_base.pat_str
    );
    tikParser.free() // <<< IMPORTANT!
    console.timeEnd("new parser")
}
```


with free

**^ Test result with** `tikParser.free()`

without free

**^ Test result without** `tikParser.free()`

As you can see, if there is free, it will maintain a constant speed,
if there is no free, the speed will gradually slow down.
I tested it on my computer and didn't get a runtime error, but it could be different in a mobile environment.

sub-hub:
I fix it by add `tikParser?.free()` in tikJS function, 
Can you check it please with my caching tokenize flow?
Thank you!

PR #831, opened 2025-04-22
feat: Allow deleting an orphaned summary from HypaV3 

alattalatta:
# PR Checklist

# Description

Currently, you can:

- Manually rewrite summaries
- Delete orphaned summaries (all or none, in the settings)
- Delete all summaries after one

But you can't:

- Manually add summaries
- Delete a summary

This PR aims to add a feature for the second one - you can pick and delete a specific summary.

As I'm not yet familiar with the implementation of HypaV3, this PR only allows deleting orphaned summaries to minimize any unforeseen side-effects.

cn, de, es, vi, zh-Hant strings were made with Gemini. Reviews welcome.


Bo26fhmC5M:
Thank you for your PR.

Regarding your suggestion for "Manually add summaries," I believe this doesn't align with the design philosophy of HypaV3.
Hypa's role is to automatically generate a summary for multiple messages based on settings and retrieve it when needed.

In fact, Hypa was designed with the assumption that a summary would be deleted alongside its connected messages. The orphaned memory preservation feature was mainly implemented to mimic the behavior of previous HypaV2 version.

Nevertheless, I do agree that users might still want to delete individual summaries when the orphaned memory preservation option is enabled.

alattalatta:
I somewhat hoped to utilize HypaV3 as pseudo-RAG, but, I respect you the author's direction. In that case, restricting it to only orphans was a better choice than I anticipated.

PR #834, opened 2025-04-26
Add support for remote asset server

niceandneat:
# PR Checklist
- [ ] Have you checked if it works normally in all web, local, and node hosted versions? If it doesn't, have you blocked it in those versions? - I've checked web version only.

# Description

This setting allows assets to be loaded from a remote server rather than from local cache. If no setting is provided, assets will continue to be loaded from the local environment as before.

It is designed to enable functionalities such as asset replacement, random usage of multiple assets under a single asset name, and dynamic asset generation via an external server — without the need to upload assets locally or modify prompts.

# Example

## Advanced Settings

image

## Chat

`{{raw:asset_name}}`

image

`{{image:asset_name}}`

image

`{{video:asset_name}}`

image


kwaroran:
Thanks for your PR, but after consideration, I have decided to not to merge this PR, because adding a remote asset server might hurt the propose of the local/web version splitting, and might cause fragmentation, which might make hard to maintain.

PR #856, opened 2025-05-18
feat: Add 'divider' and 'group'/'groupEnd' type toggle

enzi221:
# PR Checklist

# Description

## Dividers

| Before | After |
| --- | --- |
|image | image|

Instead of pseudo-dividers with unwanted checkboxes, this allows creators to define a "divider" *toggle*.

image

Dividers are created by keyless `=name=divider`, or keyless valueless `==divider`.

```
=My Module=divider
==divider
```

Multiple dividers with equal label will merge into one to prevent too many dividers from multiple modules covering everything up.

image

```
// these won't merge
==divider
=K=divider
// these will merge
==divider
==divider
```

## Groups

image

Groups are opened by keyless `=name=group` and closed by keyless valueless `==groupEnd`. Nesting is not supported.

| Opened groups | MobileGUI |
|---|---|
| image | image |

### DIFF
--- a/src/lib/SideBars/Toggles.svelte
+++ b/src/lib/SideBars/Toggles.svelte
@@ -46,7 +46,7 @@
                     class:justify-end={$MobileGUI}>
                     {toggle.value}
                 </summary>
-                <div class="flex flex-col px-2 py-4 pt-0">
+                <div class="flex flex-col px-2 pt-2 pb-4">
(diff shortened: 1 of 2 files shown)


enzi221:
@kwaroran Here you go, pushed.

image

It looks fine in the character tab.

image

A bit too large in the chat list tab I think?

PR #869, opened 2025-05-24
Fix: Variable name highlighting inconsistency in v2setvar and v2If

cubicj:
# PR Checklist

# Description
Fixed inconsistent variable name styling in v2setvar and v2If. All variable names now display consistently in yellow.


kwaroran:
This will break backward compatibility, since this modifies the type from string-only to variable.

PR #881, opened 2025-05-28
feat: add per-message user commentary feature

poroyo:
# PR Checklist

## What's New
- Allows users to add, edit, and delete comments for each message
- Comments are saved manually, with an optional lock feature to prevent editing or deletion
- Displays the creation date, last updated date, and byte size for each comment.
- Shows UTF-8 byte size live while editing
- Compatible with the existing session JSON save/load

## Why this feature may be useful

There may be times when users want to leave a short note or thought related to a specific message in a session. Until now, they needed to use a separate note-taking app, which could be inconvenient and distract from the chat experience, since the notes weren't directly connected to the messages.

With this feature, users can directly add comments, notes, or reminders to each message inside RisuAI, making it easier to read, edit, or delete them later.

---

- No comments yet

---

- Users can create a new comment by clicking the "+" button at the bottom.
- Clicking the "lock" button locks the comment, preventing accidental editing or deletion, while still allowing the comment to be read.
- Clicking the delete button shows a confirmation dialog to avoid accidental deletion.
- To prevent storage or management issues, each message currently allows up to 20 comments.

---

- While editing, the current byte size of the comment is shown on the bottom-left.
- Users need to press the "Save" button to save any changes.

---

- Screenshot after deleting `test3`

---

### Why manual saving rather than automatic?

- **Clearer update dates**: Automatic saving may cause confusion, as the update date would change with every character typed. In contrast, manual saving updates the date only when the user intentionally saves changes, making it easier to keep track of meaningful updates.
- **Reducing mistakes**: Manual saving gives users the ability to easily cancel edits if they make a mistake, without needing an "undo" feature.

### Why dates are stored as `strings` instead of `Date` objects?

- For compatibility when exporting and importing sessions as JSON, storing dates as strings (in ISO format) was safer. Internally, these strings are converted back to `Date` objects only when needed.

```ts
export interface Commentary {
    id: string
    title: string
    createdAt: string   // ISO string for JSON compatibility
    updatedAt: string   // (Date becomes string when exported/imported)
    content: string
    locked?: boolean
}
```

### Why use the field name `commentaries`?

- I felt `memo` or `note` might have been clearer, but those names were already in use for other purposes. The term `comments` also seemed likely to be used internally in the future, so I chose `commentaries` to avoid potential confusion.
  On the user interface, the simpler word "comments" is displayed.

## Possible next steps and ideas

I believe this feature could enhance user engagement, allowing them to easily capture thoughts and notes within RisuAI itself.

Possible future improvements might be:
- Showing visually in the chat UI whether a message has comments.
- Showing a comment count in the message UI.
- Adding sorting or reordering features for convenience.

If you have any concerns or suggestions, or if you feel this feature does not align with the project's direction, please let me know. I'll be happy to discuss further!

Thanks a lot!

### DIFF
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -1128,6 +1128,8 @@ export const languageEnglish = {
     promptInfoEmptyMessage: "No prompt information is available for this message.",
     promptInfoEmptyToggle: "No custom toggles are currently active.",
     promptInfoEmptyText: "No prompt text has been saved.",
+    commentaries: "Comments",
+    commentariesEmptyMessage: "No comments yet.",
     escapeOutput: "Escape Output",
     claudeBatching: "Claude Batching",
     claude1HourCaching: "Claude 1 Hour Caching",    
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -1130,6 +1130,7 @@ export const languageEnglish = {
     promptInfoEmptyText: "No prompt text has been saved.",
     commentaries: "Comments",
     commentariesEmptyMessage: "No comments yet.",
+    commentariesDelete: "Are you sure you want to delete this comment?",
     escapeOutput: "Escape Output",
     claudeBatching: "Claude Batching",
     claude1HourCaching: "Claude 1 Hour Caching",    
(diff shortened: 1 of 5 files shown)


kwaroran:
In my opinion, I think the user commentary is too resource consuming, since it can be in every messages, and it has more features than just comments like locks, while we still have lightweight `{{// }}` CBS for memo.

I think this should be a plugin, not in a main branch.if you need a API for making plugins, please make a PR about it.

PR #884, opened 2025-05-31
refactor: Do not iterate whole messages in messageForm

enzi221:
# PR Checklist

# Description
Current impl of `messageForm()`:
1. Iterate messages from 0 to end, in forward, unshift (reverse) to a new array.
2. Slice to `loadPages`.
3. Return.

New impl:
1. Iterate messages from end to `end - loadPages`, in backward, push to a new array.
2. Return.

This isn't much, but still a free real estate of optimization. My 243 messages was taking 0.3ms in `messageForm()` on first load, now more or less 0.1ms.


kwaroran:
Thanks for your PR, but there is a reason of Iterating messageForm from 0 to end. without this, svelte sometime updates the comp incorrectly.

PR #898, opened 2025-06-05
refactor: Improve chat rendering performance

enzi221:
# PR Checklist

# Description

| Upstream | PR |
| --- | --- |
| [upstream.webm](https://github.com/user-attachments/assets/2c411d97-a865-4538-b6c3-a77b4d748c0c) | [modified.webm](https://github.com/user-attachments/assets/7fd0c58e-a4ca-4a67-ab4d-b5e5786d86a2) |

It takes more than 22 seconds (I gave up) to traverse 243 chat messages with fair amount of Lua scripts while in upstream. With this PR, I can move up to the top in less than 10 seconds.

Other metrics: One button click costs 2.43s in said chat. Reduced to 1.2s with PR changes.

### Limiting reactivity

The main goal is to suppress unnecessary re-renders of `Chat` which is VERY expensive. By reducing direct global store accesses, reducing further reactivity by moving some props out of reactive object with `@const`, and defining key in `#each`, I limited `Chat` re-renders to those which *actually* changed.

### Granular re-renders

Also in this PR is a new global store `ReloadChatPointer: Record<number, number>.` It is like `ReloadGUIPointer` which redraws the whole chat messages, but controlled per-index.

```
{@const chatReloadPointer = $ReloadChatPointer[idx] ?? 0}
{#key `${totalLengthPointer}|${chatReloadPointer}`}
```

(I'll get into `totalLengthPointer` below)

`ReloadChatPointer` changes for/when:

- Button clicks increase the chat's pointer.
- Manually calling `reloadChat(triggerId, index)` from Lua increases the index pointer.
- Manually calling `Update Chat At Index` from TriggerV2 increases the index pointer.

The `totalLengthPointer` is to force re-renders of the most recent 5 chats.

```
{@const totalLengthPointer = (idx > totalLength - 6) ? totalLength : 0}
```

So whenever a chat gets added or removed, 5 chats can react to it.

> [!NOTE]
> Since regex results are cached and not affected by chat length changes, this re-rendering can't guarantee full update. It is what it has been, not this PR's behavior.

### Misc optimizations

The most heavy function is `ParseMarkdown()`. I've separated the caller into `ChatBody`, so:

```
DefaultChatScreen < Chat list rendering
  Chat < Wrapper, manages re-renders, interaction point
    ChatBody < Textual contents
```

Some reactive values were implemented with `$effect`s. Refactored few of those to `$derived` if applicable.

I've also delegated `risu-btn` and `risu-trigger` event listeners to `Chat`'s root element for lower memory usage.


kwaroran:
Great Job! but although the PR looks pretty good, since I am working on another optimization implementation, After I finish the implementation, I will check it could be merged or not with that optimization.

enzi221:
@kwaroran Sure, I'll try to keep it in sync with main.

PR #901, opened 2025-06-07
Added draggable presets

risuai:
# PR Checklist

# Description
Added draggable presets with minor visual changes(added dividers as the drag-and-drop zone). As there are only 9 presets available for hotkey assignments, dragging newly added presets up is needed for ease of use.
Based heavily on the PromptDataItem drag function.


kwaroran:
Resolved by #900

PR #902, opened 2025-06-07
Fix chat message copy functionality

dootaang:
# PR Checklist

# Description


kwaroran:
This PR looks to reverts #898 just to fix the chat copy functionality, which makes priorities disordered.

kwaroran:
Also, changing the option''s functionality breaks backwards compatibility, It needs to be made as a new option, rather than replacing it.

dootaang:
I understood. Thank you for explain!

PR #929, opened 2025-06-13
feat(memory): add new long-term memory provider

clwmz:
# PR Checklist

# Description

This pull request introduces **UlariMemory**, a new long-term memory provider that leverages a hybrid Retrieval-Augmented Generation (RAG) approach to RisuAI. It combines semantic vector search with traditional keyword search to provide more relevant and context-aware memories to the LLM. This new system is designed as an alternative to HypaMemory V3, reusing some of its core infrastructure and UI components.

**Note:** UlariMemory is **still highly experimental** and subject to ongoing testing and feedback. It is **not yet feature-complete** and will be actively modified based on user needs before this PR is merged. The name `UlariMemory` is also a work-in-progress name and may be changed to something like `HypaV4` upon finalization.

# Problem Context

The current HypaV3 implementation relies solely on vector search. While powerful for semantic similarity, it can be less effective for retrieving specific, fact-based information where exact keywords are crucial. This new implementation addresses this by creating a hybrid system that gets the best of both worlds: finding both semantically similar context and factually relevant data using a traditional keyword search engine (`minisearch`).

# Key Features & Changes

## 1. UlariMemory: Hybrid Long-Term Memory
- **Hybrid Search:** The core of UlariMemory is its ability to retrieve memories using two methods simultaneously:
    - **Vector Search:** For finding semantically similar memories.
    - **Keyword Search:** For exact-match term lookups, powered by the new `minisearch` dependency.
- **Reciprocal Rank Fusion (RRF):** Search results from both vector and keyword methods are combined using the RRF algorithm to produce a single, relevance-ranked list of memories.
- **Temporally-Weighted Search:** To prioritize recent context, user inputs are given less weight the older they are, ensuring that direct references in the latest messages are more likely to be found.
- **User Customization:** A new **Hybrid Search Weights Ratio** slider allows users to fine-tune the balance between vector and keyword search to suit their needs.
- **Seamless Integration:** UlariMemory is integrated into the existing HypaV3 workflow. Its state can be inspected and managed through the same memory modal used by HypaV3.

## 2. Core Memory System Refactoring
- To support both HypaV3 and the new UlariMemory in a maintainable way, core memory functions like `summarize`, `cleanOrphanedSummary`, and data serialization have been extracted into a shared module (`src/ts/process/memory/core.ts`).

# Technical Implementation Details
- Added `minisearch` dependency for efficient, client-side keyword searching.
- The database schema is updated to include `ulariPresets`, `ulariPresetId`, and `useUlariMemory`. Migration logic is included to create a default Ulari preset from existing HypaV3 presets.
- For data compatibility and to simplify switching, UlariMemory reuses the `hypaV3Data` field within the chat object for storing its state.

# Current Limitations
- To achieve the full potential of keyword searching, a morphological parser is highly recommended for languages like Korean. Due to performance and complexity constraints in a browser environment, this implementation currently uses the standard `Intl.Segmenter` API as a practical alternative.
- As the system is in its early stages, the summarization and retrieval logic may not yet be optimal for all use cases. It will be refined based on user feedback.
 
# Future Plans Before Merge

- [ ] Finalize the official name for this new long-term memory provider.
- [ ] Collect user feedback to tune the retrieval and summarization logic for higher accuracy and stability.
- [ ] (UI) Potentially implement a new, more intuitive ratio slider component for the `hybridSearchWeightsRatio` setting.


clwmz:
Just a quick note: my editor's auto-formatter introduced some minor formatting and style changes in this PR (e.g., whitespace, line wrapping). I'm here to revert those if you prefer the original styling.

Bo26fhmC5M:
I appreciate the effort to implement a next-generation memory system, and I think it's a great idea. However, I have concerns about refactoring HypaV3 code and modal components without prior discussion since I haven't abandoned HypaV3 development.

The current implementation creates dependencies between HypaV3 logic and the core module, which compromises the independence of the HypaV3 system. I have no objection to reusing HypaV3 code, but I would prefer if you could restructure this as an independent memory system without modifying the existing HypaV3 codebase.

When I developed HypaV3, I respected HypaV2 by not touching V2 and creating V3 as a separate system. I would appreciate if this same consideration could be reflected in your commits as well.

clwmz:
@Bo26fhmC5M 
Thank you for the feedback. I completely understand and agree with your position.
My intention was to eventually make this an independent system before merge, but I made a mistake by forgetting to mark this as a draft PR from the start.
This was based from an early prototype for Node Risu, and in my haste to get it working, I refactored the existing HypaV3 code instead of building it as a fully independent system. But I see now that when creating a GitHub Pull Request, in sum of not marking this as a draft PR, I skipped a crucial step and uploaded changes that included modification of shared code without discussing it. 
I will rework the PR to decouple the two systems completely, ensuring HypaV3's codebase remains untouched, just as you requested. Thanks for the guidance.

PR #958, opened 2025-07-08
ci: push 'latest' tag to ghcr

Acet-Aminophen:
# PR Checklist
  - ghcr test finished

# Description
Hi there.
Adding the 'latest' tag ensure consistency in the documentation and guides, even as new versions are released.


concertypin:
Oh I forgot that when I write that worflow
LGTM!

PR #968, opened 2025-07-17
ci: use pre-built image from ghcr instead of build

Acet-Aminophen:
# PR Checklist

# Description
Hi there.
 - Update README.md to run without the git pull process, as the 'latest' tag is now available.
 - Modify docker-compose.yml to prevent the stack name from the device name.
 
For your information, Windows(after april 2018 update) also supports curl.


Acet-Aminophen:
I think there is something wrong with ghcr as now.

PR #977, opened 2025-07-20
Feat: Add page indicator for swipe reroll in first message

cubicj:
# PR Checklist

# Description
image


kwaroran:
Since this could break the UI when using custom CSS or custom HTML, I think it should be an option at the display settings.

PR #979, opened 2025-07-22
ci : make container image build based on tag trigger instead of all commits in main

Acet-Aminophen:
# PR Checklist
  - ghcr test finished(https://github.com/Acet-Aminophen/RisuAI/actions/runs/16644870206)

# Description
Hi there.
 - make docker build only starts with condition that tagged with "v*"
   - when the main branch is built on every commit, it becomes unstable and too many builds.
 - add lables(name, version) on image
   - User couldn't know the version if they pull with "risuai:latest"


Acet-Aminophen:
1. It doesn't overwrite except latest. Would you please clarify what led you to think it does?
2. While the main branch is intended to be stable, this cannot always be guaranteed. Consequently, our present workflow with a commit hash for every push to main didn't consistently provide users with a truly "stable" image. For instance, you can see the issue if you try ghcr.io/kwaroran/risuai:4d1151b, an image from less than two weeks ago.
3. Versioning must be clear. If the Mainpage displays version 165.1.0 (as of now), that version should correspond to a single, fixed commit. This is not the case with our current process, which builds a new image on every commit to main.

concertypin:
> It doesn't overwrite except latest.

When user invoke action via workflow_dispatch, `${{ github.ref_name }}` will be either branch name like `main` or tag name. Tag name doesn't make problem, but branch name makes problem. How about refusing action when called by workflow_dispatch with not tag but branch?

> Would you please clarify what led you to think it does?

Sorry, my mistake. I thought `${{ github.ref_name }}` represents branch.

> While the main branch is intended to be stable, this cannot always be guaranteed. Consequently, our present workflow with a commit hash for every push to main didn't consistently provide users with a truly "stable" image.

> Versioning must be clear. If the Mainpage displays version 165.1.0 (as of now), that version should correspond to a single, fixed commit. This is not the case with our current process, which builds a new image on every commit to main.

Tag is good. But there are some bleeding-edge commits which aren't exist on tag.
image
Upper 3 commits are not in tag. I think building image per commit can be used as unstable, nightly image, while image per tag can be used as stable one.

Acet-Aminophen:
> When user invoke action via workflow_dispatch, ${{ github.ref_name }} will be either branch name like main or tag name. Tag name doesn't make problem, but branch name makes problem. How about refusing action when called by workflow_dispatch with not tag but branch?

Thank you for the great suggestion. The process has been updated to use the commit hash for builds not triggered by a tag.

> Tag is good. But there are some bleeding-edge commits which aren't exist on tag. Upper 3 commits are not in tag. I think building image per commit can be used as unstable, nightly image, while image per tag can be used as stable one.

Builds will now be created with the commit hash when committed to the branch, and with the tag name if it's a tag. The 'latest' tag will only be applied to tag-based builds.

However, I still have some concerns about the version display on the main page.

> Versioning must be clear. If the Mainpage displays version 165.1.0 (as of now), that version should correspond to a single, fixed commit. This is not the case with our current process, which builds a new image on every commit to 

untitled

I've made a commit that rewrite the parts handling the version, but I haven't created a pull request yet because I have concerns about its potential impact.
 - https://github.com/Acet-Aminophen/RisuAI/commit/345ca2668aaa7292c4dec4f829d177c288983dff

concertypin (approved):

Acet-Aminophen:
wh?y no me!rged??? I am sad :((((

PR #985, opened 2025-07-24
use variable

green1052:
# PR Checklist

# Description


kwaroran:
I don't know what is this for

green1052:
```js
const savePath = path.join(process.cwd(), "save")
c
onst passwordPath = path.join(process.cwd(), 'save', '__password')
```

PR #986, opened 2025-07-24
migrate from Express to Koa and replace lodash with es-toolkit

green1052:
# PR Checklist

# Description
Migrated the server framework from Express to Koa
Replaced lodash dependencies with es-toolkit


concertypin (commented):
Why did you migrate from Express to Koa? Is this significantly faster than original one? Since node RisuAI is designed for single-user usecase, I think this migration barely affect UX or something.

PR #994, opened 2025-08-02
fix: add nullish on chatFolders

sub-hub:
# PR Checklist

# Description

Users who have been using risu for a long time may not have chatFolders on old bots, so they get errors when importing chats.


sub-hub:
There are feedback on discord about this issue.
But now that I think about it, I don't think this is a fundamental solution...
When chatFolders is not available, I think we need to create an array. I'm not familiar with chat folders, so I'll leave that part to you.

PR #998, opened 2025-08-05
Refactoring scripting.ts

concertypin:
# PR Checklist

# Description
Since scripting.ts is too long to read, separate it into some pieces of file.


src/lib/ChatScreens/Chat.svelte
kwaroran: Why is `Chat.svelte` refactoring included? shouldn't this be another PR?

cubicj (commented):
The split itself checks out, but main kept adding APIs to `scriptings.ts` after this branch deleted it, so the merge is now a modify/delete conflict and redoing the split on top of the current main looks cheaper than untangling this branch. Would you be interested in continuing this on the current main?

PR #1007, opened 2025-08-11
fix: More robust undefined handling in google.ts

sub-hub:
# PR Checklist

# Description
Currently, when using the google model in risu,
`TypeError: Cannot read properties of undefined (reading ‘functionCall’)`
errors often occur.
I haven't been able to find the cause of this yet, but I expect that adding handling that is fundamentally correct in this pr will fix it.


sub-hub:
Furthermore, **even though many people experience this error, the error pop-up does not provide enough information** for developers to identify the cause.

#1005 
If you merge this PR, we will be able to find the issue more quickly.

kwaroran:
Fixed at e877ad344596721aa21c3ce7920b1a3703100e7f

PR #1012, opened 2025-08-13
fix: Enhance Vertex AI authentication and Web Crypto API error handling

SameDesu123:
# PR Checklist

# Description
- Added checks for missing Vertex AI authentication details (client email and private key).
- Added a check for the availability of the Web Crypto API (crypto.subtle) before using importKey.
- These changes should prevent 'Cannot read properties of undefined (reading 'importKey')' errors by ensuring necessary prerequisites are met or by providing clearer error messages.
- Added a hint to the Web Crypto API error message, suggesting HTTPS might be required.

### DIFF
--- a/src/lib/Setting/Pages/DisplaySettings.svelte
+++ b/src/lib/Setting/Pages/DisplaySettings.svelte
@@ -233,196 +233,188 @@
 {/if}
 
 {#if submenu === 2 || submenu === -1}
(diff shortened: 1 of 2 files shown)


SameDesu123:
It looks like this PR includes commits from other work. I'll separate the code and create a new PR.

PR #1024, opened 2025-08-25
perf: patch-based synchronization of DB on Node.js server

kangjoseph90:
# PR Checklist

# Description
### Problem
The previous database synchronization method transferred the entire file (~100MB) on every sync, regardless of the size of the change. This resulted in excessive network traffic and server disk I/O, leading to increased battery consumption for users.

### Solution
This pull request introduces a **patch-based synchronization system** for the Node.js server. By transmitting only the modified data (~10KB) as a lightweight patch, this approach significantly reduces network usage and server load.

# Key Changes
*   **Added `/api/patch` Endpoint:**
    *   Accepts a filename, patch data, and an expected hash to perform an in-memory patch on the server.
    *   Ensures data integrity using a hash comparison and automatically creates db backups.
    *   Implements a 5-second debounce to minimize frequent disk writes and reduce I/O load.

*   **Improved `/api/list` Endpoint:**
    *   The endpoint now supports a `prefix` parameter, allowing clients to fetch a filtered list of files (e.g., only `database/dbbackup-` files).
    *   This eliminates the need to download the entire file list (including all assets), greatly reducing network traffic for this operation.

*   **Implemented `RisuSavePatcher` Class:**
    *   Similar to the existing `RisuSaveEncoder`, this class leverages `changeTracker` to efficiently process only the modified data.
    *   It generates patches and hashes by operating on the changed data while reusing the rest, minimizing the computational overhead. (On average, it uses less computation than `RisuSaveEncoder`).

*   **Added New Dependency:**
    *   Introduced the `fast-json-patch` module to the project's dependencies for efficient patch creation and application.

# Safety & Requirements
*   **Graceful Fallback:** If a hash mismatch is detected after a patch is applied, the operation is aborted, and the system reverts to the original full-save mechanism.
*   **Automatic Version Guard:** Patch-based sync is automatically disabled on Node.js v22.7.0, v23, and above to prevent known server-side encoding errors with the `msgpackr` library.
*   **Secure Hashing:** Uses custom `normalizeJSON` and `calculateHash` functions to ensure fast and reliable patch/hash generation.
*   **Node.js Version:** It is highly recommended to use a supported version of Node.js to ensure stability and proper functionality.

# Expected Effects
*   **Drastic Network Traffic Reduction:** Sync payload size is reduced from **~100MB to ~10KB**.
*   **Negligible Performance Impact:** The overhead for patch/hash operations is minimal, typically around **~10ms**.
*   **Improved Battery Efficiency:** Reduced network activity on the client leads to lower power consumption.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "eventsource-parser": "^1.1.2",
     "exifr": "^7.1.3",
     "express": "^4.18.2",
+    "fast-json-patch": "^3.1.1",
     "fflate": "^0.8.1",
     "gpt-3-encoder": "^1.1.4",
     "gpt3-tokenizer": "^1.1.5",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -128,6 +128,9 @@ importers:
       express:
         specifier: ^4.18.2
         version: 4.18.2
+      fast-json-patch:
+        specifier: ^3.1.1
+        version: 3.1.1
       fflate:
         specifier: ^0.8.1
         version: 0.8.1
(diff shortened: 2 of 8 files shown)


kwaroran:
Thanks for the PR. However, we are planning our own chunk-based storage system that will be platform-agnostic rather than limited to Node. As such, we won't be able to merge this PR.

PR #1034, opened 2025-09-05
Fix: Robust argument parsing for {{call}} syntax to support nested CBS

nevaeh5379:
# PR Checklist

# Description
## The Problem
The previous implementation of the `{{call::...}}` syntax used a simple `string.split('::')` method to parse its arguments. This had a critical limitation: it could not correctly handle arguments that contained nested CBS (e.g., `{{getvar::...}}`, `{{slot::...}}`). This made it impossible to use powerful patterns, such as calling a function with a dynamic value from within an `{{#each}}` loop.

## The Solution
This PR replaces the simple split logic with a more robust, context-aware parser for `{{call}}` arguments. The new implementation iterates through the argument string while tracking the nesting level of curly braces (`{{...}}`). As a result, it now correctly identifies `::` as a top-level argument separator, while ignoring any `::` found inside nested syntaxes.

### Key Improvements
- **Nested CBS in Arguments:** `{{call}}` can now safely accept other CBS as arguments, enabling dynamic and modular script creation.
- **Enhanced Scripting Patterns:** This allows for advanced use cases, like using `{{#each}}` to loop through an array and call a function for each item, as demonstrated below.

## Demonstration

The following test case now works as intended.

**Test Code:**
```cbs
{{settempvar::my_fruits::{{array::Apple::Banana::Cherry}}}}

{{#func render_item}} 
  <div class="item">
    {{#when::{{arg::1}}::isnot::Banana}}
      - Fruit Name: {{arg::1}}
    {{:else}}
      - This one is a Banana!
    {{/when}}
  </div>
{{/func}}

<div class="fruit-list">
  {{#each {{gettempvar::my_fruits}} fruit}}
    {{call::render_item::{{slot::fruit}}}}
  {{/each}}
</div>
```

### Before:
The parser failed because it incorrectly split the nested `{{slot::fruit}}` and `{{when...}}` syntaxes, leading to broken output.
msedge_46ZN0CKbbh

### After:
The function is called correctly for each item in the loop, producing the expected HTML structure.
msedge_Vv4zglW0co

### DIFF
--- a/src/ts/parser.svelte.ts
+++ b/src/ts/parser.svelte.ts
@@ -8,7 +8,7 @@ import { get } from 'svelte/store';
 import css, { type CssAtRuleAST } from '@adobe/css-tools'
 import { SizeStore, selectedCharID } from './stores.svelte';
 import { calcString } from './process/infunctions';
-import { findCharacterbyId, getPersonaPrompt, getUserIcon, getUserName, parseKeyValue, pickHashRand, replaceAsync} from './util';
+import { findCharacterbyId, getPersonaPrompt, getUserIcon, getUserName, parseKeyValue, pickHashRand, replaceAsync } from './util';
 import { getInlayAsset } from './process/files/inlays';
 import { getModuleAssets, getModuleLorebooks, getModules, type RisuModule } from './process/modules';
 import type { OpenAIChat } from './process/index.svelte';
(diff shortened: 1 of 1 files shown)


kwaroran:
There is too much refactoring in this PR. Please separate the refactoring into its own PR."

PR #1047, opened 2025-09-21
feat: css class & button trigger support on background DOM

bangonicdd2:
# PR Checklist

# Description
Currently, when using HTML elements in background embedding slots, inline styles can be used, but direct CSS classes(ie. x-risu-something) cannot be specified. This change resolves this issue, but if you don't want this or have other ideas, please let me know.

(+@)
Thanks to comments, I noticed button triggers are not available in background embedding, so I added capturing of button triggers here as well. For performance reasons, chat reload is left at the trigger's discretion.

### DIFF
--- a/src/lib/ChatScreens/BackgroundDom.svelte
+++ b/src/lib/ChatScreens/BackgroundDom.svelte
@@ -1,19 +1,58 @@
 <script lang="ts">
     import { ParseMarkdown, risuChatParser } from "src/ts/parser.svelte";
-    import { type character, type groupChat } from "src/ts/storage/database.svelte";
-    import { DBState } from 'src/ts/stores.svelte';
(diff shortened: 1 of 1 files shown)


bangonicdd2:
> I suggest adding background embedding support for the risu-trigger attribute.

Could you elaborate a little more?

tammy3211:
It is not possible to activate V2 triggers or Lua script triggers using buttons or other elements created in background embedding. I wish this could be made possible.

tammy3211:
Sorry for another suggestion, but I noticed that when the chattext attribute is applied to background embedding, it can sometimes interfere with regex scripts.
(This issue already existed before, but it was not visible on the screen. Now it shows up visually, so users can notice it.)
Maybe we could consider adding a custom flag option that controls whether regex scripts are applied to background embedding (default off)?

bangonicdd2:
Actually separating out the background embedding area as new option from the existing display regex seems a neat idea, but implementing it in this PR seems out of scope.

PR #1048, opened 2025-09-24
feat: add {{loreinject::name}} to use @@inject_at decorator

tammy3211:
# PR Checklist

# Description
## Add features
- Add 'insert' operation to lorebook inject types in lorebook.svelte.ts
- Implement loreinjectParser using inject-based approach in index.svelte.ts
- Support both {{loreinject::name}} and {{lore_inject::name}} syntax
- Update CBS documentation for loreinject function

### How to use the new loreinject functionality:
1. **Create a lorebook entry with @@inject_at decorator:**
@@inject_at myTarget This content will be injected when {{loreinject::myTarget}} is used. You can include multiple lines and complex content here.

2. **Use loreinject in your prompts:**
{{loreinject::myTarget}} or {{lore_inject::myTarget}}

3. **How it works:**
- It works almost same as {{position}}
- The lorebook entry must start with `@@inject_at <targetName>`
- The lorebook must be active (keywords matched) to be available for injection
- Multiple lorebooks with the same target name will be combined

### Example use case:
- Create dynamic character descriptions that can be injected contextually
- Build modular world-building content that can be referenced anywhere
- Implement conditional lore injection based on active lorebooks

### DIFF
--- a/src/ts/process/index.svelte.ts
+++ b/src/ts/process/index.svelte.ts
@@ -471,24 +471,56 @@ export async function sendChat(chatProcessIndex = -1,arg:{
     const lorepmt = await loadLoreBookV3Prompt()
     
     const loreinjectRegex = /{{(?:loreinject|lore_inject)::(.+?)}}/g
-    const loreinjectParser = (text: string) => {
-        return text.replace(loreinjectRegex, (match, targetName) => {
-            if (!targetName || targetName.trim() === '') {
-                return ''
-            }
+    
+    const loreinjectParser = (text: string, processedTargets: Set<string> = new Set()): string => {
+        let hasMoreInjects = true
+        let result = text
+        let iterationCount = 0
+        const maxIterations = 100
+        
+        while (hasMoreInjects && iterationCount < maxIterations) {
+            hasMoreInjects = false
+            iterationCount++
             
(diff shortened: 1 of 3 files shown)


src/ts/process/index.svelte.ts
```diff
@@ -958,7 +980,7 @@ export async function sendChat(chatProcessIndex = -1,arg:{
     for(const depthPrompt of depthPrompts){
         const chat:OpenAIChat = {
             role: depthPrompt.role,
-            content: risuChatParser(depthPrompt.prompt, {chara: currentChar})
+            content: risuChatParser(loreinjectParser(depthPrompt.prompt), {chara: currentChar})
```
tammy3211, line 983: I can't find what 'depthPrompt' work. AI said it is character's system prompt and need to use 'loreinjectParser', but well.. I'm not sure.

tammy3211: Whatever depthPrompt is, there will be no errors in depthPrompt caused by 'loreinjectParser'.

kwaroran:
Appreciate the PR, but this feature is already planned for `{{position}}` CBS. As such, we can't merge this implementation at this time.

PR #1073, opened 2025-11-11
feat: implement replaceworker for UI non-blocking regex script

concertypin:
# PR Checklist

# Description
This PR implements (dedicated) web worker for regex script. Since web worker runs in their own thread, bot with huge/optimized regex script doesn't freeze the UI.
Also, there is new helper class `PromisedWorker`, which can call and get data from web worker as Promise. 
Currently, worker doesn't handle what it can't deal with its own parameter, like DB accessing. In that case, replace is done on UI thread as fallback.


cubicj (commented):
Since this branch conflicts with current main, I ported the logic delta onto main and tested there. The build side is fine, the worker file is emitted and its URL resolves. The runtime is where it breaks down.

The scripts reaching `structuredClone` in `executeScriptAsync` are Svelte `$state` proxies, since `getDatabase()` returns `DBState.db` directly, and `structuredClone` throws `DataCloneError` on a proxy before `postMessage` runs. The `needsHandover` fallback is never reached, so every script without `<...>` flag metadata is silently skipped instead of applied. I reproduced this with an actual `$state` proxy under vitest, so the worker path cannot have worked against real character or preset data.

`PromisedWorker` has no `onerror` handling and the call site passes no ttl, so a worker that fails to load leaves `processScriptFull` unresolved and chat rendering stalls with no fallback. The module scope `new Worker` also breaks six test files on current main at import time, since the vitest environment has no `Worker` global.

There are more issues behind these (per-script awaits let `{{char}}` and `@@inject` resolve against a newly selected character mid-loop, and nondeterministic CBS like `{{random}}` false-positives the dev-mode comparison), but the above already means a rebase plus a rework of the serialization and error paths, and the whole-file reformat makes that close to a rewrite. Are you interested in continuing this?

concertypin:
I measured the perf, but regex over worker makes lots of latency.

PR #1083, opened 2025-12-25
feat: Prevent form submission, allow listening to it with Lua

enzi221:
# PR Checklist

# Description

This MR allows form submission handling with Lua scripts.

## Problem

Currently, form-like interaction *can* be done with a combination of `risu-btn`/`risu-trigger` and chat variables. But each button press requires separate Lua calls and a full or partial chat update, which moves scroll position and gets more and more slower as the chat session gets longer.

With form submission, creators can implement a "batch" variable update with faster, more reactive UX, while also utilizing browser built-in form controls.

> [!NOTE]
> This MR also cancels `submit` events dispatched from inside the chat body which should help mitigate some form of attacks, such as a button that submits to an advertisement link.

## Usage

Similar to the `risu-btn` attribute, creators can add a `risu-form="key"` attribute to their `<form>`s:

```html
<form risu-form="submit-controlpanel">
  <input name="boolean" type="checkbox">
  <input name="foo" type="text">
  <button>submit</button>
</form>
```

On Lua side, they can define `onFormSubmit` like so:

```lua
onFormSubmit = function (triggerId, key, meta)
  print(key) -- 'submit-controlpanel'

  local data = json.decode(meta).data
  print('data.boolean', data.boolean) -- 'on' or nil
  print('data.foo', data.foo) -- whatever text the user has typed
end
```

## Implementation

- Added `risu-form` to sanitizer allowlist
- Renamed `runLuaButtonTrigger()` to `runLuaInteractionTrigger()`
  - `runLuaButtonTrigger(char, value)` => `runLuaInteractionTrigger('button', char, value)`
  - Used for form submissions as well: `runLuaInteractionTrigger('form', char, value, formData)`

### DIFF
--- a/src/lib/ChatScreens/Chat.svelte
+++ b/src/lib/ChatScreens/Chat.svelte
@@ -4,7 +4,7 @@
     import { ColorSchemeTypeStore } from "src/ts/gui/colorscheme"
     import { longpress } from "src/ts/gui/longtouch"
     import { getModelInfo } from "src/ts/model/modellist"
-    import { runLuaButtonTrigger } from 'src/ts/process/scriptings'
+    import { runLuaButtonTrigger, runLuaFormTrigger } from 'src/ts/process/scriptings'
     import { risuChatParser } from "src/ts/process/scripts"
     import { runTrigger } from 'src/ts/process/triggers'
     import { sayTTS } from "src/ts/process/tts"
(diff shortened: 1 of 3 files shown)


cubicj (commented):
Thanks for this. `form` and `action` are already in DOMPurify's default allowlist, so a chat form navigates the webview on native submit today, and the unconditional `preventDefault()` blocks that inside the chat container. I confirmed `runLuaButtonTrigger` has no remaining references and the removed group branch was already unreachable. `pnpm check` and `pnpm build` pass locally, and `pnpm test` shows no new failures against the merge base.

One thing on the form data collection in `handleFormTriggerWithin`. `new FormData(target)` without `event.submitter` drops the submitting button's `name` and `value`, so a form with multiple submit buttons cannot tell which one fired. The `if (obj[k])` dedupe also collides with inherited keys, so an input named `toString` corrupts to `[null, "value"]` and `__proto__` is dropped, and a field whose first value is an empty string is read as absent. Could you take another look at the collection here?

enzi221:
@cubicj Applied, thank you.

cubicj (approved):
Thanks for the follow-ups. I went through 8ab0b8c8b and both points from the last round hold. `new FormData(target, event.submitter)` carries the submitting button's entry, and the null-prototype collection with `Object.hasOwn` handles inputs named `toString` or `__proto__` and no longer reads a field whose first value is an empty string as absent. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One thing I noticed while tracing the form data, not blocking. An `<input type="file">` entry passes through `JSON.stringify` on the way to `onFormSubmit`, so a File arrives in Lua as `{}` with no signal that anything was dropped.

On the feature itself, buttons are reachable from both sides through `risu-trigger` and `risu-btn`, while forms land Lua-only here, and `risu-form` plus `onFormSubmit` becomes a compatibility surface once cards use it. Personally I would prefer the two systems to keep covering the same interactions. Whether form submission should become a creator-facing surface, and whether the trigger side should reach it too, I think is kwaroran's call, and his #998 hold is still standing.

PR #1084, opened 2025-12-25
refactor: remove redundant safe structured clone

random-username-423:
# PR Checklist

# Description
Since July 2024, $state.snapshot already uses structuredClone internally
 for non-proxy objects, making the additional safeStructuredClone wrapper redundant.

See: https://github.com/sveltejs/svelte/pull/12413


random-username-423:
This change was introduced in svelte@5.0.0-next.184 on July 15, 2024. This repo uses 5.1.9 (5.46.1 after pnpm update), so there's no issue.

PR #1085, opened 2025-12-25
Fix: Assign New chatId when chat doesn't have chatId in toggleBookmark

sub-hub:
# PR Checklist

# Description
The bookmark feature uses chatId as an identifier, but this is currently optional in the Message interface.
Therefore, chats without a chatId cannot be bookmarked.

This PR enables bookmarking by assigning a new chatId to chats where the chatId is undefined when a user attempts to bookmark them.

And thank you so much for merge the bookmark feature last time.
For someone who's been waiting for this, it's practically a Christmas present.
I didn't even know it might get merged this year lol

Christmas is almost over, but Merry Christmas!


kwaroran:
🎄

PR #1102, opened 2025-12-25
fix: Use max_completion_tokens instead of max_output_tokens for OpenAI Completion API

random-username-423:
# PR Checklist
- [O] Have you checked if it works normally in all models? *Ignore this if it doesn't use models.*

# Description
  - Fixes `Unknown parameter: 'max_output_tokens'.` error when using OpenAI completion API
  - Uses `max_completion_tokens` which is the correct parameter name for the OpenAI API

  ### Changes
  - Replaced `max_output_tokens` with `max_completion_tokens` in openAI.ts
  - Simplified condition to only check `LLMFlags.OAICompletionTokens` flag

### DIFF
--- a/src/ts/process/request/openAI.ts
+++ b/src/ts/process/request/openAI.ts
@@ -409,8 +409,8 @@ export async function requestOpenAI(arg:RequestDataArgumentExtended):Promise<req
         delete body.logit_bias
     }
 
-    if(aiModel.startsWith('gpt4o1') || arg.modelInfo.flags.includes(LLMFlags.OAICompletionTokens)){
-        body.max_output_tokens = body.max_tokens
+    if(arg.modelInfo.flags.includes(LLMFlags.OAICompletionTokens)){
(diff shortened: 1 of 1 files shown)


random-username-423:
# OpenAI API Token Limit Parameters Guide

## Chat Completions API (`/v1/chat/completions`)

| Model | Parameter | Note |
|-------|-----------|------|
| GPT-4o, GPT-4, GPT-3.5, etc. | `max_completion_tokens` | `max_tokens` is deprecated |
| o-series (o1, o3, o4-mini, etc.) | `max_completion_tokens` | `max_tokens` not supported |

## Responses API (`/v1/responses`)

| Model | Parameter |
|-------|-----------|
| GPT-5, o-series | `max_output_tokens` |

## Notes

- `max_tokens` is deprecated and incompatible with o-series models
- Use `max_completion_tokens` or `max_output_tokens` for new implementations

PR #1113, opened 2025-12-27
feat: Scroll To chat in Bookmark list

sub-hub:
# PR Checklist

# Description
image

This PR adds a button that scrolls to the location of the original chat in the bookmark list.

image

While scrolling, a loading overlay appears to prevent user input.

This is the most challenging issue in this PR: when navigating to older chats, many images that weren't loaded initially start loading.
This causes scrolling to be pushed to the wrong position after moving to the original chat.

To address this, I've designated chats outside the loadPages scope as older chats and added a sufficient wait of at least 2 seconds after scrolling.

For chats within the loadPages scope, since they're already loaded, we skip this long wait and navigate directly.

If you know a better approach than this, please help.
However, even after trying methods like requestAnimationFrame, I couldn't find anything that worked better than this.


src/ts/stores.svelte.ts
```diff
@@ -42,6 +42,7 @@ export const ShowVN = writable(false)
 export const SettingsMenuIndex = writable(-1)
 export const ReloadGUIPointer = writable(0)
 export const ReloadChatPointer = writable({} as Record<number, number>)
+export const ScrollToMessageStore = writable(-1)
```
kwaroran: Could you use $state() instead of writable for new global states?

src/lib/ChatScreens/DefaultChatScreen.svelte
```diff
@@ -42,10 +43,82 @@
...
+                            if(!isAlreadyLoaded || !img.complete){
+                                img.setAttribute('loading', 'eager');
+                                imageLoadPromises.push(img.decode().catch(() => {}));
+                            }
+                        });
+                    }
+                }
+                if (imageLoadPromises.length > 0) {
+                    await Promise.race([
+                        Promise.all(imageLoadPromises),
```
kwaroran: Why wait for images to load?

src/ts/stores.svelte.ts
sub-hub: I'll give it a try.

src/lib/ChatScreens/DefaultChatScreen.svelte
```diff
@@ -42,10 +43,82 @@
...
+                } else {
+                    // It may seem inefficient, but since loadPages is always increasing, this wait only occurs when reading chats outside the current loadPages range.
+                    element.scrollIntoView({behavior: "smooth", block: "start"})
+                    await sleep(1000)
+                    element.scrollIntoView({behavior: "instant", block: "start"})
+                    await sleep(2000)
+                    element.scrollIntoView({behavior: "instant", block: "start"})
+                }
+
+                element.classList.add('ring-2', 'ring-blue-500')
```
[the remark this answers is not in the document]
sub-hub: @kwaroran 
This code prevents the scroll from shifting, similar to the image loading code below.
Since this code might feel unusual to users, I applied a blur effect to prevent users from scrolling during the process.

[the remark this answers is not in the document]
sub-hub: @kwaroran 
After all, I'm very afraid of scrolling being pushed after moving to the chat. Is there any other way you think about it?
I haven't fully understood how Risu displays the chat yet.

sub-hub:
This is my first time getting a code review here, so I was a bit nervous. But thank you for taking a look.

I fixed the transparency issues caused by the Tailwind version upgrade and also changed ScrollToMessageStore to $state.

Now that the errors are resolved,
The remaining issues are the image loading you asked about and the logic for repeatedly scrolling to the target.
I created this logic to prevent the scroll from shifting after moving. Is there a better way to handle this?

sub-hub:
I deleted the image loading logic since it didn't seem to work...
The wait time is long, and it seems like the scroll doesn't get pushed down after moving the chat either,
but is there a better, robust fallback?

PR #1117, opened 2025-12-27
refactor: build configuration for Svelte 5 and Tauri v2

random-username-423:
# PR Checklist

# Description
Clean up and modernize `tsconfig.json`, `vite.config.ts`, and `package.json` for Svelte 5 and Tauri v2 compatibility.

---

### 1. Group options by category

Added section comments to organize compiler options:

```json
{
  "compilerOptions": {
    /* Language & Environment */
    "target": "es2023",
    "module": "es2022",
    "useDefineForClassFields": true,

    /* Bundler Mode - Vite handles the actual build */
    "noEmit": true,
    "moduleResolution": "bundler",
    "isolatedModules": true,
    "verbatimModuleSyntax": true,

    /* Module Resolution */
    ...

    /* Type Checking */
    ...

    /* Debugging */
    ...
  }
}
```

### 2. Add `noEmit: true`

```diff
+ "noEmit": true,
```

Since Vite handles transpilation, TypeScript is only used for type checking. This option prevents TypeScript from emitting output files.

This also resolves the following errors that occurred when TypeScript tried to emit `.js` files that already exist:

```
Cannot write file 'public/sw.js' because it would overwrite input file.
Cannot write file 'src/ts/rpack/rpack_bg.js' because it would overwrite input file.
```

### 3. Remove `*.d.ts` from include

```diff
- "include": ["src/**/*.d.ts", "src/**/*.ts", "src/**/*.js", "src/**/*.svelte", "public/sw.js"],
+ "include": ["src/**/*.ts", "src/**/*.js", "src/**/*.svelte", "public/sw.js"],
```

`*.d.ts` files are already covered by `*.ts` pattern.

### 4. Remove non-existent `risu-electron` from exclude

```diff
- "exclude": ["src/**/web/*.ts", "risu-electron/**/*"],
+ "exclude": ["src/**/web/*.ts"],
```

The `risu-electron` folder does not exist in the project.

### 5. Remove unnecessary `ignoreDeprecations`

```diff
- "ignoreDeprecations": "5.0",
```

This option was no longer needed.

---

### Remove unused scripts

```diff
- "updatePlugin": "tsc public/pluginApi.ts",
- "sync": "node electron/sync",
- "electron": "node electron/dist/electron",
```

- `updatePlugin`: No longer needed
- `sync`, `electron`: risu-electron folder doesn't exist

---

### 1. Replace `sveltePreprocess` with `vitePreprocess`

```diff
- import { svelte } from "@sveltejs/vite-plugin-svelte";
- import {sveltePreprocess} from "svelte-preprocess";
+ import { svelte, vitePreprocess } from "@sveltejs/vite-plugin-svelte";

  svelte({
-   preprocess: [
-     sveltePreprocess({
-       typescript: true,
-     }),
-   ],
+   preprocess: vitePreprocess(),
  }),
```

Svelte 5 recommends using `vitePreprocess` from `@sveltejs/vite-plugin-svelte` instead of the deprecated `svelte-preprocess` package.

From [svelte-preprocess README](https://github.com/sveltejs/svelte-preprocess?tab=readme-ov-file#when-to-use-it):

> If you're using Vite and are using TypeScript or common CSS preprocessors (PostCSS, SCSS, Less, ...), use the preprocessor that comes with vite-plugin-svelte instead. If these cases don't apply to you, use svelte-preprocess.

#### Why `vitePreprocess()` without `script: true`?

`vitePreprocess({ script: true })` is needed when using advanced TypeScript syntax inside `.svelte` files that requires code transformation, such as:
- Visibility modifiers (`private`, `public`, `protected`, `readonly`)
- Constructor parameter properties: `constructor(private data: Uint8Array)`
- Enums

Svelte 5's native TS support only handles type-only syntax (types that get erased). For code-transforming syntax, you need the preprocessor with `script: true`.

However, this codebase:
- Uses these advanced TS features **only in `.ts` files** (which Vite compiles separately)
- Does **not** use classes or visibility modifiers inside `.svelte` files

Therefore, `vitePreprocess()` without `script: true` is sufficient.

### 2. Update environment variable for Tauri v2

```diff
- // to make use of `TAURI_DEBUG` and other env variables
- // https://tauri.studio/v1/api/config#buildconfig.beforedevcommand
+ // to make use of `TAURI_ENV_DEBUG` and other env variables
+ // https://v2.tauri.app/reference/environment-variables/
```

Tauri v2 renamed `TAURI_DEBUG` to `TAURI_ENV_DEBUG`.

From [Tauri v2 Environment Variables](https://v2.tauri.app/reference/environment-variables/):

> TAURI_ENV_DEBUG — true for dev command or build --debug, false otherwise.

### 3. Fix environment variable string comparison

```diff
- minify: process.env.TAURI_DEBUG ? false : 'esbuild',
- sourcemap: !!process.env.TAURI_DEBUG,
+ minify: process.env.TAURI_ENV_DEBUG === 'true' ? false : 'esbuild',
+ sourcemap: process.env.TAURI_ENV_DEBUG === 'true',
```

`process.env` values are always `string | undefined`. The previous code had a bug where `"false"` (string) would be truthy:

```js
!!"false"  // true (bug!)
"false" === 'true'  // false (correct)
```

### 4. Update build target to `baseline-widely-available`

```diff
- target:['safari15','chrome100','firefox100','edge100'],
+ target:'baseline-widely-available',
```

`baseline-widely-available` targets features that have been available for 30+ months across all major browsers.

From [Baseline Supported Browsers](https://web-platform-dx.github.io/web-features/supported-browsers/):

| Browser | Version | Release date |
|---------|---------|--------------|
| Chrome | 111 | 2023-05-09 |
| Edge | 111 | 2023-05-09 |
| Firefox | 114 | 2023-06-06 |
| Safari | 16.4 | 2023-04-11 |

This is more recent than the previous explicit targets (`safari15`, `chrome100`, `firefox100`, `edge100`).

### 5. Remove non-existent `modules` alias

```diff
  resolve:{
    alias:{
      'src':'/src',
-     'modules': '/modules'
    }
  },
```

The `modules` folder does not exist in the project.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,7 +10,6 @@
     "moduleResolution": "bundler",
     "isolatedModules": true,
     "verbatimModuleSyntax": true,
-    "types": ["vite/client"],
 
     /* Module Resolution */
     "baseUrl": ".",
@@ -23,7 +22,6 @@
 
     /* Type Checking */
     "strict": false,
-    "strictNullChecks": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
 


random-username-423:
Maybe baseline-widely-available can vary depending on the Vite version, but it is still more recent than the explicitly written values in the previous code: `['safari15','chrome100','firefox100','edge100']`

Vite main (v8.0.0-beta.5) docs
https://main.vitejs.dev/config/build-options
> Widely Available on 2026-01-01. Specifically, it is ['chrome111', 'edge111', 'firefox114', 'safari16.4'].

Vite v7 docs
https://v7.vite.dev/config/build-options
> Widely Available on 2025-05-01. Specifically, it is ['chrome107', 'edge107', 'firefox104', 'safari16'].

kwaroran:
> 2. Add noEmit: true

`noEmit: false` is required due to an unknown typescript or vite bug that fails Risuai builds in specific environments.

> 4. Update build target to baseline-widely-available

Although Risuai follows the baseline for most cases, we use Core.js or feature detection so that older browsers can use Risuai, too. However, for build targets, as there is no way to polyfill changes to syntax, we use a version that is "a bit later than the baseline", so that older browsers can still use it.

> 4. Remove non-existent risu-electron from exclude

This script is needed to monkeypatch the old build runtime, which was made during the failed Electron version development.

random-username-423:
> > 4. Update build target to baseline-widely-available
> 
> Although Risuai follows the baseline for most cases, we use Core.js or feature detection so that older browsers can use Risuai, too. However, for build targets, as there is no way to polyfill changes to syntax, we use a version that is "a bit later than the baseline", so that older browsers can still use it.

@kwaroran In my opinion, it is negligible

### before: 90.7%
before

### vite 7 (baseline-widely-available): 89.9%
vite7-baseline

### vite 8 someday (baseline-widely-available): 88.6%
vite8-baseline

evaluated from https://browsersl.ist/

note: samsung internet, opera, uc browser are not included

PR #1129, opened 2025-12-28
feat: hide jailbreak toggle when no jailbreak prompt exists

random-username-423:
# PR Checklist

# Description
- Hide the "Jailbreak Toggle" button in the sidebar when no jailbreak prompt is defined in the current preset

## Changes

**`src/lib/SideBars/Toggles.svelte`**
- Added `hasJailbreakPrompt` derived state that checks if `promptTemplate` contains any item with `type === 'jailbreak'`
- Wrapped the jailbreak toggle checkbox in both container variants with `{#if hasJailbreakPrompt}` condition

## Why

The jailbreak toggle button was always visible in the sidebar even when the user's prompt template didn't contain any jailbreak prompts. This was confusing because toggling it had no effect. Now the button only appears when it's actually useful.


random-username-423:
thanks for telling it

kwaroran:
Wouldn't checking template all the time at render cause performance issues?

PR #1139, opened 2025-12-30
refactor: Remove hardcoded maxContextTokens limits for legacy models

random-username-423:
# PR Checklist

# Description
In the early days of RisuAI, `maxContextTokens` was hardcoded per model to prevent exceeding model-specific context limits. This code is now obsolete since:

1. These model-specific limits are no longer necessary
2. Users should be able to configure context limits freely through settings
3. The hardcoded values were outdated and didn't reflect current model capabilities

## Changes
- `src/ts/process/index.svelte.ts`: Removed 22 lines of conditional maxContextTokens capping logic

### DIFF
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -1567,32 +1567,6 @@ function assignIds() {
 }
 
 /**
- * Gets the maximum context length for a given model.
- * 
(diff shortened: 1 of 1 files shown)


random-username-423:
I also removed `getModelMaxContext` unused function from `globalApi.svelte.ts`

PR #1141, opened 2025-12-30
Feat: Add Echo model for Easy Development

sub-hub:
# PR Checklist

# Description
Recently, while developing auto scroll to new message features, I wasted a significant number of tokens unnecessarily just to receive new chat messages for testing the feature.

Currently, to receive new-dummy responses in Risu, you must run an OpenAI-compatible server on your own computer and enter its URL into Risu to receive responses.
This is very annoying because if you test it on Web, it even gets stuck with the cors problem.

This PR adds a model that allows you to receive test responses without running a separate server.
The model has `recommended: false`, so it won't be visible to the general public.

I believe this model is excellent not only for developers editing Risu code but also for content creators building bots and modules.

image
image
image
image


sub-hub:
Please don't tell me to use triggers or Lua.
Just coding the features and fixing bugs already leaves me with too much to think about.
And if I use triggers, I won't be able to completely eliminate my worries about whether my functionality actually works in real-world situations.

sub-hub:
Thank you! 
I'm going to make more utilization examples on this and let the creators know.

PR #1163, opened 2026-01-04
fix: resolve syntax and import errors in multiple files

SameDesu123:
# PR Checklist

# Description

## Summary
Fix syntax and import errors that were causing build failures.

### 1. Fix missing closing brace in DefaultChatScreen.svelte
- The `scrollToBottom()` function was missing its closing `}`, causing a syntax error at line 502.

### 2. Fix `Arcodion` typo to `Accordion`
- **Files affected**: CustomModelsSettings.svelte, BanCharacterSetSettings.svelte
- The import and component usage was referencing `Arcodion.svelte` which doesn't exist. The correct file is Accordion.svelte.

### 3. Fix incorrect import paths for `isNodeServer` and `isTauri`
- **Files affected**: SettingsExportButtons.svelte, advancedSettingsData.ts, bootstrap.ts
- These were importing from `src/ts/globalApi.svelte` which doesn't export them. Changed to import from `src/ts/platform` where they are actually exported.

### DIFF
--- a/src/lib/ChatScreens/DefaultChatScreen.svelte
+++ b/src/lib/ChatScreens/DefaultChatScreen.svelte
@@ -53,6 +53,8 @@
 
     function scrollToBottom() {
         chatsInstance?.scrollToLatestMessage();
+    }
+
     $effect(() => {
         if(ScrollToMessageStore.value !== -1){
             const index = ScrollToMessageStore.value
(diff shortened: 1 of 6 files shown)


SameDesu123:
It seems to be solved in 69eafaf61ba0e4289ec843b51df96cd54c459208.

PR #1188, opened 2026-01-09
ci: Add CI workflow for linting and type checking

concertypin:
# PR Checklist

# Description
> [!NOTE]
> The action will fail, but it's due to pre-existing error, which is out of scope. It can be ignored safely.

This Pull Request introduces an automated Continuous Integration (CI) workflow and establishes a robust environment for measuring and reporting test coverage using Vitest. These changes aim to enhance code quality, maintainability, and project stability.

#### 1. Automated CI Workflow (`.github/workflows/ci.yaml`)
-   Automated Verification: Implemented a workflow that automatically executes type checking (`pnpm run check`) and unit tests (`pnpm run test`) upon Pull Requests to the `main` branch or manual triggers.
-   Coverage Status Reporting: Added a reporting mechanism that calculates line and statement coverage. The workflow updates the GitHub Commit Status with a "Vitest Coverage" check, providing immediate feedback on code coverage directly within the PR interface.

> [!NOTE]
> Due to GitHub Actions' restriction, coverage reporting is not work on fork repositories.

#### 2. Test Coverage Configuration
-   Vitest Integration: Integrated `@vitest/coverage-v8` to enable high-performance coverage reporting.
-   Configuration Updates:
    -   Added `test:coverage` script to `package.json`.
    -   Configured the `v8` provider and defined exclusion paths in `vitest.config.ts`.
    -   Added the `coverage/` directory to `.gitignore` to prevent tracking of local report artifacts.

#### 3. Project Environment Standardization
-   Engine Specification: Formally specified Node.js version `>20` in `package.json` to ensure environment consistency.
-   Package Manager Locking: Fixed the `pnpm` version to `10.17.1` to avoid discrepancies between local development and CI environments.

#### Test Coverage Threshold
The current success threshold for code coverage is **temporarily set to 20%** as an initial baseline.
-   Recommendation: As the test suite expands, it is highly recommended to periodically update the `ci.yaml` file to incrementally raise this threshold, ensuring a continuous improvement in code quality.

#### Branch Protection Rules
To fully leverage this CI implementation, it is strongly advised to configure **GitHub Branch Protection Rules** for the `main` branch.
-   Recommendation: Enable the **"Require status checks to pass before merging"** option and specify the `ci` job as a required check. This will programmatically prevent faulty or unverified code from being merged into the production branch.checked and by keeping dependencies up to date.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,13 +6,13 @@ on:
     branches:
       - main
   workflow_dispatch:
-permissions:
-  contents: read
 
 jobs:
   ci:
     timeout-minutes: 15
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -92,10 +92,12 @@ jobs:
             const description = `Lines: ${total}% | Statements: ${summary.total.statements.pct}%`;
 
             // 3. Github PR checklist update
+            const sha = context.payload.pull_request ? context.payload.pull_request.head.sha : context.sha;
+
             await github.rest.repos.createCommitStatus({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              sha: context.sha,
+              sha: sha,
               state: color, // Build is success
               context: 'Vitest Coverage', // List name
               description: description, // e.g. "Lines: 85%"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -73,6 +73,7 @@ jobs:
       - name: Update Coverage Status
         if: always()
         uses: actions/github-script@v7
+        continue-on-error: true
         with:
           script: |
             // 1. read coverage summary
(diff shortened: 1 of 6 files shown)


cubicj (commented):
I checked this against the current main and the other open CI PRs before going through the diff.

The check and test portion overlaps #1474, which adds a pr-check workflow with svelte-check, a build, and vitest. That PR is already approved and waiting on a second review, so the plain CI part is likely to land through it first. The environment pins have also moved since January. Current main pins `packageManager` to `pnpm@10.34.1` and sets `engines.node` to `^20.19.0 || >=22.12.0`, so this branch would roll both back, and those files plus the lockfile are exactly where the merge conflict sits.

The coverage setup is the part nothing else covers. Two things stood out when I traced how it would behave here. The commit status only works from branches on this repo, as you noted, so fork PRs, which are most contributions here, never see the number. And the coverage job repeats the full install and test run the ci job just did, so every PR runs the suite twice.

Given all that, the coverage tooling seems like the part of this PR that still adds something new. Would you be interested in continuing it in that direction on top of the current main?

Issue #802, opened 2025-04-01
Inlay image generation


hboxgames:
Hi !!

First, Thank you for all the great work on this app.

I have been trying to make the image generation work correctly. For now, I just managed to make it appear and disappear.
Since I wanted to keep it inside the chat, I saw that there is a function called 'inlay'. which doesn't work right now.

I noticed that you replied that tauri thing changed and you can't fix it.

But I'm new here, so it's not about old images, but the ones to come.
Will you simply abandon the inlay function, or implement something new to make us use it again for newly generated images ??

Oh, and also.. I thought it would be amazing to be able to generate image manually, not only after the text (and therefore re-generate multiple times if needed), this is something we can do in sillyT but I really love Risu so far so hope it can be done !!

Thank you so much !!

cubicj:
Inlay rendering was fixed in PR #1209 (2026-01). Manual image re-generation is not yet supported — keeping open as enhancement.

PR #1195, opened 2026-01-10
Fix imggen inlay rendering by wrapping stored image IDs

joonsoome:
## Changes
This fixes return value couldn't wrapped the inlay rendering where the generated image ID was printed as plain text in chat. especially with ComfyUI image returns. I now wrap writeInlayImage results with {{inlayed::...}}, which the parser recognizes and renders as an inlay image.

### Testing
Manual: triggered /gen with inlay screenenabled (lorebook, character setting); image renders instead of UUID.

### Result
스크린샷 2026-01-10 18-15-21

### Related Issue
https://github.com/kwaroran/RisuAI/issues/802

### DIFF
--- a/src/ts/process/inlayScreen.ts
+++ b/src/ts/process/inlayScreen.ts
@@ -26,7 +26,7 @@ export function runInlayScreen(char:character, data:string):{text:string, promis
                                 const imgHTML = new Image()
                                 imgHTML.src = v
                                 const inlay = await writeInlayImage(imgHTML)
-                                return inlay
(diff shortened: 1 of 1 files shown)


joonsoome:
> What happened to the commit history? From this link, I think the change history is 1 line. How about close it and reopen it? [main...joonsoome:Risuai:main](https://github.com/kwaroran/Risuai/compare/main...joonsoome:Risuai:main)
> 
> ```diff
> - return inlay
> + return `{{inlayed::${inlay}}}`
> ```

Oh, my mistake I will put the branch.

PR #1204, opened 2026-01-12
refactor: Cache getAssetSrc/getEmoSrc results

enzi221:
# PR Checklist

# Description

## Motivation

If you run a measurement, the waterfall which `ChatBody` draws is far from ideal.

image

And the bulk of it is caused by repeated calls to `getAssetSrc()`, taking more than 60%. Since those `get$2()` are also caused by `getAssetSrc()`, it really is a bottleneck.

image

## Implementation

~~This is somewhat lame implementation, but I'm caching `getAsserSrc()` return values. It's the most quickest way to resolve current performance issue with asset-heavy chars/modules, without major rewrites.~~

image

~~Now the same measurement only takes less than 200ms, 70% faster.~~

~~The cache is invalidated when `ReloadGUIPointer` changes, as usual.~~

New implementation: Still caches the result, but will react to any asset changes realtime, rather than `ReloadGUIPointer`. Performance measurements are similar to the above image.

### DIFF
--- a/src/ts/parser.svelte.ts
+++ b/src/ts/parser.svelte.ts
@@ -1,7 +1,7 @@
 import DOMPurify from 'dompurify';
 import markdownit from 'markdown-it'
 import { appVer, getCurrentCharacter, getDatabase, type Database, type character, type customscript, type groupChat, type triggerscript } from './storage/database.svelte';
-import { DBState } from './stores.svelte';
+import { DBState, selIdState } from './stores.svelte';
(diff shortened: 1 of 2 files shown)


kwaroran:
The reason it's not currently cached is because there are many cases where the asset list can change, such as characters changing, modules activating, assets being added, etc.
However, this PR doesn't seem to handle any of those cases, which might create bad UX.

enzi221:
Well, I'll see what I can do about asset lists changing. Or are you working on the optimization yourself?

PR #1200, opened 2026-01-11
fix: imggen inlay rendering by wrapping stored image IDs

joonsoome:
## Summary
This PR rebuilds the inlay-asset lifecycle changes in a clean, non-duplicated way while respecting upstream updates.

## What’s included
1. Metadata-backed inlay asset storage with legacy migration
2. Orphan inlay cleanup on message/chat/character deletion flows
3. Backup coverage alignment for inlay media assets
4. Chat ZIP export/import with inlay media bundling

## Commit structure
- `7efba86d` feat(inlay): move assets to metadata-backed storage with legacy migration
- `829735bb` feat(inlay): clean orphan assets when chats/messages are deleted
- `1bb072fc` feat(backup): include inlay media and legacy inlay metadata in local backups
- `3e0dbf4a` feat(chat-zip): bundle inlay media in chat zip export/import

## Validation
- `pnpm check` passed (`svelte-check found 0 errors and 0 warnings`)

## Notes
- Scope is limited to inlay lifecycle/cleanup/backup/chat-zip behavior.
- Existing unrelated functionality was intentionally left unchanged.

## ~~Summary (Outdated)~~ 
~~Fixes inlay rendering so generated IDs are wrapped and render properly, hardens the inlay pipeline, and persists inlay assets with metadata so they load consistently across devices.~~

## ~~Updates (26-02-06)~~
~~First, I apologize for continuously updating the same PR. Since this PR hasn't been merged, I should have created a separate Enhancement PR, but I had no choice but to add this update. I belatedly discovered that inlay media assets are not included when backing up individual chats or the entire chat history. Consequently, when restoring a backup after a system change or clearing the cache, the inlay assets fail to restore. Therefore, I've added the functionality below. I sincerely hope the admins will kindly approve this.~~

~~commit - 6e69abc45fc88f23da1a28c6e9a9f38d7651a906~~
~~**Enhancement: Include inlay media in backups + chat ZIP export/import**
This commit ensures inlay media (image/audio/video) is preserved across:~~

- ~~Local Full Backup/Restore (including legacy inlays still stored in localforage), and~~
- ~~Chat export/import via a new ZIP format that bundles chat JSON + inlay files.~~

~~**Key Changes**~~
- ~~Local Full Backup now includes legacy inlay assets from localforage (name: "inlay") when they are not yet present in db.inlayAssets, and injects corresponding inlayAssets metadata into the backup DB payload.~~
[...]

PR #1209, opened 2026-01-14
fix: update missing inlay tag in imggen inlay mode

luyiourwong:
# PR Checklist

# Description
This issue was a regression caused by a refactor (commit 2fee3aa7). At that time, the `writeInlayImage` function in [inlays.ts](src/ts/process/files/inlays.ts) was modified to return a raw **UUID string** instead of a formatted tag (`{{inlay::UUID}}`) to simplify the implementation of image previews in the chat input UI.

However, the `runInlayScreen` function in [inlayScreen.ts](src/ts/process/inlayScreen.ts) remained unchanged. It directly injects the return value of `writeInlayImage` into the message text. Since the return value was changed to a raw UUID, the message parser no longer recognizes it as an image asset, causing the rendering break.

### How to Reproduce
1. Set a character's **Additional Character Screen** to `Image Generation`.
2. Enable the **Inlay Screen** option on the bottom.
3. Chat with the character and trigger an output containing `<ImgGen="...">` or `{{ImgGen="..."}}`.
4. Wait for the image generation to complete.

### Behavior
*   **Actual Behavior**: After the `[Generating...]` placeholder disappears, a raw UUID string (e.g., `550e8400-e29b-41d4-a716-446655440000`) is displayed in the chat.
*   **Expected Behavior**: The generated image should be rendered within the chat message.

### Fix
Since `writeInlayImage` is now designed to return a raw ID for UI flexibility, I have updated `runInlayScreen` to manually wrap the returned ID in the required `{{inlay::...}}` tag format before performing the text replacement. This ensures the parser correctly identifies and renders the asset.

### Note to Reviewers
Since I do not have access to existing image generation providers (like NovelAI or Stable Diffusion WebUI) for testing, I performed my verification using a new provider currently under development in this branch: [`feature/chutes-image-20260114`](https://github.com/luyiourwong/RisuAI/tree/feature/chutes-image-20260114). 

While this fix works perfectly with my new implementation, I would appreciate it if someone with access to existing providers could double-check if this aligns with their current behavior.

### DIFF
--- a/src/ts/process/inlayScreen.ts
+++ b/src/ts/process/inlayScreen.ts
@@ -26,7 +26,7 @@ export function runInlayScreen(char:character, data:string):{text:string, promis
                                 const imgHTML = new Image()
                                 imgHTML.src = v
                                 const inlay = await writeInlayImage(imgHTML)
-                                return inlay
+                                return `{{inlay::${inlay}}}`
                             })())
                             return match
                         })


luyiourwong:
I noticed there is another broader PR related to inlays #1200 , but it appears to be stalled due to its complexity. My PR is intended as a **minimal, atomic fix** that specifically addresses the broken rendering path in `runInlayScreen` without introducing architectural changes or risk to asset storage. 
Since the current behavior renders the ImgGen inlay feature completely unusable (displaying raw UUIDs), this small change provides an immediate solution while the more complex refactors are being discussed.

PR #1219, opened 2026-01-16
fix: Hyper Memory V3 'ot' & 'startsWith' error Bug Resolve

joonsoome:
# Summary 1 'ot' error
Fixed the TDZ crash by initializing generationInfo before any early error paths and then filling it in place, and added inlay-token sanitization for HypaV3 summaries so {{inlay::...}} never reaches the summarizer. The TDZ fix is in index.svelte.ts, and the summary sanitation is in hypav3.ts.

index.svelte.ts now initializes generationInfo up front and uses Object.assign when the real token/model data is available, so throwError can safely reference it during HypaV3 failures.
hypav3.ts replaces inlay tokens with [Image] when building the summary prompt, covering both auto-summarize and re-summarize flows.

## Resolve
Prevent generationInfo TDZ and sanitize

## Resolve 2 'startWith' error
- Initialize generationInfo early and seed the model string to avoid TDZ errors during early error paths.
- Sanitize inlay tokens in HypaV3 summary/re-summarize prompts.
- Guard getModelInfo against missing IDs to prevent chat UI crashes.

## Stack Trace 1
스크린샷 2026-01-14 02-48-36

## Stack Trace 2 
Cannot read properties of undefined (reading 'startsWith')
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at  (../../src/ts/model/modellist.ts:547:10)
    at  (../../src/lib/ChatScreens/Chat.svelte:303:32)
    at fn (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/runtime.js:264:15)
    at  (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/reactivity/deriveds.js:342:11)
    at  (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/reactivity/deriveds.js:356:13)
    at  (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/runtime.js:629:3)
    at Array.map ()
    at  (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/reactivity/effects.js:368:50)
    at fn (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/runtime.js:264:15)
    at  (../../node_modules/.pnpm/svelte@5.46.1/node_modules/svelte/src/internal/client/runtime.js:439:17)
    
## Validation
- [x] Reproduce the inlay + HypaV3 summary scenario and confirm the Cannot access 'ot' error is gone.
- [x] Run pnpm check to ensure TypeScript stays clean.

### DIFF
--- a/src/ts/model/modellist.ts
+++ b/src/ts/model/modellist.ts
@@ -531,9 +531,23 @@ for(let i=0; i<LLMModels.length; i++){
     }
 }
 
-export function getModelInfo(id: string): LLMModel{
+export function getModelInfo(id?: string | null): LLMModel{
 
     const db = getDatabase()
+    if(!id){
+        return {
+            id: '',
+            name: 'Unknown',
+            shortName: 'Unknown',
+            fullName: 'Unknown',
+            internalID: '',
+            provider: LLMProvider.AsIs,
+            format: LLMFormat.OpenAICompatible,
+            flags: [],
+            parameters: OpenAIParameters,
+            tokenizer: LLMTokenizer.Unknown
+        }
+    }
     const found:LLMModel = safeStructuredClone(LLMModels.find(model => model.id === id))
 
     if(found){
--- a/src/ts/process/index.svelte.ts
+++ b/src/ts/process/index.svelte.ts
@@ -102,7 +102,9 @@ export async function sendChat(chatProcessIndex = -1,arg:{
         stage4Duration: 0
     }
 
-    const generationInfo: MessageGenerationInfo = {}
+    const generationInfo: MessageGenerationInfo = {
+        model: getGenerationModelString()
+    }
     let isAborted = false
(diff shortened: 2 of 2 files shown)


sub-hub:
I got this error in current latest risu,

[plugin:vite:import-analysis] Failed to resolve import "src/ts/util/inlayTokens" from "src/ts/process/memory/hypav3.ts". Does the file exist?

It looks like we need an additional patch for this.

Additionally, the code in this pr is importing to src/ts/util/inlayTokens, which seems to **add a new folder called util**.
Currently the src/ts/util.ts file already exists and we are using it in all codes, do we need to create a new util folder?

PR #1223, opened 2026-01-16
feat: add auto image generation feature

luyiourwong:
## PR Checklist

- Required Checks

## Summary

Adds an **Auto Image Generation** feature that automatically generates and appends images to character responses when image generation mode is active. Users can enable this feature per character through a simple toggle in the character configuration panel.

Recent advances in multimodal AI models have made image generation from natural language significantly more capable and flexible. Unlike traditional approaches that required specific keywords or secondary prompt engineering, modern image models can directly process natural language descriptions. This PR leverages that capability to eliminate manual image generation steps—when enabled, images are automatically generated from character responses without requiring additional prompt manipulation or user intervention.

## Related Issues

None

## Changes

- **Language files**: Added "Auto Image Generation" translation strings across all supported languages (EN, ZH, ZH-Hant, KO)
- [`src/lib/SideBars/CharConfig.svelte`](src/lib/SideBars/CharConfig.svelte): Added `CheckInput` control for `autoImgGen` toggle in character configuration UI
- [`src/ts/characterCards.ts`](src/ts/characterCards.ts): Updated character data parsing to handle `autoImgGen` property in character import/export logic; added initialization in character creation functions
- [`src/ts/storage/database.svelte.ts`](src/ts/storage/database.svelte.ts): Extended character database schema with `autoImgGen?: boolean` property
- [`src/ts/process/index.svelte.ts`](src/ts/process/index.svelte.ts): Integrated `generateAIImage()` and `writeInlayImage()` functions; added automatic image generation logic in chat processing pipeline (stage 3), the result will append into chat in Inlay mode (`<Thoughts>` tags excluded); wrapped in try-catch to prevent chat interruption if image generation fails

### Character Data Structure Changes
The character card structure has been extended with the `autoImgGen` boolean property. Existing characters will be initialized with `autoImgGen: false` by default, ensuring backward compatibility. No manual migration is required.

### Chat Processing Pipeline (Stage 3)
New image generation logic has been added to the stage 3 processing phase after the main response generation. This adds a potential async operation that may slightly increase response time when auto image generation is enabled. The operation respects abort signals and preview modes, so it won't interfere with preview generation or cancelled requests.

## Example
Test on Qwen-Image-Edit and FLUX.2-klein with character image, they are good at full context generate.
Screenshot 2026-01-16 220249

### DIFF
--- a/src/lang/cn.ts
+++ b/src/lang/cn.ts
@@ -879,6 +879,7 @@ export const languageChinese = {
     "joinMultiUserRoom": "加入多用户聊天室",
     "exactTokens": "精确 Tokens",
     "fixedTokens": "估算 Tokens",
+    "autoImgGen": "自动图像生成",
     "inlayViewScreen": "内嵌窗口",
     "imgGenPrompt": "图像生成提示词",
     "imgGenNegatives": "图像生成负面提示词",
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -976,6 +976,7 @@ export const languageEnglish = {
     joinMultiUserRoom: "Join MultiUser Room",
     exactTokens: "Exact Tokens",
     fixedTokens: "Approximate Tokens",
+    autoImgGen: "Auto Image Generation",
     inlayViewScreen: "Inlay Screen",
     imgGenPrompt: "Image Generation Prompt",
     imgGenNegatives: "Image Generation Negative Prompt",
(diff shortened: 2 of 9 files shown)


luyiourwong:
While this direct approach works well for natural language models (like the one in https://github.com/kwaroran/Risuai/pull/1225), it lacks support for older models that rely on keyword-based prompting (like SD) or users who prefer an intermediate LLM pass to refine the prompt.

I plan to rework this feature to be more modular, allowing users to choose the generation strategy. I will submit a new PR once the design allows for this flexibility.

PR #1229, opened 2026-01-17
support PWA

nevaeh5379:
## PR Checklist

- Required Checks
    - N/A - This PR does not modify model behavior
- [x] If your PR is highly ai generated[^2], check the following:
    - [x] Have you understanded what the code does?
    - [x] Is is not a huge change?

## Summary

Enable PWA (Progressive Web App) installation support for browsers including Samsung Internet. Users can now install RisuAI as a standalone app on both desktop and mobile devices.

## Related Issues

None

### `public/manifest.json`
- Added `short_name`, `description`, `scope` for PWA installability
- Changed `display` to `fullscreen` for immersive app experience
- Added `maskable` purpose to 512x512 and 192x192 icons for Android adaptive icons
- Fixed `share_target` accept format for Samsung Internet compatibility

### `public/sw.js`
- Added `install` and `activate` event handlers for PWA lifecycle
- Added navigation request handling with network-first strategy
- Added cross-origin request filtering

### `index.html`
- Added Apple-specific meta tags for iOS Safari PWA support:
  - `apple-touch-icon`
  - `apple-mobile-web-app-capable`
  - `apple-mobile-web-app-status-bar-style`
  - `apple-mobile-web-app-title`

## Impact

- **Positive**: Users can now install RisuAI as a standalone app from Chrome, Edge, Samsung Internet, and Safari
- **No breaking changes**: All existing functionality remains unchanged
- **Fullscreen mode**: When installed as PWA, the app runs in fullscreen without browser UI

## Additional Notes

- Samsung Internet requires specific manifest format - some advanced features like `share_target` with certain formats may cause installation failures
- The `share_target` feature allows users to share `.charx`, `.risup`, `.risum` files directly to RisuAI from other apps
- The `file_handlers` feature enables opening these file types directly with the installed PWA on desktop

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,6 +6,12 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/logo_32.png" />
     <link rel="icon" type="image/png" sizes="256x256" href="/logo_256.png" />
     <link rel="manifest" href="/manifest.json" />
+    <!-- PWA iOS Support -->
+    <link rel="apple-touch-icon" href="/logo_192.png" />
+    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="default" />
+    <meta name="apple-mobile-web-app-title" content="RisuAI" />
+
     <link rel="stylesheet" href="/src/styles.css" />
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.27/dist/katex.min.css" integrity="sha384-Pu5+C18nP5dwykLJOhd2U4Xen7rjScHN/qusop27hdd2drI+lL5KvX7YntvT8yew" crossorigin="anonymous">
     <meta name="description" content="An AI frontend for both light and core users.">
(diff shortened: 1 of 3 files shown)


cubicj (commented):
The share_target action change looks like the real fix in this PR. sw.js has handled `/sw/share` since PWA support first landed while the manifest pointed at `/receive-files/`, so shared files never reached that handler.

But the new navigation handler breaks the same flow again. A share target launch is a navigation, so the POST to `/sw/share` hits the `request.mode === 'navigate'` branch before the `/sw/` switch and gets forwarded to the network, where nothing serves it. The existing share handler and the `#share_character` redirect never run. Connecting the flow also exposes an existing issue in that handler, `registerCache()` passes the relative `/sw/share/...` string to `new URL()` without a base, which throws before the cache write, so the share still fails even past the navigation branch.

The accept change also moves away from the spec. `accept` takes a string or an array of MIME types and file extensions, and the object form here is the `file_handlers` syntax. Is there a Samsung Internet source for it? The manifest also renames the app back to `RisuAI` while main standardized on `Risuai` in `e6ee605cd`, and drops `theme_color`.

Could you take another look at the whole share path in sw.js and the share_target format, and rebase on the current main?

PR #1233, opened 2026-01-17
perf: replace state snapshots with DB proxy layer for saveDb performance

kangjoseph90:
## PR Checklist

- Required Checks
    - [X] Have you added type definitions?
    - [X] Have you tested your changes?
    - [X] Have you checked that it won't break any existing features?
- [ ] If your PR uses models[^1], if true, check the following:

## Summary

Improve database save performance by introducing a proxy-based change detection layer, replacing the snapshot-based approach in `saveDb()` with event-driven change tracking.

## Related Issues

None

## Changes

- Introduced `onDatabaseUpdate()` API in `database.svelte.ts` for reactive database mutation detection.
  The callback receives `DatabaseUpdateInfo` containing `path`, `value`, `oldValue` and `type`
- Refactor `saveDb()`  to manage `changeTracker` using event-based callbacks from `onDatabaseUpdate()` instead of `$state.snapshot` operations.

## Impact

- Snapshot operation on large chat histories are expensive (~1s for 7MB). This refactor reduces change detection complexity from `O(n)` snapshot scanning to `O(1)` reactive event handling 
- The event-based approach is more robust than snapshot-based detection. For example, it now detects changes on unselected characters, which the previous approach missed.

## Additional Notes

Tested on various edge cases, including character/chat/message/prompt/module addition and deletion. Further testing may be beneficial.

### DIFF
--- a/src/lib/Others/WelcomeRisu.svelte
+++ b/src/lib/Others/WelcomeRisu.svelte
@@ -73,7 +73,7 @@
     $effect.pre(() => {
         if(step === 10){
             setTimeout(() => {
-                DBState.db = setPreset(DBState.db, prebuiltPresets.OAI2)
+                setPreset(DBState.db, prebuiltPresets.OAI2)
                 DBState.db.textTheme = 'highcontrast'
                 updateTextThemeAndCSS()
 
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -425,6 +425,8 @@ export async function saveDb() {
             }
             saveTimeoutExecute()
         })
+        // Persist bootstrap mutations emitted before the database listener existed.
+        saveTimeoutExecute()
 
         return () => {
             unsubscribeSel()
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -314,19 +314,32 @@ export async function saveDb() {
 
     const createChangeTracker = (): toSaveType => ({
         character: [],
-        chat: [],
         botPreset: false,
         modules: false,
         loadouts: false,
         plugins: false,
         pluginCustomStorage: false
     })
+    const cloneChangeTracker = (tracker: toSaveType): toSaveType => ({
+        ...tracker,
+        character: [...tracker.character]
+    })
+    const mergeChangeTrackers = (target: toSaveType, source: toSaveType) => {
+        for (const characterId of source.character) {
+            if (!target.character.includes(characterId)) {
+                target.character.push(characterId)
+            }
+        }
+        target.botPreset ||= source.botPreset
+        target.modules ||= source.modules
+        target.loadouts ||= source.loadouts
+        target.plugins ||= source.plugins
+        target.pluginCustomStorage ||= source.pluginCustomStorage
+    }
     let changeTracker = createChangeTracker()
+    let savetrys = 0
 
     let encoder = new RisuSaveEncoder()
-    await encoder.init(getDatabase(), {
-        compression: forageStorage.isAccount
-    })
 
     $effect.root(() => {
 
@@ -348,57 +361,14 @@ export async function saveDb() {
             }, debounceTime);
         }
 
-        type TrackableChat = { id?: string }
-        type TrackableCharacter = { chaId?: string, chatPage?: number, chats?: TrackableChat[] }
-
-        function parseArrayIndex(value: unknown): number | undefined {
-            if (typeof value === 'number') {
-                return value
-            }
-            return typeof value === 'string' && /^\d+$/.test(value) ? Number(value) : undefined
-        }
-
-        function isTrackableChat(value: unknown): value is TrackableChat & { id: string } {
-            return typeof value === 'object' && value !== null && 'id' in value && typeof value.id === 'string'
-        }
-
-        function isCharacterForSaveTracking(value: unknown): value is TrackableCharacter & { chaId: string } {
-            return typeof value === 'object' && value !== null && 'chaId' in value && typeof value.chaId === 'string'
-        }
-
-        function getTrackableCharacters(value: unknown): TrackableCharacter[] {
-            return Array.isArray(value) ? value.filter(isCharacterForSaveTracking) : []
-        }
-
-        function getTrackableChatIds(value: unknown): string[] {
-            return Array.isArray(value) ? value.filter(isTrackableChat).map((chat) => chat.id) : []
-        }
-
-        function updateChangeTrackerForCharacter(
-            char: TrackableCharacter | null | undefined,
-            options: { affectedChatIds?: (string | undefined)[], fallbackToCurrentChat?: boolean } = {}
-        ) {
-            if (!char?.chaId) {
+        const unsubscribeDb = onDatabaseUpdate((info) => {
+            if (info.path.length === 0) {
+                requiresFullEncoderReload.state = true
+                savetrys = 0
+                saveTimeoutExecute()
                 return
             }
-            if (!changeTracker.character.includes(char.chaId)) {
-                changeTracker.character.unshift(char.chaId)
-            }
-            const affectedChatIds = options.affectedChatIds?.filter((id): id is string => Boolean(id)) ?? []
-            if (affectedChatIds.length === 0 && options.fallbackToCurrentChat !== false) {
-                const currentChatId = char.chats?.[char.chatPage ?? 0]?.id
-                if (currentChatId) {
-                    affectedChatIds.push(currentChatId)
-                }
-            }
-            for (const chatId of affectedChatIds) {
-                if (!changeTracker.chat.some(([characterId, trackedChatId]) => characterId === char.chaId && trackedChatId === chatId)) {
-                    changeTracker.chat.unshift([char.chaId, chatId])
-                }
-            }
-        }
 
-        const unsubscribeDb = onDatabaseUpdate((info) => {
             const rootKey = info.path[0]
 
             if (rootKey === 'botPresets' || rootKey === 'botPresetsId') {
@@ -420,50 +390,39 @@ export async function saveDb() {
             }
 
             if (rootKey === 'characters') {
+                const affectedCharacters: Database['characters'] = []
+
                 if (info.path.length === 1) {
-                    for (const character of [...getTrackableCharacters(info.oldValue), ...getTrackableCharacters(info.value)]) {
-                        updateChangeTrackerForCharacter(character)
+                    const oldCharacters = info.oldValue as Database['characters'] | undefined
+                    const newCharacters = info.value as Database['characters'] | undefined
+                    affectedCharacters.push(...(oldCharacters ?? []), ...(newCharacters ?? []))
+                }
+                else if (typeof info.path[1] === 'number') {
+                    if (info.path.length === 2) {
+                        const oldChar = info.oldValue as Database['characters'][number] | undefined
+                        if (oldChar) {
+                            affectedCharacters.push(oldChar)
+                        }
                     }
-                    saveTimeoutExecute()
-                    return
+                    const char = DBState.db.characters[info.path[1]]
+                    if (char) {
+                        affectedCharacters.push(char)
+                    }
                 }
 
-                const characterIndex = parseArrayIndex(info.path[1])
-                const targetChar = characterIndex === undefined ? undefined : DBState.db.characters?.[characterIndex]
-                const fallbackChar = isCharacterForSaveTracking(info.oldValue) ? info.oldValue : undefined
-                const char = targetChar ?? fallbackChar
-                const isChatPath = info.path[2] === 'chats'
-
-                if (info.path.length === 2 && fallbackChar?.chaId !== targetChar?.chaId) {
-                    updateChangeTrackerForCharacter(fallbackChar)
+                for (const char of affectedCharacters) {
+                    if (!changeTracker.character.includes(char.chaId)) {
+                        changeTracker.character.unshift(char.chaId)
+                    }
                 }
-
-                if (isChatPath && info.path.length === 3) {
-                    updateChangeTrackerForCharacter(char, {
-                        affectedChatIds: [
-                            ...getTrackableChatIds(info.oldValue),
-                            ...getTrackableChatIds(info.value),
-                        ],
-                        fallbackToCurrentChat: false,
-                    })
-                    saveTimeoutExecute()
-                    return
-                }
-
-                const chatIndex = isChatPath ? parseArrayIndex(info.path[3]) : undefined
-                const oldChat = info.path.length === 4 && isTrackableChat(info.oldValue) ? info.oldValue : undefined
-                const affectedChatIds = chatIndex === undefined
-                    ? undefined
-                    : [targetChar?.chats?.[chatIndex]?.id, oldChat?.id]
-                updateChangeTrackerForCharacter(char, {
-                    affectedChatIds,
-                    fallbackToCurrentChat: !isChatPath,
-                })
                 saveTimeoutExecute()
                 return
             }
 
-            updateChangeTrackerForCharacter(DBState.db.characters?.[selIdState])
+            const char = DBState.db.characters[selIdState]
+            if (char && !changeTracker.character.includes(char.chaId)) {
+                changeTracker.character.unshift(char.chaId)
+            }
             saveTimeoutExecute()
         })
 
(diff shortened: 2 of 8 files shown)


cubicj (commented):
The snapshot cost problem is real, so I wanted this direction to work, but I don't think the proxy layer can reliably see mutations once Svelte's state proxy is involved.

Svelte 5's `$state` proxy keeps writes in its internal signal sources and never calls `Reflect.set` on the object it wraps. When `setDatabaseLite()` receives a plain object, the database proxy ends up inside Svelte's state proxy, and its `set`/`deleteProperty` traps become unreachable. I reproduced this with a small Node script against svelte 5.55.1, and mutations made through the state proxy produce zero `onDatabaseUpdate` events. Since `changed` is now set only from that callback, the app stops saving in that layering.

I think this looked fine in testing because bootstrap still runs the old `setDatabase(getDatabase())` idiom, in `checkCharOrder()` on this branch. That call passes the already-reactive object back in, so the database proxy lands outside the state proxy, where the traps do fire. But the layering silently flips back to the dead state whenever a plain object is applied mid-session without a reload. Loading an internal backup does exactly that, and the account backup, Risu-Kei restore, and multiuser receive-chat paths do the same. After any of those, edits including the restored data itself are no longer saved, with no error anywhere.

Whether saving works at any given moment depends on which object was last passed to `setDatabaseLite()`, which seems too fragile for the save path. One more thing to keep in mind for a rebase, current main now saves `loadouts`, `plugins`, and `pluginCustomStorage` as separate blocks, so the root-key mapping in the callback would need to cover those too. Could you take another look at whether change detection can work independently of that layering?

kangjoseph90:
Thank you for reviewing and you're right that this can break when a raw object is passed in.

I think the specific case you described could be fixed by assigning the value twice in `setDatabaseLite()`, forcing the proxy layers into the correct order:

```ts
DBState.db = data
DBState.db = createDatabaseProxy(DBState.db)
```

This would ensure that the database proxy remains outside Svelte's state proxy.

However, I realized that the approach can still be bypassed fairly easily. For example, some code could retain and mutate a reference to the original raw object instead of accessing it through `DBState.db`:

```ts
const raw = data
setDatabaseLite(data)

raw.foo = 'updated'
```

That kind of update pattern would be quite unusual in the current codebase, and I don't recall any existing call sites that work this way. Still, I agree that the proxy-based approach is fundamentally rather fragile.

This PR is also very old at this point and would need a rebase, although I think that part should be fairly quick.

What do you think about the issue above? Do you think enforcing the proxy layering would be sufficient, or would it be better to avoid the proxy-based approach altogether?

cubicj:
I tested your double assignment against svelte 5.55.1 before answering, in the same harness as last time, and it works. Svelte sees `STATE_SYMBOL` through the database proxy's `has` and stores it without re-wrapping, so the traps fire, reactivity stays intact, and re-passing an already wrapped object is a no-op.

The raw reference case I wouldn't count against the approach. I checked and main is equally blind there, a write to a retained raw object never reaches Svelte's signals either, so the UI wouldn't update even today. If you want to close it anyway, cloning the incoming object in `setDatabaseLite()` would make any retained reference point at a harmless orphan, and the plain-object paths are rare enough that the clone cost doesn't matter.

What kept me on the fence is that the layering is a convention rather than a guarantee. There are already direct `DBState.db` assignments outside `setDatabaseLite()`, in the plugin API and in `WelcomeRisu.svelte`, and they stay safe today only because they happen to assign the already wrapped object back. Any future site that assigns a fresh object kills saving silently, and this PR's own history shows how quietly that slips through testing. I also looked for an alternative in Svelte's public API and found nothing that avoids the O(size) walk, so dropping the proxy just means keeping the snapshot cost. I'd keep the direction and make the layering loud instead of assumed, whether that's `setDatabaseLite()` verifying its own wiring, a regression test around the plain-object restore path, or the clone above, any of them would do.

One thing I'd still change is wrapping only plain objects and arrays the way Svelte does, since the current version proxies everything it touches. Does that sound workable?

kangjoseph90:
I’ve updated the PR.

`DBState` and the database proxy logic have now been moved into a dedicated `databaseState.svelte.ts` module. Database replacement is only exposed through `setDatabaseLite()`, so the required proxy layering is enforced structurally rather than relying on call-site conventions.

There were three existing direct `DBState.db = ...` assignments outside the database state layer. All three turned out to be redundant reassignments of the same object, so they could be removed safely without changing behavior.

With this structure, I don’t expect callers to be able to accidentally construct the proxy layers in the wrong order.

Since this code is close to the app's core, I also added a dedicated `databaseState.svelte.test.ts` suite covering proxy layering, nested mutations, replacements, array operations, and related edge cases.

The change tracker was also updated for the current save format and strengthened around list operations, including replacement, deletion, `splice()`, and direct array truncation.

For transparency, I used AI assistance while investigating, implementing, and testing this revision.

cubicj (commented):
Thanks for the update. The new structure covers what we discussed, and `svelte-check`, `vitest`, and `vite build` all pass on the current head. I also verified the three removed `DBState.db` assignments were all self-assignments.

Two problems remain, both reproduced against svelte 5.55.1.

Element proxies keep their creation-time `path` after they move. When `removeChar()` splices `db.characters`, the shifted characters keep emitting their old index, so the `saveDb()` listener marks the wrong character and the edited one is never re-encoded. After any permanent character deletion, edits to shifted characters silently drop from the save file unless a later `sendChat` happens to re-save them. The new splice test misses this because it never edits a moved element afterwards.

Replacing the whole database emits no event, so restores no longer save. The Kei restore, the account backup load, the internal backup restore, and multiuser `receive-chat` all swap in a fresh object and previously relied on the `$effect` rerun to persist it. Restoring a backup and closing the app before any edit loses the restore.

There is also a smaller one. A failed save now discards the whole `changeTracker`, while main kept the `loadouts`, `plugins`, and `pluginCustomStorage` flags across retries.

Could you take another look at these?

kangjoseph90:
Updated. Building a path-based proxy while preserving correctness turned out to be trickier than I expected. 
The previous approach captured paths when proxies were created, which becomes incorrect after array operations.
I replaced that approach entirely. Proxies now retain only their target, parent metadata, preferred key, and child cache. When a mutation event is emitted, the current path is resolved by walking through the parent chain.

At each level, the cached preferred key is checked first. In the common case, this resolves the path in O(depth) without scanning the parent. When an operation changes a key or array index, the cache check fails and that parent is searched linearly for the same underlying target. 

The cost of one event is O(depth + Σ scanned parent keys for cache misses). In normal mutations, it remains O(depth).

Database root replacement is structurally restricted to `setDatabaseLite()`. That function now directly emits an event with `path: []`, and the save layer interprets it as requiring a full encoder reload and full save.

I also intentionally removed the `chat` field from `changeTracker`. It had no effective consumer and keeping separate chat IDs substantially increased the complexity. It can be added again later if the persistence format gains a real chat-level save operation.

The change tracker lifecycle is now:
1. At the beginning of a save, snapshot the tracker and immediately replace it with an empty tracker.
2. Mutations occurring during the save accumulate in the new tracker.
3. On success, discard the saved snapshot.
4. On failure, merge the original snapshot back into the current tracker so both the failed work and mutations received during the save are preserved for the next attempt.

cubicj (commented):
I went through the new revision on 624f87e64. Resolving paths through the parent chain at event time and confining database replacement to `setDatabaseLite()` settle all three issues from the last round, and this is a considerably more robust design than the path-baking version. On this head svelte-check, the full test suite, and a production build are clean, and I checked that `toSaveType.chat` had no consumer in the encoder before its removal.

One write class still escapes tracking. When code captures a reference from `DBState.db` and an ancestor array is later replaced by a new array sharing the same objects, writes through that reference mutate the live database but emit no event, since `resolveProxyPath` no longer reaches the active root and drops them. The chat reorder in `SideChatList.svelte` produces exactly that replacement by rebuilding `chats` from the existing chat objects. Two paths hit it today. In `sendChat`, `currentChat.message[msgIndex].data = t` after `await inlayr.promise` loses the finished inlay image when chats are reordered during generation and the debounced save completes in between, and the write-back on the next line does not rescue it because reassigning the same object is filtered by the `Object.is` check. `preLoadChat` in `coldstorage.svelte.ts` has the same window around its storage fetch, so everything it restores onto the retained `chat` goes untracked. I reproduced the mechanism in a vitest probe against this head, where the mutated value is visible through a fresh `DBState.db` read while no listener fires.

The race window is narrow, but the snapshot scan this PR replaces was immune to it, and the UI stays correct so nothing surfaces to the user when persistence misses the write. How would you like to handle retained references across container replacement?

kangjoseph90:
I was aware of this class of issue, but assumed it would not occur in practice and left it unsupported.

Fully generalized, since objects can be retained, shared, and reattached under different containers at runtime, the database structure is a dynamic DAG, and a single parent metadata entry cannot represent it correctly.

A complete solution would allow each proxy to have multiple weakly referenced incoming edges (parent, key). Proxy wrapping would also need to become eager when a subtree is attached: new objects would be wrapped recursively, while existing proxies would register the new parent edge. Replaced or removed children must unregister their old parent edges so stale paths do not accumulate.

Path resolution would then traverse the available parent edges and keep paths that reach the current active root. 
This model provides actual correctness, and the implementation itself does not seem too difficult, but I am unsure about the runtime cost of eager wrapping and potentially large path traversal.

The currently observed case is simpler: a container is replaced while directly reusing its children. I can fix that by reconciling one level of children during set and updating their parent/key metadata.

However, that would still fail if a retained proxy is wrapped inside newly created intermediate objects, such as:
```
parent.child = { nested: { deeper: retained } }
```
Should I implement the limited direct-child fix for this PR, or redesign the metadata around the full DAG model?

kangjoseph90:
Oh, I found a similar implementation in Valtio: [https://github.com/pmndrs/valtio/blob/main/src/vanilla.ts](https://github.com/pmndrs/valtio/blob/main/src/vanilla.ts)
Instead of storing weak parent references, each parent subscribes to its child’s events and prefixes its own key as the event propagates upward. This seems like a practical way to represent multiple incoming paths while cleaning them up on replacement.
I think the full solution should be quite feasible with this approach. I’ll try implementing it and report back.

cubicj:
Before you commit to the redesign, I swept the codebase to size the two options.

The dominant pattern is the simple one. Containers are replaced while reusing their children in many places, including a hot path. sendChat rebuilds the message array on every send in the chatId backfill map in `index.svelte.ts`, and reorder or delete flows do the same in `SideChatList.svelte`, `PersonaSettings.svelte`, `RegexList.svelte`, `TriggerV1List.svelte`, `LoreBookList.svelte`, `botpreset.svelte`, `command.ts`, and `bootstrap.ts`, among others. The limited direct-child reconcile covers all of these.

The only construction I found that defeats the one-level fix is in `LoreBookList.svelte`. The drag handler copies the moved entry with a spread over the live proxy, so the object-valued fields `extentions` and `loreCache` land as retained children one level deeper inside a fresh plain object. Today this has no practical victim since nothing writes `loreCache` through a retained reference, but it is the one real instance of the nested class you described.

For completeness, some bypasses are orthogonal to either option. A caller that keeps its raw object after handing it to the DB, like the V2 plugin `setChar` path in `plugins.svelte.ts`, never hits any proxy when it mutates that object later, and `Object.defineProperty` plus descriptor-value reads skip the missing traps. No first-party code hits these today, and the full DAG model would not close them either.

Given that census I would be fine with the limited direct-child fix plus a regression test around the sendChat message array rebuild. If the Valtio-style subscription model comes out clean it also works, but note it stores proxies inside the tree, so the `safeStructuredClone` and `postMessage` paths and the Svelte `$state` layering need re-verification, and attach-time wrapping brings back an O(size) walk at load. What do you think?

cubicj:
Small correction to my own note above.

I cited the V2 setChar path as the raw-reference example, but thinking about it again, V2 is deprecated anyway, and V3 doesn't have this class at all. V3 plugins run in the iframe sandbox, so every argument crosses a structured-clone boundary and the host only ever sees fresh copies. A V3 guest can't keep a reference into the live DB or touch the proxy object in the first place. So those bypasses are legacy V2 leftovers and shouldn't really weigh on the design choice here.

kangjoseph90:
I tried implementing the full subscription-based model. The code is slightly longer, but the control flow is much more straightforward, and the result is considerably more robust. I prefer this design over the single-parent metadata approach because it represents the actual DAG directly and handles retained references, shared paths, detachment, and root replacement without reconciliation or fallback scans.

The remaining question is performance. Eager wrapping adds an O(nodes + edges) traversal when attaching a subtree or replacing the root, and event propagation is proportional to the active paths. I expect the practical impact to be small for the current database shape, but I’ll verify it before committing.

The current implementation passes all focused tests, the full test suite, svelte-check, and the production build.

cubicj (commented):
I reran the retained proxy scenarios from the earlier rounds against the new subscription model and it holds up, including the reorder and `message.map` rebuild shapes. svelte-check, tests, and build pass here.

One real regression, at startup rather than in the proxy layer. Everything `bootstrap.ts` mutates before the `saveDb()` call at line 256 emits its events while `onDatabaseUpdate` has no listener, and the old effects always scheduled a first save at registration, so a session that loads and closes without any DB change now writes nothing. The concrete damage is that `makeColdDataForCharacter` strands a fresh `crypto.randomUUID()` cold item every launch since the pointer stub never persists, and `cleanChunks` does not collect those.

Minor and V2-only, take it or leave it. A frozen object handed in by a V2 plugin gets wrapped by `isProxyableDatabaseValue`, after which any read of it breaks the proxy invariant and `JSON.stringify(getDatabase())` throws, killing the save loop. Skipping frozen objects there costs nothing.

Perf on a 10k message chat looks fine. Root attach ~106ms, per-send rebuild 94-120ms vs 18-27ms on the plain Svelte proxy, streaming ~10us per token, all linear, so typical chats sit well under 10ms per send. Clear net win over the per-save snapshots.

Could you schedule one save right after the subscription registers?

kangjoseph90:
Thanks, both issues are fixed in e6dbfb5c.
saveDb() now schedules one save through the existing debounce immediately after onDatabaseUpdate() registers. 
I also excluded frozen objects and arrays from database proxy wrapping. The regression test covers reading a nested frozen object and serializing the database through the actual JSON.stringify(getDatabase()) boundary.

cubicj (approved):
I reran both fixes on e6dbfb5c4. Mutations applied before the listener exists now round-trip through an empty-tracker save with the migrated ids and the cold pointer intact, since the encoder init snapshot already carries them and the scheduled save flushes it. The frozen probes, a nested frozen object and a frozen array with object elements, throw the get invariant TypeError on the previous head e7077db3e and pass here, so ordering the `Object.isFrozen` check ahead of `Array.isArray` was a nice catch. svelte-check, the full test suite, and a production build are clean on this head.

Since this PR owns the save path, where a silent regression means user data loss, I held it to a stricter bar than usual and retraced the whole diff against main once more before signing off, the tracker routing, the save loop error paths, and the serialization boundaries. Nothing new came up. If anything the event model fixes gaps main had. Edits to unselected characters were never tracked, in-place `botPresets` edits never marked the preset block, and a failed save dropped its pending tracker flags.

Sorry for how demanding this thread got.

Approving — thanks for working through every round with real fixes and regression tests!

kangjoseph90:
Thanks for the approval. At first, I thought this would be a relatively simple problem, but your reviews and repeated testing helped me realize that the underlying structure was much more subtle than I expected.

Because you kept thinking through the edge cases with me until the end, I was able to arrive at a design that feels correct and robust. This was also a really valuable learning experience for me.

Thanks again for all the time and effort you put into the review.

cubicj:
@sub-hub 
Sorry for the direct ping. Could you review this one when you have time? It replaces the state snapshot path in saveDb() with a mutation-tracking proxy layer, which removes the per-save deep clone and makes saves noticeably cheaper on large databases. I think this is well worth landing, and I approved it after several rounds of review and testing with the author.

sub-hub (approved):
@cubicj Thanks for the ping. I went through this PR in detail since you asked for my review. Here's where I landed.

My initial worry was memory, specifically, the proxy layer adding persistent overhead on top of Svelte's own state proxy, which could be risky for users with 300MB+ databases who are already barely avoiding OOM on web.
However, after walking through the code, the overhead per node is very small. Since large databases in our case are typically message-heavy with relatively few nested objects, the memory impact scales with the object count rather than the raw data size. It's well within manageable limits.

My bigger concern was correctness, whether future contributors who aren't familiar with this proxy layer might accidentally write code that silently breaks save detection.
To address this, I checked four common bypass vectors: `Object.defineProperty` (no usage on DB objects in production), `Object.freeze` (no usage on DB objects), raw reference retention after assignment (no instances found), and the `{...spread}` pattern with nested proxies you flagged (which is handled nicely by the get trap unwrap then set trap rewrap chain). The current codebase looks clear of these risks. I also worried about the `const db = getDatabase()` pattern, but it's safe because `getDatabase` returns the proxy itself, meaning mutations still trigger the traps.

Overall, I'm satisfied. The performance improvement is clear, and the architecture fits well with our current codebase.

Thanks to both of you for the multiple rounds of review and the detailed discussion, and reading through the thread was genuinely informative.
@kangjoseph90 Great to see you contributing! Please feel free to help us out with reviews more often, since cubicj and I cannot approve our own PRs, having another pair of eyes is incredibly helpful.

PR #1259, opened 2026-02-12
feat: Record inlay creation timestamp

enzi221:
## PR Checklist

- Required Checks

## Summary

image

Records inlay asset creation/attachment timestamps for future utilization. Simply displays them in the explorer for now.

## Related Issues

None.

## Changes

Both `postInlayAsset()` and `writeInlayImage()` records `Date.now()` into the storage.

Also fixes inlay tests as #1250 invalidated the mocking.

## Impact

None; `created` field is optional.

## Additional Notes

Proper utilizations (and display, with better localization) in the explorer will be developed further in subsequent PRs.


cubicj (approved):
Looks good. I don't think the extra `created` field meaningfully changes the storage cost here, since inlay assets already live in their own localforage store and the field stays optional for existing assets.

PR #1267, opened 2026-02-15
chore: Remove unused identifiers in globalApi.svelte.ts

enzi221:
## PR Checklist

- Required Checks

## Summary

Removes all unused imports. Also removes `appWindow`, `formDataToString()`, `trimNonLatin()`.

## Related Issues

None.

## Changes

Please refer to the Summary. Exception: AI law related codes were kept as-is.

## Impact

None; everything were never referenced.


cubicj (commented):
Thanks for the cleanup. I don't see current references to `formDataToString()` or `trimNonLatin()`, so the direction looks reasonable.

This now conflicts with `main` in `src/ts/globalApi.svelte.ts`, and the conflict is around imports that current `main` actively uses.

Could you rebase this on the current main and reapply the cleanup there?

PR #1270, opened 2026-02-17
feat: add background persistence

ModMapper:
## PR Checklist

- Required Checks
    - [✅] Have you added type definitions?
    - [✅] Have you tested your changes?
    - [✅] Have you checked that it won't break any existing features?
- [ ] If your PR uses models[^1], if true, check the following:

## Summary
Plays silent audio upon click to prevent the app from being paused in background mode.
Activated upon initial interaction due to browser media playback policies.

## Related Issues
None

## Changes
Add "Keep Background" checkbox to Advanced Settings.
When checked, silent audio plays on click after refresh.

## Impact
Existing users should experience no impact.

## Additional Notes
To avoid using the `$effect` function, activation is applied only after refresh.


cubicj (commented):
Thanks for the contribution.

This has been superseded by the current `keepSessionAlive` setting on main. The replacement came in through #1358, specifically `5c1b8cd6` (`Add sessionalive settings`), and `903cd230` later reduced the silent audio volume for that session keep-alive path.

Since this PR now conflicts with main and would duplicate the existing `Via Sound` option, I’ll close this as stale. If there is still a behavior gap that the current setting does not cover, please feel free to open a new PR against the current main.

PR #1273, opened 2026-02-17
fix: sanitize AbortSignal in V3 plugin callback args before postMessage

ruyari-cupcake:
## Problem

When a V3 plugin registers a provider via `addProvider`, the callback is never invoked. Instead, the user sees:

`
Plugin Error from MyProvider: {}
`

## Root Cause

In `factory.ts`, the `deserializeArgs` callback proxy passes `innerArgs` directly to `postMessage`. When `requestPlugin` (in `request.ts`) calls the provider function as `v2Function(args, abortSignal)`, the `AbortSignal` object is included in `innerArgs`.

**`AbortSignal` cannot be structured-cloned** (per the HTML spec), so `postMessage` throws a `DataCloneError`. Since `DOMException` has non-enumerable properties, `JSON.stringify(error)` produces `{}`, resulting in the unhelpful error message.

## Fix

Filter `AbortSignal` to `null` in the callback args before `postMessage`:

`	ypescript
const sanitizedArgs = innerArgs.map(arg =>
    arg instanceof AbortSignal ? null : arg
);
`

This is the minimal change needed. The plugin still receives the full request arguments (`prompt_chat`, `mode`, `max_tokens`, etc.) — only the non-cloneable `AbortSignal` is replaced with `null`.

## Impact

- **All V3 plugins using `addProvider`** are currently broken by this bug
- This fix enables V3 plugin providers to work correctly
- No security implications: `AbortSignal` is a cancellation signal, not sensitive data
- Plugins lose abort/cancel functionality, but gain the ability to function at all

### DIFF
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -313,7 +313,7 @@ export class SandboxHost {
             return { __type: 'REMOTE_REF', id } as RemoteRef;
         }
 
-        if(val instanceof Response) {
+        if (val instanceof Response) {
             return {
                 __type: 'CALLBACK_STREAMS',
                 __specialType: 'Response',
@@ -326,7 +326,7 @@ export class SandboxHost {
             };
         }
 
-        if(
+        if (
             val instanceof ReadableStream
             || val instanceof WritableStream
             || val instanceof TransformStream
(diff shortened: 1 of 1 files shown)


cubicj (commented):
This has already been resolved on main. AbortSignal is now serialized into a ref and forwarded via a separate message, so plugins retain cancel functionality.

cubicj:
Following up on my earlier comment. The root cause here was fixed on main through #1276 in `efebc622a`, which serializes `AbortSignal` into a ref and reconstructs it with an `AbortController` on the guest side of the iframe boundary, so plugins keep cancel functionality instead of receiving `null`. #1311 later restored that handling in `SandboxHost` via `fb2588bc5`.

This PR now conflicts with main, and replacing the signal with `null` would be a regression against the current ref-based path, so I am closing it as superseded. If V3 provider callbacks still misbehave on the current main, please open a new issue or PR against it.

PR #1290, opened 2026-02-25
feat: Toggle Preset

sub-hub:
## PR Checklist

- Required Checks

## Summary

image

https://github.com/user-attachments/assets/8660eabd-5987-4bc1-b7d0-5e599b7c6992

This PR adds a advanced setting that saves prompts and module toggles as presets, allowing you to switch them all at once.
Users can use this to always enable only the minimum toggles necessary for their chat environment.

## Changes

Only the logic in the src/lib/SideBars/Toggles.svelte file is modified.
Additionally, only components related to the preset are added without significantly altering the existing frontend.

## Impact

Nothing changes in the environment of users who have not enabled this feature.


cubicj (commented):
Since this PR was opened, the Loadout feature landed on main and covers most of what Toggle Presets does. There are a few gaps, but these feel like Loadout improvements rather than a separate system.

What do you think about building on Loadout instead?

cubicj (commented):
I took another pass at this since it has been quiet for a while.

On the overlap question, I compared this against the current Loadout feature on main. Loadout already snapshots `globalChatVariables`, which is where every `toggle_` value lives, so the core save and restore behavior is covered there. The remaining gaps are that Loadout does not capture `jailbreakToggle`, restores variables as a whole object rather than per toggle, and has no per preset scoping.

I also verified a few issues in the current diff. `resetToggles` deletes every `toggle_` key in `globalChatVariables` while saving only captures the active preset's toggles, so resetting under one bot preset wipes toggle state that belongs to the others. The store is keyed by the bot preset display name, so renaming a preset orphans its saved toggle presets, and two presets sharing a name share one namespace. `loadPreset` only writes keys that exist in the saved object and saving skips toggles that were never touched, so loading does not reproduce the saved state when a toggle was flipped after saving.

Given the overlap, I still think this fits better as a set of Loadout improvements than as a separate system, but I would like to hear your take before going further. Would you prefer reworking this on top of Loadout, or continuing this PR as is with the issues above addressed after a rebase?

sub-hub:
@cubicj Actually, I've been using Loadout recently and have been quite satisfied with it too, haha. Thanks for the thought.

Currently, there is a problem with Loadout where you cannot save only specific sections (for example, if you want to save only the prompt toggle, the module is saved as well), but I haven't found it to be a major issue, so I think it will be fine.

If we ever need it, we can reconsider it later.

PR #1295, opened 2026-03-01
feat: Add Input field next to SliderInput

sub-hub:
## PR Checklist

- Required Checks

## Summary

https://github.com/user-attachments/assets/ad710954-e501-45ac-8f0e-e13351259bbc

This PR changes src/lib/SettingRender.svelde to create an input next to the slider when you render the slider.

## Related Issues

There are many people who have difficulty adjusting the number of sliders in detail at the moment.

## Impact

Now that src/lib/Setting/SettingRender.svelde has been modified, the input will be added next to all sliders shown with this renderer in the future.


cubicj (commented):
The number input direction looks useful, but it exposes an existing odd edge case around `thinkingTokens`.

That setting is currently `min: -1` with `step: 200`, while the actual slider rounding uses a zero-based step grid. With the new number input, the browser input grid and the saved value grid can disagree. For example, pressing the down spinner at `600` can leave `599` displayed even though the saved value is still `600`, and `-1` can be rewritten to `0`.

Could you take another look at this area? The fix may be in the new input normalization, or in cleaning up the `thinkingTokens` minimum value itself.

PR #1301, opened 2026-03-02
fix: remove leaked message event listener in SandboxHost.terminate()

OnlookerRs:
## PR Checklist

- Required Checks
    - [ ] Have you checked if it works normally in all web, local, and node-hosted versions? If it doesn't, have you blocked it in 

## Summary

`SandboxHost.terminate()` does not call `window.removeEventListener()` for the message handler registered by `run()`. After each plugin reload cycle (e.g. installing a new plugin triggers `loadPlugins()` → `loadV3Plugins()`), zombie message listeners accumulate on `window`, and their callback proxies hold references to destroyed iframes, causing V3 plugin handlers (such as `addRisuScriptHandler` callbacks) to silently fail.

## Related Issues

Continuation of #1301 — this PR addresses the remaining resource leaks that were not covered by the message listener fix in `SandboxHost.terminate()`.

## Changes

- Added `messageHandlerRef` instance property to `SandboxHost` to store the message handler reference created in `run()`
- `terminate()` now calls `window.removeEventListener('message', this.messageHandlerRef)` before removing the iframe
- Unified `run()`'s returned cleanup function to delegate to `terminate()`, removing duplicate code

## Impact

- Fixes V3 plugin handlers breaking after plugin reload cycles
- Prevents zombie event listener accumulation on `window`
- No breaking changes — `run()` return value and `terminate()` public API remain the same

## Additional Notes

The `run()` method already had a correct cleanup function as its return value (with `removeEventListener`), but `unloadV3Plugin()` calls `terminate()` instead, which was missing the listener removal. This was likely an oversight when `terminate()` was added separately from the `run()` cleanup.


cubicj (commented):
Reviewed the changes. The `run()` cleanup and `terminate()` had duplicate logic, but only the closure path had `removeEventListener`. I checked whether replacing `run()`'s inline cleanup with a single `this.terminate()` call could drop any behavior. No callers use the returned cleanup, so safe to unify. Build and type check pass.

Branch has conflicts against main (`abortControllers` was added in the same area). Could you rebase?

OnlookerRs:
> Reviewed the changes. The `run()` cleanup and `terminate()` had duplicate logic, but only the closure path had `removeEventListener`. I checked whether replacing `run()`'s inline cleanup with a single `this.terminate()` call could drop any behavior. No callers use the returned cleanup, so safe to unify. Build and type check pass.
> 
> Branch has conflicts against main (`abortControllers` was added in the same area). Could you rebase?

It has been rebased!

cubicj (approved):
#1379 landed right after and re-introduced a trivial conflict in `factory.ts`, but it's mechanical.

sub-hub (approved):
It is a reasonable correction and concise.
Thanks for the PR.

PR #1304, opened 2026-03-03
feat: add request / response body logging to Claude batch API

tasoo-oos:
## PR Checklist

- Required Checks

## Note

This PR is a continuation of the pre-existing PR called https://github.com/kwaroran/Risuai/pull/1263.
I seperated this PR from that because this PR introduces seperate feature based on the new batch API implementation.

## Summary

Unlike Other streaming 
Add request / response body logging logic to the claude batch API.

## Related Issues

None

### Before changes
<details>
<summary>Example images</summary>

image

image

</details>

- For per-message debug logging
  - Batch API does not leave any log
- Request Logs viewer UI in advanced settings
  - Are filled with polling requests, which is useless for end users
  - No response body logged at all, only "Streamed Fetch"

### After changes
<details>
<summary>Example images</summary>

image

image

image

image

</details>

- Batch API does leave a request / response log
- Now there's two log (initial batch API call, final batch result / errored API call) per one batch request
  - No redundant polling request logs saved
- Now request uses header format defined in `anthropic.ts`. Refer to Example images for more detail.

#### No breaking changes
- `fetchNative` now has optional function parameter
- No more polling request visible on the native UI
  - visible only with f12 developers panel
  - Also, it was not so helpful for devs anyway


cubicj (commented):
The batch logging change looks good overall.

One unrelated change: `redacted_thinking` now exposes `content.data` only in the batch path, while the other Claude paths still render `{{redacted_thinking}}`. It also opens a backtick without closing it.

Could you check if that was intentional?

tasoo-oos:
Done! I rolled it back to `{{redacted_thinking}}`.

Changed redacted thinking text was intentional back then, except for the typo. But come to think of it, there would be almost no practical reason to add informational message inside `<thinking>` block.

That was clumsy change on my side, thanks for the feedback!

cubicj (approved):
Verified the updated batch path. `redacted_thinking` now matches the regular Claude path again. Check, tests, and build pass on a trial merge with current `origin/main`.

sub-hub (approved):

PR #1307, opened 2026-03-03
fix: clean up leaked resources on V3 plugin unload

OnlookerRs:
## PR Checklist

- Required Checks

## Summary

When a V3 plugin is unloaded (via `loadV3Plugins()` → `unloadV3Plugin()`), three types of host-side resources are never cleaned up:

1. Document event listeners registered through `SafeElement.addEventListener()` remain on `document` after the plugin's iframe is destroyed. These zombie listeners fire on every user interaction but their RPC callbacks point to a removed iframe.
2. `SafeMutationObserver` has no `disconnect()` method, so plugins cannot stop observing even if they want to, and the framework cannot stop them on unload either.
3. `pluginChannel` entries persist in the global Map after unload, leaving stale callbacks from destroyed plugin instances.

These issues do not occur in V2.1 because `loadV2Plugin()` explicitly clears all handler Sets (`editdisplay`, `editoutput`, etc.) before reloading, and V2.1 plugins run synchronously in the main scope without iframe isolation.

## Related Issues

None

## Changes

Document event listeners (`SafeElement`)
- Added a module-level `documentEventListeners` array to track all listeners registered via `SafeElement.addEventListener()`
- `addEventListener()` pushes to this array when registering on `document`
- `removeEventListener()` splices from this array when the plugin explicitly removes a listener
- `loadV3Plugins()` bulk-clears any remaining entries after all `unloadV3Plugin()` calls complete, following the same pattern as V2.1's `pluginV2.editdisplay.clear()` etc.

SafeMutationObserver
- Added `disconnect()` method wrapping native `MutationObserver.disconnect()`
- `createMutationObserver()` now auto-registers an `addPluginUnloadCallback` that calls `disconnect()`, following the same pattern as `registerButton()` / `registerSetting()`
- Updated `risuai.d.ts` type definition accordingly

pluginChannel
- `addPluginChannelListener()` now auto-registers an `addPluginUnloadCallback` that deletes the channel entry, following the same pattern as `registerButton()` / `registerSetting()`

## Impact

- Prevents zombie document event listeners from accumulating after plugin reload cycles
- Prevents MutationObservers from running indefinitely after their plugin is destroyed
- Prevents stale pluginChannel callbacks from being invoked after unload
- No breaking changes — all existing plugin APIs remain the same, `disconnect()` is a new addition only
- All cleanup operations are idempotent (safe to call multiple times, no-op if already cleaned)

## Additional Notes

The `SafeElement` class does not know which plugin it belongs to, so per-plugin tracking of document event listeners is not feasible without architectural changes to `SafeElement`. Instead, a bulk-clear approach is used in `loadV3Plugins()` as a safety net, consistent with V2.1's reload strategy. Plugins that properly clean up via `onUnload` will have already removed their listeners before this point — the bulk-clear only catches leaked ones.

### DIFF
--- a/src/ts/plugins/apiV3/risuai.d.ts
+++ b/src/ts/plugins/apiV3/risuai.d.ts
@@ -861,6 +861,11 @@ interface SafeMutationObserver {
      * @returns Promise that resolves when observer is set up
      */
     observe(element: SafeElement, options: MutationObserverInit): Promise<void>;
+
+    /**
+     * Stops observing all target elements for changes
+     */
+    disconnect(): Promise<void>;
 }
 
 // ============================================================================
--- a/src/ts/plugins/apiV3/v3.svelte.ts
+++ b/src/ts/plugins/apiV3/v3.svelte.ts
@@ -37,6 +37,7 @@ import { LLMFlags, LLMFormat, LLMProvider, LLMTokenizer, type LLMModel } from "s
 */
 
 const pluginChannel = new Map<string, Function>();
+const documentEventListeners: Array<{type: string, listener: EventListenerOrEventListenerObject, options: any}> = [];
 
 class SafeElement {
     #element: HTMLElement;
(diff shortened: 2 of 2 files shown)


cubicj (approved):
The bulk-clear approach for document listeners looked like it could leak if a single plugin unloaded on its own, but `unloadV3Plugin` is only reachable through `loadV3Plugins`, so the safety net always runs. The observer and channel fixes just plug into the same `addPluginUnloadCallback` path the existing register APIs use.

Build and tests pass locally.

sub-hub (approved):

PR #1308, opened 2026-03-03
fix: handle stream callback response via MessagePort in SandboxHost

rsyumi:
## PR Checklist

- Required Checks

## Summary

Added MessagePort-based stream bridging to fix transmission failures that occurred when returning a ReadableStream in the Plugin API v3 `addProvider` callback.

## Related Issues

None.

## Changes

- Added the `replaceStreamsWithPorts` function to `GUEST_BRIDGE_SCRIPT` to convert ReadableStream into MessagePort within the sandboxed iframe.
- During transmission, the ReadableStream is replaced with a placeholder (`__type: "STREAM_PORT"`) and sent alongside a `portIndex` to reference the corresponding MessagePort in the separately transferred `ports` array.
- The parent frame uses the `reconstructStreamsFromPorts` function to rebuild the ReadableStream from the placeholder.
- To ensure reliability, original stream errors are propagated. In exceptional cases, the process will safely abort after bubbling up the error.
- Applied the same MessagePort-based stream bridging in both directions (guest → host and host → guest) to fix ReadableStream transfer failures on iOS Safari. The bridging logic (`replaceStreamsWithPorts` / `reconstructStreamsFromPorts`) is shared across both the bridge script and the host-side script, keeping the structure fully symmetrical.
- Removed the `CALLBACK_STREAMS` handling from the host-side serialization, as ReadableStream is now uniformly processed via the `STREAM_PORT` placeholder through MessagePort transfer.

## Impact

- Returning a ReadableStream for `content` in the Plugin API v3 `addProvider` callback now works as documented without errors, enabling real-time chat processing.
- Fixes ReadableStream transfer failures on iOS Safari by applying MessagePort-based bridging in both directions (host ↔ sandboxed iframe).
- Note that deeply nested ReadableStream structures are not currently supported. If future plugin APIs require this, the current implementation will need updates. (I am happy to implement this if requested, but omitted it for now since there are no active use cases).

## Additional Notes

- Labeled as a `fix` because existing types and documentation claimed ReadableStream support, despite the runtime errors.
- Please let me know if any changes are needed, including coding style adjustments.
- The initial code was AI-generated, but I have manually reviewed all the code and implemented the necessary adjustments.


OnlookerRs:
Missing `break` in the iframe-side (vanilla JS) `replaceStreamsWithPorts` after the stream ends.

Host side (TypeScript) is correct:
```js
if (done) { ch.port1.postMessage({ done: true }); break; }
```

Iframe side (vanilla JS):
```js
if (done) ch.port1.postMessage({ done: true });
// no break → infinite loop after stream ends
```

The receiver closes the port on the first `done: true` and finishes normally, so it works functionally. However, the sender's `while(true)` loop never exits:
- `reader.read()` keeps returning `{ done: true }` immediately
- `port1.postMessage()` silently drops messages since the other port is closed
- Tight loop pins the CPU, and the `finally` block (`port1.close()`) is never reached → resource leak

Fix:
```js
if (done) { ch.port1.postMessage({ done: true }); break; }
```

rsyumi:
> Missing `break` in the iframe-side (vanilla JS) `replaceStreamsWithPorts` after the stream ends.
> 
> Host side (TypeScript) is correct:
> 
> ```js
> if (done) { ch.port1.postMessage({ done: true }); break; }
> ```
> 
> Iframe side (vanilla JS):
> 
> ```js
> if (done) ch.port1.postMessage({ done: true });
> // no break → infinite loop after stream ends
> ```
> 
> The receiver closes the port on the first `done: true` and finishes normally, so it works functionally. However, the sender's `while(true)` loop never exits:
> 
> * `reader.read()` keeps returning `{ done: true }` immediately
> * `port1.postMessage()` silently drops messages since the other port is closed
> * Tight loop pins the CPU, and the `finally` block (`port1.close()`) is never reached → resource leak
> 
> Fix:
> 
> ```js
> if (done) { ch.port1.postMessage({ done: true }); break; }
> ```

Oh my, how did I miss that? A `while(true)` loop without a `break`...? I didn't expect the lack of syntax highlighting to trip me up like this. Sorry for not catching it during review, and thank you so much for spotting it.

rsyumi:
I've updated the PR by merging the latest version and ran the tests again. I also fixed a small bug I came across along the way to make the code a bit more robust.

It's been running stably in the environments I'm able to test on. If someone run into any issues or would like further clarification on anything, please let me know.

rsyumi:
I've pushed a few follow-up commits to make the stream bridging more robust: pull-based backpressure (bounds memory on large bodies), cleanup of active streams on teardown and on send failure (MessagePort has no `close` event in stable browsers, so these would otherwise hang or leak), and a small consistency fix so unsupported stream placements fail the same way on all browsers.

Side note: ReadableStream transfer is unsupported in all WebKit-based webviews, so this also fixes the Tauri macOS/Linux builds, not just iOS Safari.

It's been running stably in the environments I'm able to test on. If anyone runs into issues, please let me know.

cubicj (approved):
Looks good. I checked the current head against both stream directions: provider callback streams now bridge back to the host through `MessagePort`, and `nativeFetch()` `Response.body` is reconstructed before the plugin receives the `Response`.

The pull-based bridge keeps backpressure and cancel/teardown cleanup in place. `pnpm check`, `pnpm test`, and `pnpm build` pass locally, and CodeQL is green.

sub-hub (approved):

PR #1309, opened 2026-03-03
Feat/get database field filtering

OnlookerRs:
## PR Checklist

- Required Checks

## Summary

`getDatabase()` uses `$state.snapshot()` to deep-copy requested keys, but top-level keys like `characters`, `modules`, and `pluginV2` contain heavy nested data (full chat histories, lorebooks, CJS code) that gets copied entirely even when plugins only need lightweight metadata. On mobile devices with 200+ characters × 500+ chat turns, this causes significant performance and memory pressure.

This PR adds optional nested field filtering to `getDatabase()` and explicit write-back control to `setDatabaseLite()`/`setDatabase()`, allowing plugins to avoid copying unnecessary nested data.

## Related Issues

None

## Changes

getDatabase — nested field filtering
- Added optional second parameter `options: { include?, exclude? }` where keys are top-level DB key names and values are arrays of nested field names
- `include`: only snapshot the specified nested fields (iterates include list directly, skips `Object.keys`)
- `exclude`: snapshot everything except the specified nested fields
- Handles arrays (e.g. `characters[]`), plain objects (e.g. `pluginCustomStorage`), and primitives (no filtering applied) correctly
- Null/non-object array items are passed through safely via `$state.snapshot()` without field iteration
- When `options` is not provided, follows the exact same code path as before — zero overhead

setDatabaseLite / setDatabase — write-back control
- Added optional second parameter `options: { include?, exclude? }` where values are arrays of top-level key names
- `include`: only write back the specified keys
- `exclude`: skip the specified keys during write-back
- Allows plugins that fetched partial data via `getDatabase` filtering to safely write back without overwriting fields they didn't fetch

Type definitions (risuai.d.ts)
- Added `GetDatabaseOptions` and `SetDatabaseOptions` interfaces
- Updated `getDatabase`, `setDatabaseLite`, `setDatabase` signatures and JSDoc with usage examples

## Impact

- No breaking changes — all new parameters are optional, existing calls work identically
- Plugins can now fetch `characters` without copying full chat histories, or `modules` without copying lorebook/CJS data
- All data passes through `postMessage` structured clone (iframe-safe) — no non-enumerable tricks or hidden markers
- Write-back safety is explicit rather than automatic, giving plugin authors full control

## Additional Notes

Usage examples:

```ts
// Fetch character metadata without chat histories
const db = await risuai.getDatabase(['characters'], {
  exclude: { characters: ['chats', 'globalLore'] }
});

// Fetch only names and types
const db = await risuai.getDatabase(['characters'], {
  include: { characters: ['name', 'type', 'nickname'] }
});

// Safe write-back of only what you modified
await risuai.setDatabaseLite(db, { include: ['personas', 'selectedPersona'] });
```


cubicj (commented):
The filtering direction looks useful, but I found one write-back safety issue.

A filtered `getDatabase` result can be passed back to `setDatabaseLite` or `setDatabase`, and the current code replaces the whole top-level DB key. That can overwrite the user's DB with partial data, for example dropping `characters[].chats`.

Since this is a DB write API, I think the unsafe partial-read write-back path should be blocked in code rather than documented as something plugin authors must avoid. Could you check this?

OnlookerRs:
> The filtering direction looks useful, but I found one write-back safety issue.
> 
> A filtered `getDatabase` result can be passed back to `setDatabaseLite` or `setDatabase`, and the current code replaces the whole top-level DB key. That can overwrite the user's DB with partial data, for example dropping `characters[].chats`.
> 
> Since this is a DB write API, I think the unsafe partial-read write-back path should be blocked in code rather than documented as something plugin authors must avoid. Could you check this?

Good catch. The issue was setDatabase replacing the whole key, so writing a filtered read back dropped whatever you left out (chats etc).

Now a filtered key gets merged instead of replaced: array items match by id (chaId for characters, id otherwise) and anything you didn't read is kept from the current db, so nothing gets wiped. getDatabase always keeps the id even if you filter it out so the match can't break, and full (unfiltered) writes still replace like before.

One caveat: you need to pass the object getDatabase returned straight back in. If you build a fresh object yourself the tag is lost and it falls back to replace.

cubicj (commented):
Thanks for the follow-up. The partial merge looks safer for `characters` and other `id`-based arrays, but `plugins` still seems risky.

Installed plugins are keyed by `name`, while `mergePartialDbValue()` matches non-`characters` arrays by `id`. A filtered read like `getDatabase(['plugins'], { include: { plugins: ['name', 'enabled'] } })` can still write back a partial plugin entry and drop fields such as `script`, `version`, `arguments`, or `realArg`.

Could you check the partial write-back behavior for `plugins` too?

Issue #584, opened 2024-08-02
import character & load backup locally not working on Safari / Tauri on Mac OS


dohyun-ko:
File selector does not appear when "import character" or "load backup locally" button clicked.

Safari security policy seems to be the cause of this issue - input.click() is not working when display='none' or that isn't the direct result of user interactions.
Refer to Stack Overflow comment: https://stackoverflow.com/questions/26501221/calling-click-event-of-input-doesnt-work-in-safari#:~:text=Click%20event%20in%20safari%20on%20input%20type%20file%20will%20work%20if%20it%20is%20not%20hidden%20and%20click()%20is%20called%20out%20of%20user%20interaction%20(e.g%20inside%20onclick%20handler%20etc)

I've tried to fix it, but it requires too many huge structural changes.

Asherathe:
I've noticed the same issue with the Linux AppImage with the file selector not showing up to load a backup locally. Importing characters does work, though. No Safari in this case, of course, so I'm not sure if the cause is related or not.

dilluti0n:
I guess it is related cause both use webkit to render pages.

Safari uses webkit natively, and linux build uses libwebkit2gtk.

I experience the same issues from both macOS dmg build and Linux rpm build.

PR #1324, opened 2026-03-15
fix: unify file picker flows and harden DOM fallback for WebKit/Tauri

tasoo-oos:
## PR Checklist

- Required Checks

fix: unify file picker flows and harden DOM fallback for WebKit/Tauri

## Summary

Migrate all remaining direct DOM file input usages to the shared `selectSingleFile`/`selectMultipleFile` utilities, enable the Tauri native file dialog path, and fix several correctness issues in the DOM fallback.

## Related Issues

Addresses the file selector not appearing on Safari/WebKit and Linux AppImage when clicking "import character" or "load backup locally".
[External bug report](https://arca.live/b/commit/164505075?p=1)
https://github.com/kwaroran/Risuai/issues/584

## Changes

- `selectSingleFile`/`selectMultipleFile` now route to Tauri native dialog when running in Tauri, instead of always using the DOM picker
- Migrated importCharacter (`characterCards.ts`), LoadLocalBackup (`backuplocal.ts`), trigger import (`TriggerV2List.svelte`), and whisper file select (`PlaygroundSubtitle.svelte`) to the shared utilities
- Moved backup file selection before confirmation dialogs so the picker opens directly from the user click, avoiding WebKit gesture restrictions
- Refactored `LoadLocalBackup` into `loadBackupFromStream`/`loadBackupFromBytes` to separate file selection from restore logic
- Fixed `selectSingleFile` DOM path crash on cancel (v[0] on empty array)
- Unified cancel return types: `selectSingleFile` returns `null`, `selectMultipleFile` always returns an array (never null)
- Replaced `display:none` with offscreen positioning for the hidden input
- Fixed extension filtering to use `lastIndexOf('.')` instead of `split('.').pop()` to handle edge cases correctly
- Increased focus-based cancel detection delay from 0ms to 300ms
- Added Tauri FS permissions for $HOME, $DESKTOP, $DOCUMENT directories
- Fixed null-unsafe callers in `scripts.ts`, `lorebook.svelte.ts`, and OtherBotSettings.svelte that would crash on picker cancellation

## Impact

File selection should now work on Tauri desktop builds (macOS, Linux) where it previously failed. Web/Safari users should see fewer cases of the file dialog not appearing. Cancelling a file picker no longer causes runtime errors in any flow.

## Notes

Working on backup file load selection.
Testing on arch linux.

### DIFF
--- a/src-tauri/Info.plist
+++ b/src-tauri/Info.plist
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>CFBundleAllowMixedLocalizations</key>
+    <true/>
+    <key>CFBundleLocalizations</key>
+    <array>
+        <string>en</string>
+        <string>ko</string>
+        <string>zh-Hans</string>
+        <string>zh-Hant</string>
+        <string>vi</string>
+        <string>de</string>
+        <string>es</string>
+    </array>
+</dict>
+</plist>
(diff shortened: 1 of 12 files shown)


tasoo-oos:
This is beyond my comprehension and I don't use desktop version anyway.

PR #1326, opened 2026-03-16
fix: prevent localStorage quota overflow in plugin safety cache

OnlookerRs:
## PR Checklist

- Required Checks

## Summary

`checkCodeSafety` in `pluginSafety.ts` stores the full `modifiedCode` in localStorage for each plugin, keyed by code hash (`safety-${hash}`). When plugins are updated, old entries are never removed and new ones accumulate indefinitely. Once localStorage quota is exceeded, the unhandled `QuotaExceededError` from `localStorage.setItem` crashes `checkCodeSafety`, which in turn crashes `loadV2Plugin`, preventing all plugins (including V3) from loading.

This affects all platforms, environments with smaller quotas (e.g. Safari iOS at 5 MB) hit it sooner, but any environment will eventually reach the limit given enough plugin updates.

## Related Issues

None

## Changes

- Wrapped cache read (`localStorage.getItem` + `JSON.parse`) in try-catch; removes stale entry on checkerVersion mismatch or parse failure
- Wrapped cache write (`localStorage.setItem`) in try-catch; on quota error, purges all `safety-*` entries and retries once
- Added `purgeSafetyCache` helper that removes all `safety-*` localStorage entries except an optional keep key

## Impact

- Fixes all-plugins-dead scenario caused by localStorage quota overflow
- No behavioral change in normal operation, try-catch only activates on error
- Worst case after purge: one-time AST re-parse per plugin on next load (cache rebuilds itself)
- Only `safety-*` keys are touched; no other localStorage data is affected

## Additional Notes

The `safety-*` cache entries can be large (hundreds of KB each for big plugins) because they store the entire transformed source code. A single large plugin updated a few times can consume over 1 MB of orphaned cache entries. This fix is reactive (purge on quota error) rather than proactive, keeping the change minimal and preserving the existing caching strategy.

### DIFF
--- a/src/ts/plugins/pluginSafety.ts
+++ b/src/ts/plugins/pluginSafety.ts
@@ -57,12 +57,17 @@ export async function checkCodeSafety(code: string): Promise<CheckResult> {
 
     const hashedCode = await hasher(new TextEncoder().encode(code));
     const cacheKey = `safety-${hashedCode}`;
-    const cachedResult = localStorage.getItem(cacheKey);
-    if (cachedResult) {
-        const got =  JSON.parse(cachedResult) as CheckResult;
-        if (got.checkerVersion === checkerVersion) {
-            return got;
+    try {
+        const cachedResult = localStorage.getItem(cacheKey);
+        if (cachedResult) {
+            const got = JSON.parse(cachedResult) as CheckResult;
+            if (got.checkerVersion === checkerVersion) {
+                return got;
+            }
+            localStorage.removeItem(cacheKey);
         }
+    } catch {
+        try { localStorage.removeItem(cacheKey); } catch {}
     }
 
     try {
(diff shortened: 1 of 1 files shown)


cubicj (approved):
Verified. The try-catch + purge-and-retry correctly downgrades the crash into a one-time cache rebuild, and the read path also cleans stale entries. `safety-` prefix is only used here, so purge is safe.

One thing left open: orphan entries from old plugin versions still accumulate until the quota is hit. This fix reacts when that happens, but doesn't prevent the buildup itself. Worth considering as a follow-up, but out of scope here.

sub-hub (approved):
Since this PR only touches the cache within `checkCodeSafety`, and `purgeSafetyCache` is executed at the point where an error occurs in the existing code,
It seems like it will be better than the current situation even if a bug occurs during the process of clearing the safety cache.

Thanks for the PR.

PR #1345, opened 2026-03-31
fix: Change font color to be seen even in the light theme

sub-hub:
## PR Checklist

- Required Checks

### Before
image

### After
image

You can choose several color palettes in Risu, but there was the part where the letter color was fixed in white without using the color of this changing palette.
So I changed the fixed white to textcolor.

And I also modified AGENTS.md to prevent this from happening next time.

# Impact
image

When you modify the code like this, tailwind uses color-mix to create colors dynamically.
I think the color-mix function was created recently, but it seems to be supported by all browsers now, so it seems ok.
https://caniuse.com/?search=color-mix

It's a simple code change, but please make a judgment on this color-mix.
And please see if it's okay with the AGENTS.md prompt.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -142,6 +142,18 @@ Key stores:
 - `loadedStore`, `alertStore` - Application state
 - `DynamicGUI` - Responsive layout switching
 
+### Styling & Theming
+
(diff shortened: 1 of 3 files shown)


cubicj (approved):
Verified the fix. `textcolor` is a theme CSS variable, so this properly respects all palettes. The AlertComp change also removes a latent `text-white` override bug.

PR #1361, opened 2026-04-04
fix: [node] prevent /api/login crash from undefined writeFileSync

rndka9-alt:
## PR Checklist

- Required Checks

## Summary

Replace the function with one that actually exists.

https://github.com/rndka9-alt/Risuai/blob/32ff3518180119144bbbdc0d9eee2755601a9cf1/server/node/server.cjs#L7-L8
```js
const { existsSync, mkdirSync, readFileSync, writeFileSync } = require('fs');
const fs = require('fs/promises')
```

## Related Issues

None.

## Changes

Replace `fsPromises.writeFileSync` (not defined) with `fs.writeFileSync`

## Impact

Fixes the server crash that occurred when typing the password.

## Additional Notes

`fs` is imported from `fs/promises`, which can be misleading — consider renaming the const.

### DIFF
--- a/server/node/server.cjs
+++ b/server/node/server.cjs
@@ -1384,7 +1384,7 @@ app.get('/api/oauth_callback', async (req, res) => {
         },
     )
 
-    fs.writeFileSync(authCodePath, tokens.access_token, 'utf-8')
+    writeFileSync(authCodePath, tokens.access_token, 'utf-8')
 
     res.send(tokens)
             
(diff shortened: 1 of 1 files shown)


cubicj (approved):
Looks good. Same fs.writeFileSync pattern existed on line 1387 (OAuth callback), so I pushed an additional commit to fix that too.

rndka9-alt:
Thank you for the review, and also for helping keep the project maintained.

It looks like there were a few things I missed because I was not able to check everything closely, so thank you for addressing them!

It seems that related issues have already been reported, 
so I’m leaving them here in a comment: #1365 #1366

PR #1362, opened 2026-04-05
fix: accumulate Ollama streaming chunks for correct display

### LINKED ISSUE #587: Bug: Ollama source stream issue
When using Ollama (url `http://localhost:11434`), the response only shows the last streamed chunk/token/word.

It looks like it's replacing the response message with the last streamed token/chunk instead of appending to the previously streamed chunks/tokens.

This is on latest master (commit `952ee5f`) checked out a few minutes ago.

cubicj:
## PR Checklist

- Required Checks

## Summary

Fixes Ollama streaming responses only showing the last token instead of the full accumulated response.

## Related Issues

Fixes #587

## Changes

**`src/ts/process/request/request.ts`** — Accumulate streamed chunks in the Ollama request handler.

The Ollama streaming path was enqueueing each chunk's `message.content` (a single token) directly, but the stream consumer in `index.svelte.ts` uses assignment (`result = value`) rather than accumulation (`result += value`). This caused only the last received token to be displayed.

All other streaming providers (OpenAI, Anthropic, Ooba) already accumulate text before enqueueing. This fix aligns the Ollama path with that pattern.

## Additional Notes

An external patch for the same issue was previously shared in #587: https://github.com/orsonteodoro/oiledmachine-overlay/blob/master/games-rpg/RisuAI/files/RisuAI-139.2.0-ollama-fix.patch

### DIFF
--- a/src/ts/process/request/request.ts
+++ b/src/ts/process/request/request.ts
@@ -1011,9 +1011,11 @@ async function requestOllama(arg:RequestDataArgumentExtended):Promise<requestDat
 
     const readableStream = new ReadableStream<StreamResponseChunk>({
         async start(controller){
+            let text = ''
             for await(const chunk of response){
+                text += chunk.message.content
                 controller.enqueue({
-                    "0": chunk.message.content
+                    "0": text
                 })
             }
             controller.close()


cubicj:
This has already been resolved on main. The current Ollama streaming path accumulates content before enqueueing, so #587's last-token display issue is covered.

Closing this as superseded.

PR #1369, opened 2026-04-07
perf: optimize chat hot paths

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This PR reduces repeated work in chat rendering.

## Related Issues

None

## Changes

- Memoize parser metadata encoding in `addMetadataToElement()` so repeated renders reuse the same encoded watermark payload instead of rebuilding it every time.
- Fix the Google Cloud tokenizer cache to read with the same key used for cache writes, restoring cache hits for repeated token counts.
- Reduce repeated work in `ChatBody` by preparing normalized asset lookup structures once per image scan instead of rebuilding them for every image.
- Narrow `Chats` DOM lookups to the local chat container instead of document-wide queries during incremental message updates.

## Impact

- Improves chat-screen hot-path efficiency, especially on repeated renders and image-heavy messages.
- Keeps user-facing behavior unchanged, except that the Google Cloud tokenizer cache now works as intended.

## Additional Notes

- Verified with `pnpm test` and `pnpm build`.
- `pnpm check` still reports the existing baseline issues in `CustomModelsSettings.svelte`, `BotSettings.svelte`, and the `SliderInput.svelte` accessibility warning; this PR does not add new ones.


cubicj (commented):
The `showDebugLogs` gating feels like an in-between state to me. If they help debugging, keeping them visible in production might be more useful. If not, removing them would reduce log spam more effectively than conditionally gating. The codebase has ~270 ungated `console.log` calls elsewhere, so this could also introduce an inconsistency.

Verified that the tokenizer cache `get()` was reading with a different key than `has()` and `set()`, so the cache never actually hit. The rest of the optimizations don't change behavior. Build and tests pass locally.

SameDesu123:
> The `showDebugLogs` gating feels like an in-between state to me. If they help debugging, keeping them visible in production might be more useful. If not, removing them would reduce log spam more effectively than conditionally gating. The codebase has ~270 ungated `console.log` calls elsewhere, so this could also introduce an inconsistency.
> 
> Verified that the tokenizer cache `get()` was reading with a different key than `has()` and `set()`, so the cache never actually hit. The rest of the optimizations don't change behavior. Build and tests pass locally.

Thank you for contribution! I thought about the `showDebugLogs` gating again, and I agree with your point. Since RisuAI is mostly used by advanced users, your approach makes a lot of sense.

Therefore, I will revert the log-related changes shortly.

cubicj (approved):
Log gating revert confirmed. No other changes from the previous review. Build and tests still pass.

PR #1374, opened 2026-04-10
feat: always deobfuscate alert stack traces

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This PR improves error visibility in the existing alert flow. Stack traces are now always translated through sourcemaps when possible, the Risu version is shown above the log, and users see a localized fallback message when translation fails.

## Related Issues

None.

## Changes

- Removed the settings toggle for sourcemap stack trace translation and made the behavior always on in the existing error alert.
- Added the current Risu version above the error log in the alert details view.
- Kept the original obfuscated stack trace visible when sourcemap translation fails.
- Localized the translation failure message for all supported Risu languages: en, ko, cn, zh-Hant, vi, de, and es.
- Simplified the sourcemap helper so it only reports translation success or failure.
- Added tests for the sourcemap translation helper.

## Impact

- Users get more useful error details without needing to enable a setting first.
- Localization now covers the fallback error message, so failure handling matches the current app language.
- Existing error alert behavior stays intact, with changes scoped to the details display and sourcemap translation path.

## Additional Notes

- Validation: pnpm test and pnpm build passed.
- pnpm check still reports the existing baseline issues in CustomModelsSettings.svelte, BotSettings.svelte, and SliderInput.svelte that are unrelated to this PR.


SameDesu123:
I think this could be helpful for dev when they get a report from a user.

cubicj (dismissed):

PR #1385, opened 2026-04-11
feat(alert): include environment metadata in error details 

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This PR makes the error details alert more useful when someone reports a crash or shares a stack trace. The code block now includes the user's OS, user agent, and the Risu runtime environment right below the Risu version line. When the runtime can provide it, the OS line also includes a more specific version instead of just the OS name.

## Related Issues

None

## Changes

- Added `OS`, `User-Agent`, and `Risu environment` to the existing error details block in `AlertComp.svelte`, directly under the Risu version line.
- Moved the environment detection logic into shared helpers in `src/ts/platform.ts` so the alert component can use one consistent source of truth.
- Added a browser-side fallback for OS detection, with richer version info from `navigator.userAgentData` when the browser exposes it.
- Added native OS version support for the Tauri build through `@tauri-apps/plugin-os`.
- Updated the runtime label so Vite dev sessions show up as `web(dev)` instead of being confused with node-hosted environments.

## Impact

This should be low risk because it only changes the metadata shown in error alerts. Normal chat and settings flows should not be affected. The main limitation is that browser-hosted sessions may still expose only partial OS version information, depending on the browser's privacy restrictions.

## Additional Notes

- Checked with `pnpm test`, `pnpm build`, and `pnpm check`.

| Preview |
|---------|
| more-metadata-error |

### DIFF
--- a/src/lib/Others/AlertComp.svelte
+++ b/src/lib/Others/AlertComp.svelte
@@ -26,41 +26,22 @@
     import { getChatBranches } from "src/ts/gui/branches";
     import { getCurrentCharacter } from "src/ts/storage/database.svelte";
     import { translateStackTrace } from "../../ts/sourcemap";
-    import { isIOS, isNodeServer, isTauri } from "src/ts/platform";
+    import { getDetailedOSLabel, getFallbackOSLabel, getRisuEnvironmentLabel } from "src/ts/platform";
     import versionData from "../../../version.json";
 
     let showDetails = $state(false);
     let translatedStackTrace = $state('');
     let stackTraceTranslationFailed = $state(false);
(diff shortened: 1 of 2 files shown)


sub-hub (approved):

sub-hub:
It's a feature I made, and thank you for improving it well!

cubicj (approved):
Verified no side effects on existing exports. import * as tauriOs matches the codebase pattern (every Tauri plugin uses top-level import), and isIOS() guard removal was already dead code since the module accesses navigator at the top level.

Build and tests pass locally.

PR #1387, opened 2026-04-12
fix: prevent Loadout crash when referenced preset or persona is missing

cubicj:
## PR Checklist

- Required Checks

## Summary

Fixes multiple crash paths in the Loadout system when a saved Loadout references a preset or persona that has since been deleted. The existing `!== -1` guards did not account for `undefined` returned by optional chaining on `findIndex`, allowing downstream functions (`changeToPreset`, `changeUserPersona`, `saveCurrentPreset`) to receive invalid indices and crash on property access.

## Related Issues

None (reported via user feedback).

## Changes

**`src/ts/loadout.ts`**
- Guard `botPresets` access in `makeLoadout` with optional chaining to prevent crash when `botPresetsId` is out of bounds
- Prevent `undefined` from being pushed into `characterIds` when no character is active
- Replace `!== -1` with `>= 0` for both persona and preset `findIndex` results, blocking `undefined` from passing the guard

**`src/ts/storage/database.svelte.ts`**
- Strengthen `saveCurrentPreset` early return: `=== -1` → `< 0 || !pres?.[db.botPresetsId]`, covering both negative indices and out-of-bounds access after preset deletion

## Impact

- All changes only affect error paths (missing/deleted presets or personas). Normal operation where presets and personas exist is byte-identical to prior behavior.
- `saveCurrentPreset` early return skips saving a nonexistent preset but does not block the subsequent preset switch or download.
- Existing Loadouts with `undefined` entries in `characterIds` are unaffected — the fix only prevents future pollution.

### DIFF
--- a/src/ts/loadout.ts
+++ b/src/ts/loadout.ts
@@ -19,7 +19,7 @@ export function makeLoadout(options:{
 }): Loadout {
     const character = getCurrentCharacter()
     const id = crypto.randomUUID()
-    const preset = DBState.db.botPresets[DBState.db.botPresetsId]
+    const preset = DBState.db.botPresets?.[DBState.db.botPresetsId]
     return safeStructuredClone({
         name: options.name,
         id: id,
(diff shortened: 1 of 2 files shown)


cubicj:
Closing — the crash paths addressed here were already resolved by 0c9571e and e75654e.

PR #1391, opened 2026-04-12
fix: release doingChat lock in sendChat plugin API

cotton100:
## PR Checklist

- Required Checks

## Summary
- Release `doingChat` lock after `processSendChat` completes in the plugin API path
- Reorder the `pluginmodel:::` check so a blocked model doesn't leave an orphan user message

## Problem

When a v3 plugin calls `risuai.sendChat(message)`, the `doingChat` lock is set to `true` (inside `processSendChat`) but never reset when generation finishes. The UI send path handles this in `DefaultChatScreen.svelte`, but the plugin API path does not go through that code.

As a result, subsequent `sendChat` calls throw `"A chat is already in progress"` even after the previous generation completed normally.

## Changes

1. Wrap `processSendChat` in `try/finally` so `doingChat` is released on both success and failure.
2. Move the `pluginmodel:::` check above the user message push so that a blocked model doesn't leave an orphan user message in the chat log.

`return true` still means "generation completed successfully" — same as before, just with the lock now properly released.

## Note

This started as an OOM investigation (original PR title), but on further testing the OOM could not be cleanly reproduced against `main`. The confirmed, reproducible issue is the `doingChat` lock, so the PR scope has been narrowed to that fix plus the small validation reordering.

### DIFF
--- a/src/ts/plugins/apiV3/v3.svelte.ts
+++ b/src/ts/plugins/apiV3/v3.svelte.ts
@@ -1168,6 +1168,11 @@ const makeRisuaiAPIV3 = (iframe:HTMLIFrameElement,plugin:RisuPlugin) => {
                 throw new Error("A chat is already in progress");
             }
 
+            if(getModelInfo(DBState.db.aiModel).id.startsWith('pluginmodel:::')){
+                // Executing plugin provider is block because it can be used for loopholes for ipc right now.
+                throw new Error("Sending chat with plugin-based model is currently blocked");
+            }
+
             const charId = get(selectedCharID);
             const char = DBState.db.characters[charId];
             if(!char){
@@ -1187,12 +1192,13 @@ const makeRisuaiAPIV3 = (iframe:HTMLIFrameElement,plugin:RisuPlugin) => {
                 });
             }
 
-            if(getModelInfo(DBState.db.aiModel).id.startsWith('pluginmodel:::')){
-                // Executing plugin provider is block because it can be used for loopholes for ipc right now.
-                throw new Error("Sending chat with plugin-based model is currently blocked");    
+            try {
+                await processSendChat(-1, {});
+            } finally {
+                // Plugin API path does not pass through the UI unlock logic,
+                // so release doingChat here on both success and failure.
+                doingChat.set(false);
             }
-
-            await processSendChat(-1, {});
 
             return true;
         },


sub-hub:
Thank you for the PR!
OOM bug was really weird, and it's nice that you tried to fix it.

I have a couple of questions and concerns regarding the implementation:

1. doingChat state not being reset on success
In your code, you are only setting doingChat.set(false) inside the .catch() block.
The issue is that processSendChat does not internally reset doingChat to false when a generation finishes successfully.
Shouldn't this be placed inside a .finally() block so the lock is released regardless of success or failure?
I know that the existing code didn't have doingChat in the first place! It's something that seems a little weird to me in logic, and it's not a serious error.

2. Synchronous to Asynchronous transition
You changed the sendChat function from awaiting the response to a fire-and-forget approach. I read your PR description mentioning that the previous await caused OOM, meaning no existing plugin could have successfully relied on the awaiting completion anyway.
However, this is still a significant semantic change to the API's behavior.

I'm not a plugin expert myself, so I believe this sync-to-async change definitely needs more testing.

Could anyone with deeper knowledge of the plugin ecosystem review this specific transition?

cotton100:
> Thank you for the PR! OOM bug was really weird, and it's nice that you tried to fix it.
> 
> I have a couple of questions and concerns regarding the implementation:
> 
>     1. doingChat state not being reset on success
>        In your code, you are only setting doingChat.set(false) inside the .catch() block.
>        The issue is that processSendChat does not internally reset doingChat to false when a generation finishes successfully.
>        Shouldn't this be placed inside a .finally() block so the lock is released regardless of success or failure?
>        I know that the existing code didn't have doingChat in the first place! It's something that seems a little weird to me in logic, and it's not a serious error.
> 
>     2. Synchronous to Asynchronous transition
>        You changed the sendChat function from awaiting the response to a fire-and-forget approach. I read your PR description mentioning that the previous await caused OOM, meaning no existing plugin could have successfully relied on the awaiting completion anyway.
>        However, this is still a significant semantic change to the API's behavior.
> 
> 
> I'm not a plugin expert myself, so I believe this sync-to-async change definitely needs more testing.
> 
> Could anyone with deeper knowledge of the plugin ecosystem review this specific transition?

1. (doingChat .finally())

Thanks for pointing this out! I checked and processSendChat doesn't reset doingChat on success — the UI path handles it separately in DefaultChatScreen.svelte, but the plugin API path doesn't go through there. Moved doingChat.set(false) to .finally() so it resets regardless of success or failure. Also added a comment explaining why. Pushed!

2. (sync→async)

Agreed, it's a big semantic change. Since the previous await always caused OOM, I don't think any plugin was relying on awaiting completion, but more testing would definitely be good. I'll test on my end once merged.

cubicj (commented):
Tested the await path on main with a v3 plugin calling risuai.sendChat(). Full streaming response completed normally, no OOM.

The issue I could reproduce is doingChat not being reset after completion. The UI path handles this in DefaultChatScreen.svelte, but the plugin API path does not. The pluginmodel::: check move is a nice improvement.

Wondering if adding doingChat.set(false) after the await would be a simpler fix that preserves the current API semantics.

cotton100:
> Tested the await path on main with a v3 plugin calling risuai.sendChat(). Full streaming response completed normally, no OOM.
> 
> The issue I could reproduce is doingChat not being reset after completion. The UI path handles this in DefaultChatScreen.svelte, but the plugin API path does not. The pluginmodel::: check move is a nice improvement.
> 
> Wondering if adding doingChat.set(false) after the await would be a simpler fix that preserves the current API semantics.

Went with your suggestion — kept `await` and wrapped it in `try/finally` so `doingChat` is released on both success and failure.

You're right that the OOM isn't cleanly reproducible in a clean environment; I was likely conflating an environment-specific issue with the API semantics. Narrowed the PR scope to just the lock fix plus the pluginmodel check reorder, and updated the title and description to match.

cubicj (approved):
Verified the updated diff. try/finally preserves the await semantics while fixing the doingChat lock on the plugin API path. Tested this approach locally and the chat lock releases correctly after generation completes.

sub-hub (approved):
Now this PR changes the pluginmodel check to happen earlier and ensures that doingChat.set(false) is always called.
This change looks reasonable.

PR #1398, opened 2026-04-16
fix: drag-and-drop reordering rework

tasoo-oos:
## PR Checklist

- Required Checks

## Summary

Fixes sidebar character drag-and-drop reordering/folder creation and Linux/Chromium external file drops.

## Changes

- Adds scoped drag types for app, sidebar, preset, and prompt drag operations.
- Prevents internal app/sidebar drags from reaching the file-import path.
- Lets external file drops over sidebar elements bubble to `<main>` for import.
- Uses `dropEffect = "copy"` for file drops and `move` for internal reordering.
- Clears sidebar drag state after drag end, drop, or window blur.
- Persists folder creation by updating `characterOrder` and running `checkCharOrder()`.

## Impact

Improves drag-and-drop reliability by scoping internal drag types and separating them from external file drops. This should fix sidebar character/folder reordering and Linux/Chromium file-drop behavior without breaking prompt, preset, TriggerV2, lorebook, persona, or chat reordering.

The main risk is browser/OS-specific drag-and-drop behavior, since this PR changes shared app-level drop handling in `App.svelte`.

### Currently tested with:

- Both with Linux/Wayland and Windows:
  - sidebar character reorder
  - sidebar folder creation
  - move character into/out of folder
  - external .risu/image/card drop over main area
  - external file drop over sidebar
  - drag existing in-app /asset
  - prompt template reorder
  - bot preset reorder
  - TriggerV2 trigger reorder
  - TriggerV2 effect reorder
  - lorebook reorder
  - persona reorder
  - chat reorder and chat-folder reorder

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -200,6 +200,7 @@ See `plugins.md` for comprehensive plugin development guide.
 - Multiple UI modes: Classic, WaifuLike, WaifuCut
 - Dynamic GUI switching based on viewport
 - No traditional router; uses conditional rendering in App.svelte
+- In-app drag-and-drop uses custom MIME types to avoid conflicting with file imports; see `src/ts/dragTypes.ts`
 
 ## Supported AI Providers
 
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -44,13 +44,18 @@
     let aprilFools = $state(new Date().getMonth() === 3 && new Date().getDate() === 1)
     let aprilFoolsPage = $state(0)
     let keepingSessionAlive = $state(false)
+
+    const getMainDropEffect = (e:DragEvent): DataTransfer['dropEffect'] => {
+        const types = Array.from(e.dataTransfer?.types ?? [])
+        return types.includes('Files') ? 'copy' : 'link'
+    }
 </script>
 
 <!-- svelte-ignore a11y_click_events_have_key_events -->
 <!-- svelte-ignore a11y_no_noninteractive_element_interactions -->
 <main class="flex bg-bg w-full h-full max-w-100vw text-textcolor" ondragover={(e) => {
     e.preventDefault()
-    e.dataTransfer.dropEffect = 'link'
+    e.dataTransfer.dropEffect = getMainDropEffect(e)
 }} ondrop={async (e) => {
     e.preventDefault()
     if (e.dataTransfer.types.includes('application/x-risu-internal')) {
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -36,7 +36,7 @@
     import IrisModal from './lib/Others/IrisModal.svelte';
     import Legal from './lib/Others/Legal.svelte';
     import CustomSidebarConfig from './lib/Others/CustomSidebarConfig.svelte';
-    import { RISU_SIDEBAR_DRAG_TYPE } from './ts/dragTypes';
+    import { RISU_APP_INTERNAL_DRAG_TYPE, RISU_SIDEBAR_DRAG_TYPE } from './ts/dragTypes';
 
 
   
@@ -46,12 +46,19 @@
     let aprilFoolsPage = $state(0)
     let keepingSessionAlive = $state(false)
 
-    const getMainDropEffect = (e:DragEvent): DataTransfer['dropEffect'] => {
+    const getMainDropEffect = (e:DragEvent): DataTransfer['dropEffect'] | null => {
         const types = Array.from(e.dataTransfer?.types ?? [])
         if(types.includes(RISU_SIDEBAR_DRAG_TYPE)){
             return 'none'
         }
-        return types.includes('Files') ? 'copy' : 'link'
+        if(types.includes(RISU_APP_INTERNAL_DRAG_TYPE)){
+            return null
+        }
+        return types.includes('Files') ? 'copy' : null
+    }
+
+    const markAppInternalDrag = (e:DragEvent) => {
+        e.dataTransfer?.setData(RISU_APP_INTERNAL_DRAG_TYPE, 'true')
     }
 
 </script>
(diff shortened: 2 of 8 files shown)


cubicj (commented):
Tested on current main. Drag-and-drop reordering and folder creation both work in my environment. What browser/OS were you seeing this on?

The code changes are reasonable. checkCharOrder() was missing from createFolder, and stopPropagation correctly prevents \<main\> from overriding dropEffect.

One side effect I confirmed locally: external file drops (.risup, .risum, images) on the sidebar area no longer reach \<main\>'s ondrop, since stopPropagation fires unconditionally for all drags.

tasoo-oos:
This issue have been tested with [risuai.xyz](https://risuai.xyz/) inside brave web browser (which is chromium-based). I also use Arch linux, KDE Plasma / wayland desktop environmenet. It still persists in Version 2026.4.181 too.

Also, I just booted up windows 11 and tested the behavior there with brave browser and risuai.xyz Version 2026.4.181. I found that the feature works just fine there. So, this would be OS-level bug of some sorts.

I haven't handled external file drop feature because it did not work in my environment in the first place. I tested about that feature and this was also the one that only having issue within linux environment. I will make that taken card of too. Thanks for checking this PR out and give feedbacks!

tasoo-oos:
Fixed! The root cause was not just stopPropagation, but `<main>` in App.svelte hardcoding dropEffect = 'link' for all drags.

I found that browser drag-and-drop has two sides that must agree on what operation to perform. The drag source (e.g. the OS file manager) sets effectAllowed to declare which operations it supports as values like "copy", "move", "link", "copyMove", or "all". The drop target then sets dropEffect during dragover to request one specific operation from that allowed set. If the requested dropEffect is not included in the source's effectAllowed, the browser considers it an incompatible drop and may refuse to fire the drop event entirely.

On Windows, file drags from the OS report effectAllowed: "all", which permits every operation including link. So <main> requesting dropEffect = 'link' was technically wrong but still accepted. On Linux (KDE Plasma / Wayland / Chromium), file drags report effectAllowed: "copyMove" -- the file manager only allows copy or move, not link. Since link is not in that allowed set, Chromium refuses the drop and never fires the drop event. The difference comes from how each OS's native drag-and-drop stack advertises allowed operations to the browser, not from any web API difference.

Fixed in b81b0f6c: `<main>` now uses dropEffect = 'copy' when dataTransfer.types includes Files, and keeps link for everything else. copy is universally compatible -- every OS file drag source includes copy in its allowed set.

The stopPropagation side effect you noted (file drops over sidebar elements being swallowed) is also fixed at a612ee98 with conditional returning early logic that prevents unnecessary `stopPropagation`.

Tested with web version deployment, both in Linux / Windows.

cubicj (commented):
Re-checked the follow-up patch. The `dropEffect = 'copy'` change looks right, and the conditional `stopPropagation` fixes the fresh external-drop case.

One small thing left: `currentDrag` doesn't seem to be cleared after a sidebar drag. Since it now decides whether sidebar handlers stop propagation, external file drops over the sidebar may still be swallowed after one internal drag.

Could you check that once?

tasoo-oos:
Fixed! You were right, `currentDrag` was not being cleared after sidebar drags, and since it became part of the conditional `stopPropagation` check, it could keep swallowing later external file drops over the sidebar.

I fixed this by clearing `currentDrag` on `dragend` and after handled internal drops. I also typed it as `DragData | null`, and the sidebar handlers now only treat a drag as internal when both conditions are true:

- `currentDrag` exists
- `dataTransfer.types` includes `application/x-risu-internal`

So even if the local state somehow became stale, normal external file drags still bubble to `<main>` and import normally.

I re-tested the concerned flow: after doing a successful sidebar drag, external file drops over the sidebar still reach the app-level import handler. I also updated the PR description to match the current behavior.
<br>

However, there is another bug.
In linux, for some images included in the main page (character assets, etc. things that are in `` tags), when it is dragged risu tries to import it. I am working on that.

tasoo-oos:
Fixed the bug of internal image dragging cause risuai import behavior too.

cubicj (commented):
I could reproduce a regression with prompt preset reordering.

`markInternalDrag` is attached to `<main>`, so prompt preset drags also get `application/x-risu-internal`. Since the prompt preset drop targets don't stop propagation, `<main>` sets `dropEffect = 'none'` afterward and the drop no longer works.

The sidebar path looks fine, but this image-drag fix seems broader than intended.

tasoo-oos:
Thanks for the review!

I'll look into that.

tasoo-oos:
Fixed!
Sorry for taking too long, my life got a bit more busy all of a sudden.

The issue was that the app-level internal drag marker was still too broad. I changed the drag handling to use scoped drag types per feature instead of relying on one generic internal marker:
- sidebar: `RISU_SIDEBAR_DRAG_TYPE`
- prompt template: `RISU_PROMPT_DRAG_TYPE`
- bot preset: `RISU_PRESET_DRAG_TYPE`
- TriggerV2 triggers/effects: `RISU_TRIGGER_DRAG_TYPE` / `RISU_EFFECT_DRAG_TYPE`
- app-level internal marker: only used to prevent in-app drags from entering the file import path

The prompt/preset/trigger/effect drop targets now stop propagation after validating their own drag type, so `<main>` no longer overrides their `dropEffect`.

cubicj (commented):
Sorry for the extra back-and-forth. The scoped drag types fixed the prompt preset reordering regression on my side.

The only remaining issue I found is TriggerV2 trigger/effect auto-scroll in long lists. It looks like the new stopPropagation on the drop targets prevents the parent scroll container's dragover handler from running.

After that, I think this should be ready to approve.

tasoo-oos:
Done! I merged scattered drag / auto-scroll handling code snippets into single handler, `handleTriggerDragOver` and `handleEffectDragOver`.

cubicj (approved):
Sorry for the repeated follow-ups on this PR, and thanks for sticking with it.

Verified the latest TriggerV2 follow-up. Auto-scroll now runs through the scoped dragover handlers before propagation is stopped, while trigger/effect reordering keeps the same scoped drop handling.

sub-hub (approved):
I've heard reports of in-app drag-and-drop operations accidentally triggering file imports. This PR handles that well by tagging each drag operation with a custom MIME type via dataTransfer.setData(), so the file-import handler in App.svelte can reliably distinguish internal drags from actual file drops.

However, I was concerned that someone adding a new drag feature might not be aware of this pattern and skip the type check. I added the comments so they'll discover the convention when they look at this file and follow it consistently.

Thanks for the PR.

PR #1403, opened 2026-04-19
feat(tts): plugin hooks for preprocessing and postprocessing

hashcoko:
## PR Checklist

- Required Checks

## Summary

Two new Plugin API v3 methods — `addTTSPreprocessor` and `addTTSPostprocessor` — that let plugins transform TTS input text and output audio. Bundled with a `playAudio` refactor that fixes `stopTTS()` for four providers, and an `allowPlugins` opt-in on `runLLMModel` so plugins can legitimately reach a user's plugin-supplied main or aux model when they need to.

## Related Issues

None.

## Changes

**Plugin API v3** (`src/ts/plugins/apiV3/v3.svelte.ts`, `risuai.d.ts`)
- `addTTSPreprocessor(func)` / `addTTSPostprocessor(func)` — new hooks with full JSDoc and TypeScript types. Sequential pipeline, error isolation, skip short-circuit. No enforced timeout — same trust model as `addRisuScriptHandler` / `addRisuReplacer`.
- `runLLMModel` gains an optional `allowPlugins` parameter. Default stays `false` (blocked, same as today); plugins that need the user's plugin-supplied main or aux model can opt in explicitly.

**TTS runtime** (`src/ts/process/tts.ts`, new `src/ts/process/ttsHooks.ts`)
- Extract a shared `playAudio(audio, mimeType, ctx)` helper and route every audio-buffer provider through it.
- As a side benefit `stopTTS()` now reliably stops NovelAI, HuggingFace, GPT-SoVITS, and FishSpeech. Each previously declared a local `const sourceNode` inside its `decodeAudioData` callback, shadowing the module-level `let` so `stopTTS()` could not find the node to stop.
- Defensive buffer slice inside `playAudio` so the plugin sandbox's ArrayBuffer transfer does not leave the main thread with a detached buffer.

**Adjacent fixes folded into the refactor commit**
- `FixNAITTS` threw `TypeError` when `naittsConfig` was undefined.
- `fishspeech` case was missing its terminal `break`.

**Docs**
- New "TTS Hooks" section in `plugins.md` with two worked examples.
- Short note in `migrationGuide.md` under "API 3.0".

## Impact

- **Existing behavior is preserved.** The refactor does not change how any provider synthesizes audio; it only consolidates the decode-and-play sequence. `allowPlugins` defaults to `false`, so `runLLMModel` behaves exactly as before for existing plugins.
- **`stopTTS()` now works on every audio-buffer provider.** Before this PR only `elevenlab`, `VOICEVOX`, and `openai` actually stopped on user request. Four other providers silently ignored it.
- **No permission prompt on the new hooks** — same trust model as `addRisuScriptHandler`. Hooks that call slow services (aux LLMs, remote audio processors) are responsible for their own cancellation via `AbortController`.
- **`allowPlugins` is opt-in.** The existing blanket block inside `runLLMModel` prevents a common legitimate case: the user's main or aux model is supplied by a provider plugin (Copilot integrations, OpenRouter-via-plugin, custom proxies). Opt-in preserves the loop-protection default while unblocking those workflows. Opting-in plugins take responsibility for avoiding provider-to-provider call loops.

## Additional Notes

- `webspeech` calls only the preprocessor (no audio buffer exists) and `vits` goes through `runVITS` with its own playback path and bypasses both hooks. Both are documented in `plugins.md`.
- Manual smoke: web build verified — NovelAI plays unchanged; a preprocessor plugin using `allowPlugins: true` successfully routed through a plugin-provided aux model to rewrite a roleplay message into clean Japanese, and the rewritten text is what NovelAI spoke. Tauri desktop and node self-hosted builds are not yet smoke-tested — every changed path goes through shared infrastructure (`globalFetch`, `AudioContext`, plugin iframe postMessage) so target-specific regressions would be surprising, but happy to add those checks if preferred before merge.
- Commits are split so each stands alone. The first commit is a self-contained `stopTTS()` bug fix and could be reviewed / merged independently. If you'd rather see the `allowPlugins` change as its own PR, happy to split it out.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -951,6 +951,56 @@ await Risuai.setCustomTextTheme({
 });
 ```
 
+### TTS Hooks
+
+Risuai's plugin API lets you intercept Text-to-Speech just before synthesis and just before playback.
+
+#### `Risuai.addTTSPreprocessor(func)`
+
+Runs before TTS synthesis for every message, across every provider (including `webspeech`). The hook receives the text that is about to be spoken along with the active provider's name and the character id, and may return a replacement text or abort playback.
+
+```javascript
+//@name simple-tts-prefix
+//@api 3.0
+
+(async () => {
+  await Risuai.addTTSPreprocessor(async (ctx) => {
+    return { text: '[narrator] ' + ctx.text };
+  });
+})();
+```
+
+#### `Risuai.addTTSPostprocessor(func)`
+
+Runs after the provider returns its audio, just before it is decoded and played. The hook receives the raw encoded bytes (mp3, wav, etc.) plus the MIME type. Return a replacement `{ audio, mimeType }` to swap the audio, `{ skip: true }` to suppress playback, or `undefined` / `void` to pass through unchanged.
+
+Does **not** run for the `webspeech` provider (the browser synthesizes and plays internally, so there is no audio buffer to intercept) or the `vits` provider (uses its own playback path).
+
+```javascript
+//@name volume-normalize-tts
+//@api 3.0
+
+(async () => {
+  await Risuai.addTTSPostprocessor(async (ctx) => {
+    // Decode to PCM inside the plugin iframe.
+    const audioCtx = new AudioContext();
+    const decoded = await audioCtx.decodeAudioData(ctx.audio);
+    // ... analyse peak amplitude, produce a normalised Float32Array,
+    //     re-encode to wav bytes as `newBytes` ...
+    return { audio: newBytes, mimeType: 'audio/wav' };
+  });
+})();
+```
+
+#### Pipeline semantics
+
+- **Sequential.** Hooks run in registration order; each receives the previous hook's output.
+- **Error isolation.** If a hook throws, its result is discarded and the next hook runs. TTS playback itself is not interrupted.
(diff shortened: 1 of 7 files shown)


hashcoko:
Addressed in follow-up commits:

- **Defensive array copy in getters** (`ec0e241e`) — `getTTSPreprocessors` / `getTTSPostprocessors` now return `.slice()` so iteration in `runHookPipeline` is stable even if a plugin unloads mid-pipeline.
- **Per-hook audio slice in postprocessor pipeline** (`623f825a`) — extracted `runPostprocessorPipeline` and now clone the ArrayBuffer **per hook** instead of once per pipeline, so the second postprocessor onward no longer receives a detached buffer after the plugin sandbox's transfer.
- **GPT-SoVITS volume path now runs postprocessors** (same commit `623f825a`) — the volume != 1.0 branch used to bypass playAudio and therefore skip the hook chain. The new helper is called up-front and the GainNode stage is applied after, so hooks now fire consistently across providers and per-character volume is preserved.
- **FixNAITTS populates the full default shape** (`afc86599`) — seeds `customvoice`, `voice`, and `version` so the NovelAI request URL never templates in `undefined` (FixNAITTS is currently exported-but-unused; this is a preventative consistency fix).

The test suite (148 passing) stays green after each commit.

PR #1404, opened 2026-04-19
feat(tts): custom baseURL for OpenAI-compatible TTS providers

hashcoko:
## PR Checklist

- Required Checks

## Summary

Let each character override the OpenAI TTS endpoint (base URL, API key, model, voice, response format) so self-hosted OpenAI-compatible TTS servers (openedai-speech, Kokoro-FastAPI, custom wrappers around newer Japanese TTS models, etc.) can plug into Risuai as the native `openai` provider without a plugin. Default behavior is unchanged — existing characters keep hitting `api.openai.com` with `tts-1` exactly as today.

## Related Issues

None.

## Changes

**Type + runtime** (`src/ts/storage/database.svelte.ts`, `src/ts/process/tts.ts`)
- Add `character.oaiTTSConfig { enabled, baseURL, apiKey, model, voice, format }` to the character type and the groupChat lazy-typecheck block.
- Rewrite the openai case in `tts.ts` around a fallback chain: `cfg` (only read when `oaiTTSConfig.enabled === true`) → legacy `character.oaiVoice` + `db.openAIKey` → hardcoded defaults (`https://api.openai.com/v1`, `tts-1`, `alloy`, `mp3`).
- Omit the `Authorization` header entirely when both the per-character override and the global `openAIKey` are empty, so unauthenticated local servers work out of the box.
- Trim trailing slashes on `baseURL` so `https://host/v1/` is safe.

**UI** (`src/lib/SideBars/CharConfig.svelte`)
- Add an "Advanced (OpenAI-compatible endpoint)" `<Check>` toggle inside the existing openai branch of the character TTS settings. With it off, the UI is pixel-identical to before: just the voice dropdown bound to `oaiVoice`. With it on, the dropdown is replaced by a freeform voice text input (placeholder shows the current fallback value), and four more fields appear: Base URL, API Key, Model, Response Format.
- The API key input uses `hideText={DBState.db.hideApiKey}` to respect the user's existing "hide API keys" preference.
- Response Format select includes an explicit "mp3 (default)" option with an empty value, so users can express "use the default" without storing a specific choice.
- A new `$effect.pre` initializer matches the existing NAI / GPT-SoVITS / FishSpeech pattern: on first entry into `ttsMode === 'openai'`, set `oaiTTSConfig = { enabled: false }` so the Advanced checkbox has a defined target to bind to. Every other override field stays undefined; the runtime fallback chain handles the rest.

## Impact

- **Byte-for-byte backward compatible** for existing characters. `oaiTTSConfig.enabled` defaults to `false`; with it unset or false, the pipeline ignores the new object entirely and behaves exactly like today — same URL, same key, same `tts-1`, same `alloy` / user-selected voice, same implicit mp3.
- **Unblocks local TTS servers.** The two most common shapes (openedai-speech / Kokoro-FastAPI style, and custom wrappers around newer Japanese TTS models) both expose an OpenAI-compatible `/v1/audio/speech`. Today users hit the hardcoded `api.openai.com` and can't point Risuai at these without writing a provider plugin. After this change they tick Advanced, paste the URL, and pick a voice ID the local server knows.
- **No CORS change.** The request continues to go through `globalFetch`; only the destination URL differs.
- **No schema migration.** The new object is additive and guarded by `enabled`; absent saves keep working.

## Additional Notes

- Not unit-tested. The openai case is pure fetch-and-play with a straightforward fallback chain — mocking `globalFetch` + `AudioContext` adds no confidence vs. the 148 existing unit tests (none regress). Manually smoke-tested in the web dev server: (1) official OpenAI with an existing `oaiVoice` → regression OK; (2) Advanced ON + a local OpenAI-compatible Japanese TTS endpoint (no API key) → plays correctly; (3) toggling Advanced OFF → ON → OFF preserves both `oaiVoice` and `oaiTTSConfig` values so users don't lose their settings when flipping modes.
- Tauri desktop and node self-hosted builds were not smoke-tested manually. Every path in this diff flows through the shared `globalFetch` + `AudioContext` infrastructure, so target-specific regressions would be surprising — happy to add those checks if reviewers want boxes ticked before merge.
- Commits are split so the type/runtime change and the UI change can be reviewed or reverted independently.


hashcoko:
Addressed in follow-up commits:

- **`baseURL` is now trimmed** (`e7708d30`) — leading/trailing whitespace no longer reaches `new URL(url)` inside globalFetch. An all-whitespace value falls back to the default.
- **Empty-value option removed from the Response Format select** (`a2048f02`) — every dropdown value is now a valid codec literal. The `$effect.pre` initializer seeds `format: 'mp3'` for a character freshly switched to the openai provider so the dropdown binds to a typed default instead of an empty string.

Regarding the CodeQL "Insecure randomness" alert on `tts.ts:118`: that line is `const apiKey = (cfg?.apiKey || db.openAIKey || '').trim()` — this PR does not introduce any `Math.random()` call. A quick repo-wide grep shows the closest `Math.random()` usages are in `src/ts/plugins/apiV3/factory.ts` (iframe request IDs) and `src/ts/globalApi.svelte.ts` — i.e. pre-existing code unrelated to this diff. The most plausible explanation is that CodeQL's taint-flow analysis re-computed a path from an existing `Math.random()` source to a new sink (the apiKey line), which would make this a pre-existing latent finding surfaced by the diff rather than a regression introduced here. Happy to investigate further or mark as false positive per your call.

PR #1405, opened 2026-04-19
fix: correct receiver allowedIPC check in postPluginChannelMessage

rsyumi:
## PR Checklist

- Required Checks

## Summary

Fix the `allowedIPC` check in `postPluginChannelMessage` to use the sender's name instead of the receiver's name.

## Related Issues

None.

## Changes

In `postPluginChannelMessage`, changed `receiverPlugin.allowedIPC?.includes(pluginName)` to `receiverPlugin.allowedIPC?.includes(currentPluginName)` so that the receiver's `allowedIPC` list is checked against the sender's name as intended.

## Impact

Plugin-to-plugin IPC now works as intended. This is a single-variable change with no other side effects.

## Additional Notes

None.

### DIFF
--- a/src/ts/plugins/apiV3/v3.svelte.ts
+++ b/src/ts/plugins/apiV3/v3.svelte.ts
@@ -1209,7 +1209,7 @@ const makeRisuaiAPIV3 = (iframe:HTMLIFrameElement,plugin:RisuPlugin) => {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


cubicj (approved):
Verified the IPC allowlist path. This correctly checks the receiver's allowlist against the sender plugin name.

sub-hub (approved):

PR #1420, opened 2026-04-27
fix: remove duplicate anthropic beta declaration

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Removes a duplicate `betas` declaration in the Anthropic request builder that caused the production build parser to fail.

## Related Issues

None

## Changes

- Removed the earlier `betas` calculation block before additional parameters are applied.
- Kept the later `betas` calculation after `applyAdditionalParameters`, preserving custom `header::anthropic-beta` precedence.

## Impact

This fixes the build parse error without changing the intended Anthropic beta header behavior. Custom `anthropic-beta` headers supplied through additional parameters continue to take precedence over auto-generated beta headers.

## Additional Notes

Validation:

- `rg -n "let betas" src/ts/process/request/anthropic.ts` now reports a single declaration.
- `pnpm check` was run, but it still fails on pre-existing unrelated diagnostics in `src/lang/zh-Hant.ts` and the existing `src/lib/UI/GUI/SliderInput.svelte` a11y warning.

### DIFF
--- a/src/ts/process/request/anthropic.ts
+++ b/src/ts/process/request/anthropic.ts
@@ -567,21 +567,6 @@ export async function requestClaude(arg:RequestDataArgumentExtended):Promise<req
         delete headers["x-api-key"]
     }
 
-    let betas:string[] = []
-
-    if(body.max_tokens > 8192){
-        betas.push('output-128k-2025-02-19')
-    }
-
-
-    if(db.claude1HourCaching){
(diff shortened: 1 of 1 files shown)


sub-hub (approved):
I also confirmed that the current docker build faillure is causeed by a duplicate `betas` declaration.
This error appears to have occurred during the merging process, It's not anyone's fault.

Anyway, thank you for the quick PR!

PR #1422, opened 2026-04-28
feat: optimize streaming display rendering

SameDesu123:
## PR Checklist

- Required Checks
       - This PR is limited to an opt-in streaming display performance setting and related render-path changes.

## Summary

Adds an experimental, opt-in Streaming Display Optimization setting for long streamed assistant responses. The default `off` mode keeps the existing rendering behavior, while `balanced` and `strong` reduce UI churn during active streaming.

## Related Issues

None.

## Changes

- Added a persisted `streamingDisplayOptimizationMode` setting with `off`, `balanced`, and `strong` options in Advanced Settings.
- Added labels and help text for all supported locales.
- Batched active streaming display updates in `balanced` and `strong` modes, with pending output flushed on completion or abort.
- Kept the active streaming message row identity stable while optimization is enabled to reduce remount churn.
- Moved streaming optimization eligibility to the parent chat list and passes plain props only to the active streaming assistant message.
- Added a raw streaming display path in `ChatBody` for `strong` mode to skip markdown parsing while the active assistant message is still streaming.

## Impact

The feature is disabled by default, so existing behavior should remain unchanged unless users enable the new setting. When enabled, long streamed responses can update the UI less frequently during generation, which may reduce render overhead on large chats or lower-end devices.

## Additional Notes

Validation: `pnpm check` reports 0 errors and 0 warnings.


kwaroran:
There seem to be too many effect hooks for a performance update.

Although we can expect better performance when this option is enabled when the update rate is high,

the PR adds a lot of hooks that run when the variable is updated even when the option is disabled, which can worsen performance when the option is disabled, or even if the update rate is not high enough with option enabled

SameDesu123:
> There seem to be too many effect hooks for a performance update.
> 
> Although we can expect better performance when this option is enabled when the update rate is high,
> 
> the PR adds a lot of hooks that run when the variable is updated even when the option is disabled, which can worsen performance when the option is disabled, or even if the update rate is not high enough with option enabled

Thanks, I think this concern is valid.

I’ll refactor this so the optimization state is computed once at the parent/current-chat level, and only the active streaming assistant message receives the optimized rendering props. When the mode is `off`, the normal rendering path should remain effectively unchanged.

cubicj (commented):
The optimization direction looks real. `balanced` / `strong` coalesce streaming display updates, reduce `reloadKeys` churn, avoid remounting the active streaming message on every data update, and `strong` skips `ChatBody` markdown/display parsing while streaming. `pnpm check`, `pnpm test`, and `pnpm build` pass locally.

I still found two concrete issues. `adv.streamingDisplayOpt` has `showExperimental`, but no `condition: (ctx) => ctx.db.useExperimental`, so it is visible even when experimental settings are disabled.

Also, the `strong` help text says post-processing is skipped during streaming and only tried once after the stream finishes, but the implementation still runs `processScriptFull(..., 'editoutput')` on every coalesced flush. That still executes Lua edit triggers, plugin edit hooks, and regex scripts during streaming, just less frequently.

Could you check those and rebase the branch? It currently conflicts with `main`.

SameDesu123:
Thanks for checking this. I rebased the branch onto the latest main and resolved the conflict.

I also addressed both streaming-display comments. 
1. The option only appears when `experimental settings display` are enabled.
2. `strong` mode now runs `editoutput` once after the stream finishes.

Validation: `pnpm check` and `git diff --check` both pass.

cubicj (commented):
Thanks for the follow-ups. The rebased head covers both earlier points, and `pnpm check`, `pnpm test`, and `pnpm build` pass locally.

I still found two issues.

In `strong` mode, the final `editoutput` pass is skipped when the stream is aborted, so the aborted message keeps the raw text in saved chat data and `editoutput` scripts never apply to it. In `off` and `balanced`, an aborted message keeps the post-processed content up to the abort point.

The `ChatBody` parse cache skips `processScriptFull(..., 'editdisplay')` on a hit, but display triggers, plugin hooks, and `presetRegex`/module regex are not part of the cache key, so those can render stale results. The cache also runs even when the setting is `off`. Personally I think `off` should keep the existing rendering path unchanged, especially for an experimental feature.

Could you check those?

cubicj (approved):
I went back to the abort path since that was the actual gap. Interrupting a strong-mode stream now runs the deferred `editoutput` in a `finally`, so the saved message ends up post-processed like `off` and `balanced` instead of raw. The parse cache is gone too, so `off` is back to the plain render path and editdisplay and plugin regex invalidate normally. `isStreaming` also stays inside the streaming path and the pending flush drains on both finish and abort.

`pnpm check`, `pnpm test`, and `pnpm build` pass locally.

Approving — thanks for working through all the follow-ups here!

SameDesu123:
Thanks again for the review and approval. I did one additional compatibility pass after the approval and pushed `7de191a` with three follow-up fixes:

- Scheduled display-flush failures are now propagated back through the chat request path instead of becoming detached promise rejections.
- Copy and TTS use the latest streamed text, while editing, translation, and partial editing are disabled only while the optimized message is actively changing. This keeps the strong-mode per-flush parsing savings intact.
- The display optimization mode is pinned when each stream starts, so changing the setting mid-stream cannot make request processing and UI rendering follow different modes.

Validation:
- `pnpm check` — 0 errors and 0 warnings
- `pnpm test` — 192 passed, 3 skipped
- `git diff --check` — passed

cubicj (commented):
I went through the post-approval delta too. The main thing I wanted to check was where a failed display flush lands now that it cancels the stream, and it ends up in the existing alert path with the chat state released, from the send screen and the plugin API both. Pinning the mode per stream also holds up in both directions when the setting changes mid-generation, and the edit and translate buttons come back cleanly once the finished row remounts.

`pnpm check`, `pnpm test`, and `pnpm build` pass on the current head. Still good from my side.

SameDesu123:
Fixing some merge conflicts!

SameDesu123:
@sub-hub Sorry for mention. I think this function is ready to use. Could you review this?

sub-hub (approved):
Image

Image

Both modes are working well!
It also appears that the issue where performance was affected when the feature was turned off has been resolved.

It seems like a great feature, as it gives users the choice of whether to use it and addresses the significant demand related to streaming.
Thanks for the PR.

PR #1444, opened 2026-05-12
fix: fallback missing realm card image

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Adds a fallback image for character card exports when the character has no readable avatar image.

## Related Issues

None

## Changes

- Added a character card export image resolver that first tries the character image.
- Falls back to `/none.webp` when the character image is missing, empty, or unreadable.
- Keeps the fallback limited to export/upload output without mutating the character or database image field.

## Impact

Realm uploads and character card exports no longer fail solely because a bot card has no profile image. Existing characters with valid images continue to export with their original image.

## Additional Notes

Validated with `pnpm check`.
Test it manually in dev server.

### DIFF
--- a/src/ts/characterCards.ts
+++ b/src/ts/characterCards.ts
@@ -712,6 +712,28 @@ export async function exportChar(charaID:number):Promise<string> {
     return ''
 }
 
+async function readCardExportImage(image?: string) {
+    if(image){
+        try{
+            const img = await readImage(image)
+            if(img?.length){
+                return img
+            }
+        } catch {
+        }
+    }
+
+    const res = await fetch('/none.webp')
+    if(!res.ok){
+        throw new Error(`Failed to load fallback character image: ${res.status} ${res.statusText}`)
+    }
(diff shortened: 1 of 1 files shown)


cubicj (approved):
Verified the export path. This reuses the existing `/none.webp` placeholder for missing card images without mutating the character data.

sub-hub:
The code looks fine! You’ve done a great job fixing the bug.

It is already possible to export bot cards without images locally (the reason it works locally but not on Realm is that the fix in #1128 only addressed local exports). You can also upload bot cards without images to Realm by visiting the site directly.

To summarize, this PR enables the direct upload of cards without images to Realm from risu.

However, I have some concerns about making it so easy to upload bot cards without images.
I might be overthinking it, but I worry that this fix could lead to a flood of bot cards lacking images.
If we could simply modify the Realm code to implement a filter later on, it wouldn't be an issue; however, since we don't have that authority, I would really like to hear the opinion of the developer who manages Realm. @kwaroran

PR #1445, opened 2026-05-13
fix: harden desktop file picker flows

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Refreshes the dropped file picker hardening work from #1324 by @tasoo-oos against current `main`.

The shared file picker helpers now use the Tauri native dialog on desktop, return consistent cancel values, and keep the DOM fallback safer for WebKit-style picker behavior.

## Related Issues

Follows up on #1324.

## Changes

- Hardened `selectSingleFile` and `selectMultipleFile` so cancel handling is explicit and consistent.
- Replaced remaining direct file input import flows with the shared file picker helpers.
- Moved local backup file selection to the user click path while preserving current database, asset, and cold-storage restore behavior.
- Added desktop file read scopes for common user directories and declared macOS bundle localizations.

## Impact

Desktop users should get native file dialogs for import and local backup flows, including Tauri builds where the DOM picker can fail. Web users keep the DOM fallback, with safer cancel handling and offscreen input styling.

Cancelling a picker should no longer crash null-unsafe import paths.

## Additional Notes

Validated with `pnpm check`, YAML parse check, and `git diff --check`.


SameDesu123:
It works successfully on macOS. Although I haven't been able to test it in a Linux environment, it is expected to function correctly.

If you want to test it, you can get bundled file in here: https://github.com/SameDesu123/Risuai/actions/runs/25807800518

tasoo-oos:
Hello, Thank you for finishing the task I had given up on!

I ran the appimage build from Linux.
Persona, Character card, Prompt (.risup), module (.risum), and local backup (.bin) Export / Import jobs are tested and it worked just fine.

For context, here are the environment that this test was conducted:
- OS: Arch Linux
- Kernel: Linux 7.0.3-arch1-2
- Architecture: x86_64
- Device: Samsung 940XFG laptop
- CPU/GPU platform: Intel Raptor Lake-P / Iris Xe Graphics
- Desktop: KDE Plasma 6.6.4
- Session: Wayland
- Display env: `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=KDE`, `WAYLAND_DISPLAY=wayland-0`, `DISPLAY=:1`
- Graphics stack: Mesa 26.0.6, libglvnd 1.7.0
- GTK/WebKit: gtk3 3.24.52, webkit2gtk-4.1 2.52.3
- Portal packages: xdg-desktop-portal 1.20.4, xdg-desktop-portal-kde 6.6.4

Just a few minor issues that are not necessarily have to be included to this PR:
1) Importing character card by external file dropdown does not work.
2) The AppImage itself hit an EGL startup issue when using its bundled libraries `Could not create default EGL display: EGL_BAD_PARAMETER`.
Running the extracted binary with system libraries worked: `LD_LIBRARY_PATH=/usr/lib ./RisuAI.AppDir/usr/bin/RisuAI`

SameDesu123:
Thanks for testing this on Linux.

1st Issue seems related to the same import path handling problem, so I’ve fixed it together in this PR.
https://github.com/SameDesu123/Risuai/actions/runs/25860019091

For second (EGL_BAD_PARAMETER with the bundled AppImage libraries), I think it would be better to track it separately in another PR/issue since it looks more related to AppImage packaging/runtime behavior than the file picker changes themselves.

tasoo-oos:
Agreed on those choices.

cubicj (commented):
Thanks for reviving this. I took another pass through the file picker changes and found a few side effects that look worth fixing before review.

First, the new `fs:scope` entries in `migrated.json` look broader than the read use case. This capability already grants `fs:allow-write-file`, `fs:allow-remove`, `fs:allow-rename`, `fs:allow-copy-file`, and `fs:allow-mkdir`, so adding `$HOME/**/*`, `$DESKTOP/**/*`, `$DOCUMENT/**/*`, `/mnt/**/*`, `/media/**/*`, and `/run/media/**/*` also expands write/remove/rename access over those user directories, not just read access for files selected from the native dialog.

Second, the new local backup restore path uses handle-based fs APIs: `openFile()`, `file.stat()`, and `file.read()`. The current capability has `fs:allow-read-file`, but that maps to the standalone `read_file` command. The handle path invokes `fstat` and `read` too, and those are not covered by `fs:allow-read-file`, so Tauri desktop restore can still fail after the user selects a `.bin` file.

One more regression risk: some import paths now read the selected file into `Uint8Array` before passing it on. Character import used to pass the `File` into `importCharacterProcess()`, and `CharXImporter` can stream `File` / `ReadableStream` input. With `selectMultipleFile()`, all selected character files are read into memory first. The subtitle media path also reads the file into bytes and then wraps it back into a `File`.

Also, the current conflict in `backuplocal.ts` does not look like a mechanical rebase conflict. `main` now has local backup encryption metadata handling and stricter cold-storage restore handling, while this PR refactors the same restore path into chunk readers. That logic needs to be carried into the refactored restore flow when the branch is rebased.

Could you take another look at these together? The native picker direction looks useful, but I think these side effects should be handled in the same follow-up.

SameDesu123:
Thanks for the close read. I rebased this on the current `main` and reworked the follow-up around the issues you called out.

A few concrete changes:

1. Removed the broad `$HOME`, `$DESKTOP`, `$DOCUMENT`, mounted-drive, and recursive download fs scopes. The PR no longer expands those directories under the existing write/remove/rename permissions.

2. Dropped the handle-based restore path using `openFile()`, `stat()`, and `read()`. Local backup restore now stays on the browser `File.stream()` path, so it does not need the extra fs handle permissions.

3. Kept the current `main` restore behavior while resolving the rebase conflict, including `encryption.risudat` handling and the stricter cold-storage restore checks.

4. Avoided preloading large character/subtitle inputs into `Uint8Array` where the downstream code can work with a `File`/stream. Character import and drag/drop import now pass the `File` through again.

I also ran:

- `pnpm check`

Both are clean now.

cubicj (approved):
Looks good. I rechecked the follow-up against the previous file picker concerns, and the broad fs scopes and handle-based restore path are gone. Local backup restore stays on the `File.stream()` path, and the current encryption/cold-storage restore handling from `main` is still preserved.

Character import and subtitle media selection also avoid preloading large inputs into `Uint8Array` where a `File` path is available. `pnpm check`, `pnpm test`, and `pnpm build` pass locally, and CodeQL is green.

Approving — thanks for the careful follow-up!

PR #1440, opened 2026-05-11
chore: pin pnpm to latest v10

SameDesu123:
## PR Checklist

- Required Checks
    - [ ] Have you added type definitions?
    - [x] Have you tested your changes?
    - [x] Have you checked that it won't break any existing features?
- [ ] If your PR uses models[^1], check the following:
    - [ ] Have you checked if it works normally in all models?
    - [ ] Have you checked if it works normally in all web, local, and node-hosted versions? If it doesn't, have you blocked it in those versions?
- [x] If your PR is highly AI generated[^2], check the following:
    - [x] Have you understood what the code does?
    - [x] Have you cleaned up any unnecessary or redundant code?
    - [x] Is it not a huge change?
       - We currently do not accept highly AI generated PRs that are large changes.

[^1]: Modifies the behavior of prompting, requesting, or handling responses from AI models.
[^2]: Over 80% of the code is AI generated.

## Summary

Pins package-manager resolution to pnpm 10.34.1 so CI, Docker, and server bootstrap scripts do not unintentionally move to pnpm v11.

## Related Issues

Fixes #1439

## Changes

- Added `packageManager: "pnpm@10.34.1"` to `package.json`.
- Changed GitHub Actions pnpm setup from `latest` to `10.34.1`.
- Changed Docker Corepack installation from `pnpm@latest` to `pnpm@10.34.1`.
- Changed Unix and Windows server bootstrap scripts to install `pnpm@10.34.1`.

## Impact

This keeps installs on pnpm 10.34.1 while avoiding an accidental pnpm v11 upgrade. That preserves compatibility with the project's current Node.js 20 support until a dedicated pnpm v11 migration is ready.

## Additional Notes

Validation performed:

- `node -e "JSON.parse(require('fs').readFileSync('package.json','utf8')); console.log('package.json ok')"`
- `git diff --check upstream/main..HEAD`
- `rg -n "pnpm@latest-10|latest-10" .github Dockerfile server.sh server.bat package.json`

PR #1451, opened 2026-05-17
Improve self-host server setup docs

tasoo-oos:
## PR Checklist

- Required Checks
    - [x] Have you checked that it will not break any existing features?
    - [ ] Have you checked if it works normally in all web, local, and node-hosted versions? If it does not, have you blocked it in those versions?

## Summary

Improve the self-hosting server script documentation.

## Related Issues

None.

## Changes

- Added documented optional environment variables to `server.sh` and `server.bat`.
  - Documented build-time flags such as `VITE_RISU_LEGAL_CONFIGURED`, `VITE_RISU_LITE`, and optional ad configuration.
  - Documented runtime flags such as `NODE_ENV`, `PORT`, and `TRUST_PROXY`.

## Impact

Self-hosted users and developers get clearer guidance for configuring `server.sh`.

No application runtime behavior is intended to change.

### DIFF
--- a/server.bat
+++ b/server.bat
@@ -1,4 +1,43 @@
-call npm -g install pnpm
+@echo off
+setlocal
+
+REM Optional build-time flags. Uncomment or set them before running this script.
+REM These values are baked into the frontend during `pnpm run build`.
+REM
+REM Legal configuration:
+REM Set this to TRUE only for a personal private self-hosted instance, a development/testing
+REM fork intended to contribute back upstream, or a fork whose Terms of Service,
+REM Privacy Policy, and Risuai service usage alerts have been configured correctly.
+REM Builds without this flag show a legal warning screen and should not use Risuai
+REM services that require the original legal notices.
+REM if not defined VITE_RISU_LEGAL_CONFIGURED set "VITE_RISU_LEGAL_CONFIGURED=TRUE"
+
+REM Lite mode:
+REM Enables the experimental lightweight/mobile-oriented UI mode at build time.
+REM if not defined VITE_RISU_LITE set "VITE_RISU_LITE=TRUE"
+
+REM Advertisement configuration:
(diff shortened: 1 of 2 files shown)


sub-hub:
Lately, some people have been asking about VITE_RISU_LEGAL_CONFIGURED, and it's great that you can set it up intuitively just by uncommenting various variables.

However, there seems to be an issue regarding the changes to pnpm-workspace.yaml.
When I run pnpm install on this branch in my environment, I get the following error:
`ERR_PNPM_CONFIG_CONFLICT_BUILT_DEPENDENCIES Cannot have both neverBuiltDependencies and onlyBuiltDependencies`

Could you look into this issue?

tasoo-oos:
I guess that's `sharp` related issue? I'll look into it.

tasoo-oos:
Thanks for catching this!

The issue was caused by mixing the old `pnpm.neverBuiltDependencies` config in `package.json` with the new workspace-level build approval config. `sharp: false` in `pnpm-workspace.yaml`pnpm-workspace.yaml` already preserves the old behavior, so I removed the legacy `pnpm.neverBuiltDependencies` block from `package.json`.

I also added an explicit root package entry in `pnpm-workspace.yaml` for compatibility with older pnpm 10 versions.

<br>

However, now that `neverBuiltDependencies` is gone, pnpm earlier than v10 may build `sharp`. This won't be the issue with `server.sh` users since `npm install -g pnpm` will update pnpm to latest version, but It may not be the case such as devs who uses local package manager such as `apt` or `homebrew` for pnpm instead of `npm install` method. I am not sure if this is acceptable. How do you think?

tasoo-oos:
https://github.com/kwaroran/Risuai/pull/1440

Instead of doing anything else, I pinned pnpm version to v10 just as proposed at PR 1440. pnpm-workspace.yaml is removed.
Also, added documentation at .bat too.

PR #1452, opened 2026-05-18
feat: track granular save sections

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Adds save-section change tracking so the save loop can identify which major data areas changed before the existing `database.bin` encoder writes the compatible save blocks.

## Related Issues

None

## Changes

- Added `src/ts/storage/saveSections/sectionChangeDetection.ts` with grouped section keys and helper utilities for tracking changed save sections.
- Extended `saveDb()` change tracking for root settings, presets, characters, chats, modules, plugins, plugin storage, loadouts, and existing bot resource sections.
- Preserved the existing save flags used by `RisuSaveEncoder`, so current `database.bin` save compatibility remains unchanged.
- Added development-only console logging for detected save sections before a save is written.

## Impact

This does not change the persisted save format. It records more granular change information while continuing to write through the existing save blocks. The development log should help verify which data area triggered a save without affecting production builds.

## Additional Notes

Validated with `pnpm check`.


SameDesu123:
Addressed the two review points in `18fb1e15`.

- Granular list-section tracking now keeps the corresponding legacy dirty flag set while `RisuSaveEncoder` remains the source of truth for `database.bin` writes. This covers presets/modules/plugins/loadouts/plugin storage and also keeps character/chat/root compatibility guards in place.
- Preset/module/plugin/loadout granular tracking now uses per-section item signatures instead of ready sets, so an effect rerun no longer reports every existing item as changed. Initial observation only seeds signatures; later edits/additions/deletions report the actual affected item ids.

Validation: `pnpm check` passes with 0 errors and 0 warnings.

SameDesu123:
Addressed the latest follow-up in `bec60eaf`.

- Granular-only section arrays are now cleared after cloning `toSave`, so dev save-section logs do not retain stale observational section ids. Legacy trackers (`character` / `chat`) keep their existing `[first]` reset behavior.
- `trackSaveListSection()` now compares item order separately and records `__order` in the affected granular section when stable-id items are reordered.

Validation: `pnpm check` passes with 0 errors and 0 warnings.

PR #1456, opened 2026-05-20
fix: Redesign request log in response metadata menu

sub-hub:
## PR Checklist

- Required Checks

## Summary

It's currently very inconvenient to view request responses in the request logs within the metadata window.
This is because the response is located below the body, requiring a lot of scrolling to see it.

I solved this issue with a very simple fix using details and summary tags, allowing users to open only the parts they want to see without increasing complexity.

Additionally, since this window was using hardcoded text colors, I modified it to use textcolor instead.

image


cubicj (approved):
Verified the request log display path. Seeing this PR made me realize how inconvenient the always-expanded request body was, and the collapsible body/response sections solve that neatly without changing the underlying fetch log data path.

PR #1457, opened 2026-05-20
feat: Cancel Account Sync Option

sub-hub:
## PR Checklist

- Required Checks

## Summary

image

Add third option to checkAccountSync page, **Cancel Account Sync**.

## Description
Actually, users can already cancel account sync by selecting `Save Current Data to Account` and type a random string in the confirm input.
```js
            const confirm = await alertInput(`to overwrite your data, type "RISUAI"`)
            if(confirm !== "RISUAI"){
                localStorage.setItem('dosync', 'avoid')
                return false
            }
```
However, this is a very unfriendly way of doing things. Who would risk overwriting their own data just to cancel account sync?

So, I'd really like to see this PR merged. It would be much more intuitive to have a cancel button rather than telling people who don't want the sync to just overwrite and cancel the data.

### DIFF
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -926,6 +926,7 @@ export const languageEnglish = {
     logout: "Logout",
     loadDataFromAccount: "Load Data from Account",
     saveCurrentDataToAccount: "Save Current Data to Account",
+    cancelAccountSync: "Cancel Account Sync",
     chatAssumed: "",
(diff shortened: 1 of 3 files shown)


cubicj (approved):
Verified the account sync decision path. I had also run into this before, and the new cancel option gives account sync users a clear safe path instead of the overwrite-confirm workaround.

PR #1459, opened 2026-05-21
perf: optimize Lua state writes and lightweight chat access

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Reduce avoidable Lua scripting work by skipping redundant chat-state writes and adding lightweight chat access APIs. Existing Lua-facing setters keep their no-return behavior for compatibility, while new opt-in APIs expose whether a write changed state.

## Related Issues

None.

## Changes

- Skip `setChatVar` writes when the stored value is already identical.
- Keep the existing Lua-facing `setChatVar` and `setState` APIs no-return for backward compatibility.
- Add `setChatVarChanged` and `setStateChanged` for scripts that explicitly need a changed/skipped result.
- Apply the same changed/skipped behavior to trigger-local, display-mode, and persistent chat state writes.
- Add lightweight `getChatData`, `getChatRole`, and `getRecentChats` APIs as cheaper alternatives to `getFullChat`.

## Impact

This is intended to be backward compatible. Existing Lua APIs remain available with their previous return behavior, and the new APIs are additive. Lua engine caching and lifecycle behavior are unchanged; that work is deferred to #1557.

No model request or response behavior is changed.

## Additional Notes

Validated with `pnpm check`, `git diff --check`, and `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/ts/parser/tests/chatVar.svelte.test.ts`.

### DIFF
--- a/src/ts/parser/chatVar.svelte.ts
+++ b/src/ts/parser/chatVar.svelte.ts
@@ -24,12 +24,18 @@ export function getChatVar(key:string): string {
     return state.toString()
 }
 
-export function setChatVar(key:string, value:string): void {
+export function setChatVar(key:string, value:string): boolean {
     const selectedChar = get(selectedCharID)
-    if(!DBState.db.characters[selectedChar].chats[DBState.db.characters[selectedChar].chatPage].scriptstate){
-        DBState.db.characters[selectedChar].chats[DBState.db.characters[selectedChar].chatPage].scriptstate = {}
+    const chat = DBState.db.characters[selectedChar].chats[DBState.db.characters[selectedChar].chatPage]
+    chat.scriptstate ??= {}
+
+    const stateKey = '$' + key
+    if(chat.scriptstate[stateKey] === value){
+        return false
     }
-    DBState.db.characters[selectedChar].chats[DBState.db.characters[selectedChar].chatPage].scriptstate['$' + key] = value
+
+    chat.scriptstate[stateKey] = value
+    return true
 }
 
 export function getGlobalChatVar(key:string): string {
--- a/src/ts/process/modules.ts
+++ b/src/ts/process/modules.ts
@@ -31,6 +31,12 @@ export interface RisuModule{
     mcp?:MCPModule
 }
 
+export interface ModuleTriggerRuntime {
+    trigger: triggerscript
+    moduleId: string
+    index: number
+}
+
 export async function exportModule(module:RisuModule, arg:{
     alertEnd?:boolean
     saveData?:boolean
@@ -398,16 +404,24 @@ export function getModuleAssets() {
 
 
 export function getModuleTriggers() {
+    return getModuleTriggerRuntimes().map((v) => v.trigger)
+}
+
+export function getModuleTriggerRuntimes(): ModuleTriggerRuntime[] {
     const modules = getModules()
-    let triggers: triggerscript[] = []
+    let triggers: ModuleTriggerRuntime[] = []
     for (const module of modules) {
         if(!module){
             continue
         }
         if (module.trigger) {
-            triggers = triggers.concat(module.trigger.map((t) => {
+            triggers = triggers.concat(module.trigger.map((t, index) => {
                 t.lowLevelAccess = module.lowLevelAccess
-                return t
+                return {
+                    trigger: t,
+                    moduleId: module.id,
+                    index,
+                }
             }))
         }
     }
--- a/src/ts/process/modules.ts
+++ b/src/ts/process/modules.ts
@@ -31,12 +31,6 @@ export interface RisuModule{
     mcp?:MCPModule
 }
 
-export interface ModuleTriggerRuntime {
-    trigger: triggerscript
-    moduleId: string
-    index: number
-}
-
 export async function exportModule(module:RisuModule, arg:{
     alertEnd?:boolean
     saveData?:boolean
@@ -404,24 +398,16 @@ export function getModuleAssets() {
 
 
 export function getModuleTriggers() {
-    return getModuleTriggerRuntimes().map((v) => v.trigger)
-}
-
-export function getModuleTriggerRuntimes(): ModuleTriggerRuntime[] {
     const modules = getModules()
-    let triggers: ModuleTriggerRuntime[] = []
+    let triggers: triggerscript[] = []
     for (const module of modules) {
         if(!module){
             continue
         }
         if (module.trigger) {
-            triggers = triggers.concat(module.trigger.map((t, index) => {
+            triggers = triggers.concat(module.trigger.map((t) => {
                 t.lowLevelAccess = module.lowLevelAccess
-                return {
-                    trigger: t,
-                    moduleId: module.id,
-                    index,
-                }
+                return t
             }))
         }
     }
--- a/src/ts/process/scriptings.test.ts
+++ b/src/ts/process/scriptings.test.ts
@@ -1,0 +1,92 @@
+// @vitest-environment node
+
+import { readFile } from 'node:fs/promises'
+import { resolve } from 'node:path'
+import { beforeAll, expect, test, vi } from 'vitest'
+
+vi.mock('../parser/parser.svelte', () => ({
+  hasher: vi.fn(),
+  risuChatParser: vi.fn(),
+}))
+
+vi.mock('../alert', () => ({
+  alertConfirm: vi.fn(),
+  alertError: vi.fn(),
+  alertInput: vi.fn(),
+  alertNormal: vi.fn(),
+  alertSelect: vi.fn(),
+}))
+
+vi.mock('../globalApi.svelte', () => ({ fetchNative: vi.fn(), readImage: vi.fn() }))
+vi.mock('../tokenizer', () => ({ tokenize: vi.fn() }))
+vi.mock('../util', () => ({
+  asBuffer: vi.fn(),
+  getPersonaPrompt: vi.fn(),
+  getUserIcon: vi.fn(),
+  getUserName: vi.fn(),
+}))
+
+vi.mock('../storage/database.svelte', () => ({
+  getCurrentCharacter: vi.fn(() => ({})),
+  getCurrentChat: vi.fn(() => ({ message: [] })),
+  getDatabase: vi.fn(() => ({ characters: [] })),
+  setDatabase: vi.fn(),
+}))
+
+vi.mock('../stores.svelte', () => ({
+  DBState: { db: {} },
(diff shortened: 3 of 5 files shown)


cubicj (commented):
One compatibility concern: returning `true` or `false` from the Lua-facing `setChatVar()` / `setState()` is useful, but it changes the behavior of existing setter APIs. Existing scripts may call these setters in return position assuming they return nothing, and that can now leak a boolean into the trigger result.

Could you check the compatibility impact here?

SameDesu123:
@cubicj Thanks for catching this. I pushed a follow-up commit that keeps the compatibility behavior for the existing Lua-facing setters: `setChatVar()` and `setState()` no longer return the internal boolean, so existing scripts like `return setState(...)` will not leak a boolean trigger result or accidentally stop generation.

For scripts that explicitly need to know whether a write changed state, I added opt-in APIs: `setChatVarChanged()` and `setStateChanged()`. The trigger-local setter path now also reports changed/unchanged consistently for those opt-in calls.

Validated with:
- `pnpm check`
- `git diff --check`
- `pnpm test -- src/ts/parser/tests/chatVar.svelte.test.ts`

cubicj (approved):
Verified the follow-up. Thanks for taking the compatibility feedback. Existing Lua-facing `setChatVar()` / `setState()` stay no-return, and changed/skipped is now opt-in through the new `*Changed` APIs.

Check, tests, and build pass locally. GitHub checks pass.

G4tsby:
Hi @SameDesu123 , I recently opened #1557, which overlaps with the engine-recreation part of this PR — sorry I didn't spot yours before opening it. A quick comparison so the maintainer can weigh the two approaches:

- #1557 keys the engine cache by the script code inside `runScripted` itself, so no call-site `engineKey` is needed and every path is covered, including manual trigger names (which are used as `mode` and previously got one VM per name). Since the key doesn't include `mode`, a script also runs in a single VM instead of one per mode, and the cache is LRU-bounded with evicted engines being closed.
- Your identity-key approach has one advantage mine doesn't: Lua globals stay isolated between trigger instances that happen to share identical code (mine keeps the pre-existing sharing behavior of the mode-keyed cache).
- Your `setChatVar` redundant-write skip and the lightweight chat access APIs (`getChatData`/`getChatRole`/`getRecentChats`) don't overlap with #1557 at all and look genuinely useful — cheaper accessors also reduce `getFullChat` overuse. They'd pair well with either caching approach.

If #1557 lands first, I'd be happy to help rebase this PR so the non-overlapping parts go in cleanly — or the other way around, whichever direction the maintainer prefers.

sub-hub:
@SameDesu123 @G4tsby Let's review the current status based on both PRs.
First of all, thank you both for your contributions. I think both PRs are great because, aside from the overlapping parts, they include other features that help with optimization.

The engine caching is the overlapping part — both PRs touch the same `Map`, `getOrCreateEngineState`, and lifecycle in `scriptings.ts`. One of them will need to give way. #1459's `setChatVar` skip and lightweight chat APIs (`getChatData` / `getChatRole` / `getRecentChats`) are independent of whichever caching approach lands and look useful either way.

On the engine side: #1557 covers all call paths without call-site changes, has LRU eviction, and adds listener skip. #1459 keeps trigger-instance-level globals isolation.

The globals behavior is the one point worth discussing. #1459 deliberately isolates, while #1557 shares across modes when code is identical. G4tsby notes that the old mode-keyed cache could already share engines across characters when mode and code happened to match, so #1557 extends that implicit behavior rather than introducing something entirely new. Curious whether either approach has caused issues in practice for character authors, or whether this is more of a theoretical concern.

Would it make sense to land the non-overlapping parts of #1459 first, and give the globals-sharing discussion a bit of time on #1557? Or do you two see a better way to untangle this?

G4tsby:
@sub-hub Thanks for the clear summary — I agree with the direction you proposed, and I'd like to suggest a concrete way to untangle it:

1. **#1459 lands first, with the engine-caching part left out** — keeping the `setChatVar` redundant-write skip, `setChatVarChanged`/`setStateChanged`, and the lightweight chat APIs (`getChatData`/`getChatRole`/`getRecentChats`), while dropping the `engineKey` plumbing (the parameter, the key-builder functions, and the `getModuleTriggerRuntimes` refactor, which only exists to thread the key through call sites). Those parts are valuable regardless of the caching decision, and landing them without the caching code means nothing has to be removed again later, whichever way the engine decision goes.
2. **#1557 then handles the engine-cache side**, rebased on top of that — the conflicts would be textual only (the new `declareAPI` entries), not semantic.

On the globals question: I did initially share the concern about scenarios like a character author duplicating a card into several revisions for testing. But since #1557 keys the cache by the script code itself, any revision whose script was edited gets a fresh VM automatically — sharing can only occur between byte-identical copies. And that exact case already shared a VM under the old mode-keyed cache (same mode + same code reused the engine), yet I couldn't find any related reports in the community or the issue tracker, so in practice I believe this is a theoretical concern. It's also worth noting that `setChatVar`/`getState`-based state is re-bound to the current chat on every call, so it stays per-chat under either approach; only raw Lua globals are affected.

The cross-mode sharing in #1557 is intentional rather than incidental: it fixes the behavior where a global set in `onOutput` was invisible to `editDisplay`, because each mode previously ran in its own VM — which is not what script authors generally expect.

For the engine-side comparison, a few things I noticed while reading #1459's caching part — @SameDesu123, apologies in advance if I'm misreading anything:

1. `engineKey` is currently passed in `runLuaEditTrigger` and `runLuaButtonTrigger`, but the `triggerlua` path in `runTrigger` (`triggers.ts`) doesn't pass one, so it falls back to the mode-based default. If I'm reading it right, the `onStart`/`onOutput`/manual-trigger path would still rebuild VMs whenever two scripts alternate.
2. Since keys include `chaId × trigger index × mode` and there's no eviction or engine close, the map could grow without bound over a long session as the user moves between characters.
3. Keeping `mode` in the key means one script still gets a separate VM (and a separate top-level run) per mode it participates in.
4. The listener-skip optimization — the largest win in my benchmarks (~24× for the no-listener render path) — isn't tied to the key choice, but it's currently only implemented in #1557.
5. #1557 ships functional tests for the cache behavior (reuse, the old thrashing case as a regression guard, LRU eviction, listener skip); it would be good to have equivalent coverage for whichever approach lands.

None of these are hard to address in follow-ups, and to be clear, the non-overlapping parts of #1459 look genuinely useful to me — the cheaper accessors in particular should reduce `getFullChat` overuse nicely. That's exactly why I think splitting them out and landing them first is the fastest path for everyone.

SameDesu123:
@sub-hub @G4tsby Thanks for the careful comparison and for suggesting a clean split. I agree that keeping the non-overlapping pieces here and leaving the engine-cache work to #1557 is the clearest path, so I pushed d8bb994a to remove the `engineKey` parameter and key builders, restore `getModuleTriggers` without runtime metadata, and return all Lua call sites to the existing mode-based engine behavior. This PR now only keeps the redundant chat-state write skip, the compatibility-safe opt-in `setChatVarChanged` and `setStateChanged` APIs, and the lightweight `getChatData`, `getChatRole`, and `getRecentChats` accessors. I also updated the title and description to match the narrowed scope. `pnpm check`, the focused chat-variable Vitest suite, and `git diff --check` all pass. Thanks again for helping untangle the overlap.

G4tsby:
@SameDesu123 Thank you for handling the split so quickly — removing the engine-caching part cleanly and narrowing the scope makes this easy to review. The remaining pieces (the redundant-write skip, the opt-in `*Changed` APIs, and the lightweight chat accessors) all look good to me, and they no longer overlap with #1557 in any meaningful way. Thanks again for the collaboration!

cubicj:
I re-checked the PR after d8bb994ae. The engine-cache removal is clean, modules.ts is back to byte-identical with main, and the remaining pieces match what I reviewed before the split. svelte-check, the test suite, and a production build pass on this head.

One thing I noticed while retracing the new APIs, and I should have caught it when I approved 35a2b95fe: `setStateChanged` and `setChatVarChanged` return `false` on a no-op write, and a Lua handler that ends with `return setStateChanged(...)` feeds that `false` into the `res === false` check in `runScripted`, which sets `stopSending`. On a `start` trigger that silently aborts the generation, and only on turns where the value happened to be unchanged, so it shows up as an intermittent stop. I reproduced the round trip in a minimal wasmoon harness: the no-op return comes back as a real boolean `false` and trips the check, while the legacy `return setState(...)` path stays `nil` and passes. Existing scripts are unaffected since both APIs are new opt-ins, but that failure mode would be painful to debug for anyone who adopts them.

Returning `nil` for the unchanged case instead of `false` would avoid the sentinel while keeping `if setStateChanged(...) then` working. Do you think the collision is worth designing around here, or would you rather keep the clean boolean and document not to return it from a handler?

G4tsby:
Verified the same round trip on my side — the direct-mode path at `runScripted` passes the handler's return value straight through wasmoon, so the no-op `false` does hit the `res === false` stop check. Nice catch.

I'd vote for returning `nil` on the unchanged case. Two reasons beyond avoiding the collision: `if setStateChanged(...) then` keeps working identically since `nil` is falsy, and the `false` sentinel currently carries no extra information anyway — the access-key check failure also returns `false`, so "unchanged" and "no permission" are already indistinguishable. Dropping `false` loses nothing.

Documentation alone wouldn't help much here — an intermittent, unchanged-turns-only abort is exactly the kind of thing people burn hours on. Since `return false` from a handler is the intentional stop API, the new opt-in APIs are the right side to yield. A small regression test asserting that `return setStateChanged(...)` on an unchanged write doesn't set `stopSending` would lock it in.

SameDesu123:
@cubicj @G4tsby Thanks for catching and independently verifying this. I pushed 547f13ac so setChatVarChanged now returns nil when a write is unchanged or access is denied, while still returning true for an actual change. This keeps the normal conditional usage intact and prevents a no-op result from reaching the intentional false stop sentinel in runScripted. I also added a wasmoon-backed regression test that executes return setStateChanged(...) through onStart and confirms it does not set stopSending, along with coverage that an explicit return false still stops generation. The focused scripting and chat-variable tests pass (8/8), pnpm check reports 0 errors and 0 warnings, and git diff --check passes.

sub-hub (approved):
Smoke-tested in the app: redundant writes are correctly skipped (setStateChanged returns nil on unchanged, true on change), the legacy setState keeps its nil return, and a handler ending in `return setStateChanged(...)` no longer trips the stopSending sentinel, generation continues on unchanged turns. 
The lightweight chat accessors also work. 

Thanks for the PR.

PR #1461, opened 2026-05-24
fix: cbs reverse, setdefaultvar

tammy3211:
## PR Checklist

- Required Checks

## Summary

fix cbs reverse, setdefaultvar

## Related Issues

"None".

## Changes

fix cbs reverse, setdefaultvar

### DIFF
--- a/src/ts/cbs.ts
+++ b/src/ts/cbs.ts
@@ -2121,7 +2121,7 @@ export function registerCBS(arg:CBSRegisterArg) {
     registerFunction({
         name: 'reverse',
         callback: (str, matcherArg, args, vars) => {
-            return [...str].reverse().join('')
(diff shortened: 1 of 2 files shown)


sub-hub:
Every fix is a small fix, and the bugs are critical. Thank you for the PR.

Just one thing: there was a test for reverse cbs in the strings.test.ts file, but it had been skipped because it was failing.
Since this PR fixes that error, it seems like a good to stop skipping the reverse test in this PR too.

sub-hub (approved):
Image
Now the reverse test is working properly. Thank you!

cubicj (approved):
Verified the CBS behavior. `setdefaultvar` now treats the existing `getChatVar()` `"null"` sentinel as unset, and `reverse` now operates on the parsed argument instead of the whole CBS match.

PR #1462, opened 2026-05-25
fix: Backup feature Chunk load error

sub-hub:
## PR Checklist

- Required Checks

## Summary
image

Currently, An error occurs when trying to perform a backup after restart the web risu server.
The dynamic import streamsaver package does not work until reload the page because the address changes.

So I changed streamsaver import in LocalWriter to a static import for reliability (prevents backup failures if server restarts).
And added a global handler in main.ts for Vite’s vite:preloadError event. If a dynamic import fails (e.g., after server update), users are alerted that they needs to refresh the page.
There are other dynamic imports remaining, but I'm leaving them as is since they aren't important enough to convert to static.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -7,6 +7,11 @@ import { initHotkey } from "./ts/hotkey";
 import { preLoadCheck } from "./preload";
 import { mount } from "svelte";
 
+window.addEventListener('vite:preloadError', (event) => {
+    console.error("Chunk load error detected:", event);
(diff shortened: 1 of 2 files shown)


cubicj (approved):
Verified the backup export path and preload-error handling. Making `streamsaver` a static import removes the late chunk load from `LocalWriter`, and the global preload handler gives users a clear refresh path for remaining stale chunks.

PR #1464, opened 2026-05-29
feat: add Claude Opus 4.8 model

tasoo-oos:
## PR Checklist

- Required Checks

## Summary

Add Claude Opus 4.8 to the Anthropic model list and add the current AWS Bedrock Claude Opus 4.7 model ID.

## Related Issues

None.

## Changes

- Add `claude-opus-4-8` as a recommended Anthropic model.
- Configure Opus 4.8 with adaptive thinking and `xhigh` effort support.
- Omit unsupported sampling parameters for Opus 4.8.
- Expose the `claudeXHighEffort` custom flag in model settings.

## Impact

Users can select Claude Opus 4.8 through the RisuAI Official Anthropic API implementation. Existing models are unchanged.

## Additional notes

Manually tested with streaming on / off and various thinking settings.

### DIFF
--- a/src/lib/Setting/Pages/Advanced/CustomModelsSettings.svelte
+++ b/src/lib/Setting/Pages/Advanced/CustomModelsSettings.svelte
@@ -163,6 +163,7 @@
                 {@render CustomFlagButton(index,'noCivilIntegrity', 20)}
                 {@render CustomFlagButton(index,'claudeThinking', 21)}
                 {@render CustomFlagButton(index,'claudeAdaptiveThinking', 22)}
+                {@render CustomFlagButton(index,'claudeXHighEffort', 23)}
                 {@render CustomFlagButton(index,'deepSeekThinkingToggle', 24)}
             </Accordion>
                 </div>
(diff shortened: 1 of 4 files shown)


cubicj (commented):
One Bedrock-specific issue: this adds `claudeAdaptiveThinking` to AWS Claude 4.7 Opus, so the existing adaptive thinking UI can expose `effort: "max"` for it. AWS documents `max` as Claude Opus 4.6-only, while other models should use `low`, `medium`, or `high`.

https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html

Could you narrow the exposed effort options for Bedrock Opus 4.7?

tasoo-oos:
I removed AWS support for Claude Opus 4.7 from this PR.

At first, I thought it won't be so troublesome since there's already an API implementation for AWS and new model's behavior seems to be the same as the last one (max adaptive thinking behavior change part was not mentioned in [model docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html)), so It would be okay to just slip it into the PR.

However, if model behavior has changed and I am unable to test it, I think it's better not to add it.

image

Besides, who would possibly try to contact AWS sales to use Opus 4.7 for RolePlay when they could just go to Anthropic or Openrouter?

But at the same time, I should have searched for the docs more rigorously, or be more conservative with what I cannot test by myself. Sorry about that.

cubicj (approved):
Verified the follow-up. The AWS Bedrock Claude 4.7 entry was removed, and the remaining Anthropic Claude 4.8 path matches the documented model ID, adaptive thinking support, and omitted sampling parameters.

Issue #1466, opened 2026-06-01
(Docker) Unable to bypass the legal screen (VITE_RISU_LEGAL_CONFIGURED=TRUE does not work)


GamingDaveUk:
<img width="761" height="560" alt="Image" src="https://github.com/user-attachments/assets/c10e3640-b3f6-4f2d-8870-3a181d97103c" />
So self hosting for private use, ideally to replace silly tavern. I have a lot of utility bots, all needing their own system prompts so Risuai looks ideal.... However... I cant actually get into my private hosted instance! 

My docker compose is:
```
name: risuai

services:
  risuai:
    container_name: risuai
    image: ghcr.io/kwaroran/risuai:latest
    restart: always
    ports:
      - 6001:6001
    environment:
         - VITE_RISU_LEGAL_CONFIGURED=${VITE_RISU_LEGAL_CONFIGURED}
    volumes:
      - /home/mrgreaper/docker/risuai/saves:/app/save
      - /home/mrgreaper/docker/risuai/.env:/app/.env
```

I am using dockhand. ${VITE_RISU_LEGAL_CONFIGURED} is configured to put TRUE 
I also tried - VITE_RISU_LEGAL_CONFIGURED=TRUE
The .env file at /home/mrgreaper/docker/risuai/.env has VITE_RISU_LEGAL_CONFIGURED=TRUE in it as does /home/mrgreaper/docker/risuai/saves/.env (because just incase) 

I have tried firefox and brave
I have cleared cache
I have redeployed the docker multiple times. 
I can not find any info on this in the wiki 

The pop up asking me to accept the terms of service does not have an accept button, the linked place holder terms, do not have accept buttons, i can not edit the source code as its a docker image, plus i am using it private so thats not needed? 

I am completly at a loss as to how to fix this? 

EDIT: 
can confirm its set in the docker cmd:

```
Connecting to risuai...
Shell: /bin/bash, User: root

root@16343d431f36:/app# printenv | grep VITE
VITE_RISU_LEGAL_CONFIGURED=TRUE
```

GamingDaveUk:
Decided to attempt to fix my issue with gemini.... i must be doing something wrong as only one other person has the issue, on discord.... 

Gemini says the issue is:
```
[Bug] Docker image ignores VITE_RISU_LEGAL_CONFIGURED runtime environment variable
### Describe the bug
[...]

PR #1472, opened 2026-06-08
fix: Check VITE_RISU_LEGAL_CONFIGURED in process runtime

sub-hub:
## PR Checklist

- Required Checks

### Error Reproduction Method

Run `pnpm build` without `VITE_RISU_LEGAL_CONFIGURED` set. Then, set `VITE_RISU_LEGAL_CONFIGURED` to TRUE, run the server with `pnpm runserver`, and when you enter the server, a popup appears.

### Solution

When `server.cjs` is executed, inject `process.env.VITE_RISU_LEGAL_CONFIGURED` into the HTML so that the environment variable value can be retrieved even after the build is complete.

Also, previously, the popup would not appear as long as the value existed, even if `VITE_RISU_LEGAL_CONFIGURED` was not TRUE; however, the newly added code strictly checks for TRUE.

## Related Issues

#1466 

# Result
image
image

### DIFF
--- a/server/node/server.cjs
+++ b/server/node/server.cjs
@@ -615,7 +615,8 @@ app.get('/', async (req, res, next) => {
         const mainIndex = await fs.readFile(path.join(process.cwd(), 'dist', 'index.html'))
         const root = htmlparser.parse(mainIndex)
         const head = root.querySelector('head')
-        head.innerHTML = `<script>globalThis.__NODE__ = true</script>` + head.innerHTML
+        const legalConfigured = process.env.VITE_RISU_LEGAL_CONFIGURED?.trim().toUpperCase() === 'TRUE';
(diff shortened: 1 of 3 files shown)


cubicj (approved):
Verified the Node server runtime path. A build without `VITE_RISU_LEGAL_CONFIGURED` now picks up `VITE_RISU_LEGAL_CONFIGURED=TRUE` through `server.cjs`, and the reporter confirmed the Docker retest gets past the legal screen.

sub-hub:
@kwaroran Could you please take a look at this feature? We will maintain the requirement that users must manually set VITE_RISU_LEGAL_CONFIGURED to TRUE for consent to be granted, while fixing only the bug that prevents consent to VITE_RISU_LEGAL_CONFIGURED in Docker.

jurassic410:
just writing to know if this is still being looked at or has been missed :-)

rainbuilt:
@kwaroran @sub-hub I guess every Docker user can just clone the repository, reimplement exactly what this PR already fixes, build their own Docker image, and run that instead. It's only a few minutes of extra work for every affected user.

PR #1469, opened 2026-06-07
feat: renew Gemini API support

tasoo-oos:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions?
    - [ ] Have you tested your changes?
    - [ ] Have you checked that it will not break any existing features?
- [x] If your PR uses models[^1], check the following:
    - [ ] Have you checked if it works normally in all models?
    - [ ] Have you checked if it works normally in all web, local, and node-hosted versions? If it does not, have you blocked it in those versions?
- [ ] If your PR is highly AI generated[^2], check the following:
    - [ ] Have you understood what the code does?
    - [ ] Have you cleaned up any unnecessary or redundant code?
    - [ ] Is it not a huge change?
       - We currently do not accept highly AI generated PRs that are large changes.

[^1]: Modifies the behavior of prompting, requesting, or handling responses from AI models.
[^2]: Over 80% of the code is AI generated.

## Summary

Renew Gemini API support for newer models, including Gemini 3/3.5 thinking levels, updated tool-call handling, and thought-signature preservation.

## Related Issues

None

## Changes

- Added Gemini thinking-level support with model-specific allowed levels and defaults.
- Added new Gemini model metadata, including Gemini Flash 3.5.
- Split Gemini thinking behavior into budget-based and level-based modes.
- Updated Gemini request bodies to send the correct `thinkingConfig` shape.
- Updated Gemini tool declarations and function responses for newer protocol compatibility.
- Preserved full model-returned parts during Gemini tool-call loops.
- Improved Gemini streaming SSE parsing and stream metadata tracking.
- Added `geminiThinkingLevel` to settings, database state, and bot presets.
- Preserved Gemini `thoughtSignature` values in inlayed signatures.

## Impact

This should improve compatibility with newer Gemini models and Gemini tool use. Existing Gemini 2.5 thinking models continue to use token budgets, while Gemini 3-style models use thinking levels. Gemini tool-use conversations may behave differently because full model parts are now preserved instead of reconstructed.

## Additional Notes

Testing has not been run yet.

PR #1475, opened 2026-06-09
fix: use thinkingLevel for Gemini 3.x instead of thinkingBudget

LLMLover452:
## PR Checklist

- Required Checks

## Summary

Two fixes:
1. Fix the model match. The old regex `/^gemini-3-/` missed `gemini-3.1` and `gemini-3.5`.
2. Set Gemini 3.x models to use `reasoning_effort` (Minimal / Low / Medium / High), and send it as `thinkingLevel`.

Gemini 3.x models use `thinkingLevel` (a text value: `minimal`, `low`, `medium`, or `high`), not a token number (`thinkingBudget`).

The old code used the regex `/^gemini-3-/` to find Gemini 3 models. This only matches IDs that start with `gemini-3-` (like `gemini-3-pro-preview`). It does not match `gemini-3.1-...` or `gemini-3.5-...`, because of the dot after `3`. So those models fell through and sent the wrong field (`thinkingBudget` instead of `thinkingLevel`).

This PR fixes it. Gemini 2.5 still uses `thinkingBudget` (no change).

## Related Issues

#1469 (draft) also works on Gemini thinking level, as part of a larger Gemini API update. This PR is a smaller, focused take that only fixes the thinking-level handling and reuses the existing `reasoning_effort` control. I'm happy to defer to #1469 or close this if that PR already covers it.

## Changes

`src/ts/model/types.ts`
- Add a new `geminiThinkingNoMinimal` model flag.

`src/ts/model/providers/google.ts`
- For `gemini-3.1-pro-preview`, `gemini-3-pro-preview`, and `gemini-3-flash-preview`: change the parameter from `thinking_tokens` to `reasoning_effort`.
- Now these models show the Minimal / Low / Medium / High buttons (same UI as GPT-5).
- Add the `geminiThinkingNoMinimal` flag to `gemini-3.1-pro-preview`, because it does not support the `minimal` level.

`src/ts/process/request/google.ts`
- Send `reasoning_effort` as `thinkingConfig.thinkingLevel` (lowercase).
- Remove the old `/^gemini-3-/` code. It did not match `gemini-3.1`, it used uppercase values, and it guessed the level from the token slider.
- For models with the `geminiThinkingNoMinimal` flag, map `minimal` to `low` (this matches what Google's own API does for Gemini 3.1 Pro).
- Gemini 2.5 still puts its token number inside `thinkingConfig`.

## Impact

- Gemini 3.x now sends the correct `thinkingLevel`. Before, `gemini-3.1` sent the wrong field.
- `gemini-3.1-pro-preview` maps `minimal` to `low`, because it does not support `minimal`.
- Gemini 2.5 does not change.
- The Gemini 3.x setting UI is now the same as GPT-5 (level buttons, not a token slider).
- No new setting and no new database field.

image

## Additional Notes

- Small change (3 files). No new settings or database fields, only a model flag.
- The code is AI-assisted, but I read it and understand it.
- I checked the UI: the level buttons show for Gemini 3.x, and Gemini 2.5 still shows the token slider.
- Tested live in my self-hosted Docker client. Working: `gemini-3.1-pro-preview`, `gemini-3-flash-preview`, and `gemini-2.5-pro`.
- Note: `gemini-3-pro-preview` now returns 404 (Google deprecated it server-side).

### DIFF
--- a/src/ts/model/providers/google.ts
+++ b/src/ts/model/providers/google.ts
@@ -9,7 +9,7 @@ export const GoogleModels: LLMModel[] = [
         provider: LLMProvider.GoogleCloud,
         format: LLMFormat.GoogleCloud,
         flags: [LLMFlags.hasImageInput, LLMFlags.poolSupported, LLMFlags.hasAudioInput, LLMFlags.hasVideoInput, LLMFlags.hasStreaming, LLMFlags.requiresAlternateRole, LLMFlags.geminiThinking, LLMFlags.hasFirstSystemPrompt],
-        parameters: ['thinking_tokens', 'temperature', 'top_k', 'top_p', 'presence_penalty', 'frequency_penalty'],
+        parameters: ['reasoning_effort', 'temperature', 'top_k', 'top_p', 'presence_penalty', 'frequency_penalty'],
         tokenizer: LLMTokenizer.GoogleCloud,
         recommended: true
     },
@@ -20,7 +20,7 @@ export const GoogleModels: LLMModel[] = [
         provider: LLMProvider.GoogleCloud,
         format: LLMFormat.GoogleCloud,
         flags: [LLMFlags.hasImageInput, LLMFlags.poolSupported, LLMFlags.hasAudioInput, LLMFlags.hasVideoInput, LLMFlags.hasStreaming, LLMFlags.requiresAlternateRole, LLMFlags.geminiThinking, LLMFlags.hasFirstSystemPrompt],
-        parameters: ['thinking_tokens', 'temperature', 'top_k', 'top_p', 'presence_penalty', 'frequency_penalty'],
+        parameters: ['reasoning_effort', 'temperature', 'top_k', 'top_p', 'presence_penalty', 'frequency_penalty'],
         tokenizer: LLMTokenizer.GoogleCloud,
         recommended: true
     },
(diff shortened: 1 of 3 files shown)


tasoo-oos:
Hello! Author of #1469 here.

I like this direction. Compared with my previous PR, this one is much smaller and should be much easier for maintainers to review and merge.

I noticed one small model-specific edge case: `gemini-3.1-pro-preview` does not support `thinkingLevel: "minimal"` according to the Gemini thinking docs:

https://ai.google.dev/gemini-api/docs/thinking

Since this PR exposes the same Minimal / Low / Medium / High `reasoning_effort` UI for Gemini 3.x models, selecting Minimal for Gemini 3.1 Pro can make the API reject the request.

Example error:

image

A possible fix would be to handle allowed thinking levels per model (refer to OpenAI / Anthropic / Deepseek API's reasoning effort implementation if you are curious), or just map/guard `minimal` to `low` for Gemini 3.1 Pro.

Apart from that edge case, this PR looks good. Thanks for splitting this into a focused change!

LLMLover452:
Thank you!

At first I thought that if a request fails because the user picked an unsupported option, it should just be left as it is. But looking at it from a normal end user's view, I agree it makes more sense to map it to a valid level, like you said.

Since future Gemini Pro models might also not support `minimal`, I implemented it with a model flag `geminiThinkingNoMinimal` instead of a hardcoded model name.

Tested on Gemini 3.1 Pro at the minimal level:

image

The older `gemini-3-pro-preview` only supported `low` and `high`, but it is no longer available. So I did not add the flag to it.

tasoo-oos:
Good job! Thanks for the contribution!

cubicj (approved):
Looks good. I checked the Gemini 3.x request path and confirmed the switch from `thinking_tokens` to `reasoning_effort` reaches `thinkingConfig.thinkingLevel`, while Gemini 2.5 still keeps the `thinkingBudget` path.

I also tested `gemini-3-flash-preview` with `minimal` and `low`; both requests worked. `gemini-3-pro-preview` is already unavailable from the API side, so the remaining live path looks fine to me.

sub-hub (approved):

PR #1477, opened 2026-06-10
feat: Recommend Response API alongside Chat Completion

tasoo-oos:
## PR Checklist

- Required Checks

## Summary

Recommend OpenAI Responses API too.

## Related Issues

None.

## Changes

- Make generated OpenAI Responses API models inherit the source model recommendation flag.

## Impact

Recommended OpenAI model's Responses API variant also appear in the default model list.

## Demo

image

### DIFF
--- a/src/ts/model/modellist.ts
+++ b/src/ts/model/modellist.ts
@@ -591,8 +591,7 @@ for(let i=0; i<LLMModels.length; i++){
             id: `${LLMModels[i].id}-response-api`,
             name: `${LLMModels[i].name} (Response API)`,
             fullName: `${LLMModels[i].fullName ?? LLMModels[i].name} (Response API)`,
-            recommended: false
-
+            recommended: !!LLMModels[i].recommended
         })
     }
     if(LLMModels[i].provider === LLMProvider.GoogleCloud){
--- a/src/ts/model/modellist.ts
+++ b/src/ts/model/modellist.ts
@@ -814,7 +814,7 @@ export function getModelList<T extends boolean>(arg:{
 } = {}): T extends true ? GetModelListGroup[] : LLMModel[]{
     let models = LLMModels
     if(arg.recommendedOnly){
-         models = models.filter(model => model.recommended)
+         models = models.filter(model => model.recommended && (DBState.db.useExperimental === true || model.format !== LLMFormat.OpenAIResponseAPI))
     }
     const pluginGroup: GetModelListGroup = {
         providerName: 'Plugins',


cubicj (commented):
Personally, I think choosing between Chat Completions and Responses API is an advanced detail, so I am not sure if it should be exposed in the normal model picker by default. Responses API variants are already available through Show unrecommended.

The code change itself looks fine to me.

tasoo-oos:
I am thinking about this patch as a beta release.

Personally, I think RisuAI should migrate OpenAI-API call method to Response API someday. 

OpenAI [said](https://developers.openai.com/api/docs/guides/migrate-to-responses) "While Chat Completions remains supported, Responses is recommended for all new projects.". Also, Some features may not be shipped since Response API is "Future-proof" but Chat Completions is not. One good example is Summarized thinking. Response API ships it, but Chat Completion omits thinking completely.

In contrast, I think most end users won't know Response API is even supported, let alone try to use it after scrolling piles of unrecommended models.

With this patch, users now have easier access to the Response API path and share their opinions or report bugs with it.

Though this is my personal opinion and I think this patch and direction may need more debate.

cubicj:
Personally, I agree with the long-term direction toward Responses API, but I think we should also hear the maintainer’s view on this.

My concern is mostly about exposing the API choice as extra model entries in the normal picker. In the default state, before users enable advanced settings, having too many choices can raise the entry barrier. I think complex or advanced options should stay hidden as much as possible unless the user explicitly opts into them.

If Responses is the preferred path, it may be cleaner to eventually make it the default OpenAI path and keep Chat Completions as a legacy/advanced fallback, rather than showing both API variants as regular model choices. For beta feedback, an advanced/beta option might make the rollout intent clearer without adding API-choice noise to the default picker.

tasoo-oos:
Agreed with your opinion. UI/UX would be pretty disruptive with dozens of models.

The idea of "`Advanced Settings -> Enable Experimental Features` makes Response API recommended" sounds better then the current patch. I'll implement that.

tasoo-oos:
Done! Now Response API variant is recommended only when Experimental features toggle is active.

cubicj (approved):
Looks good. I checked the model picker path in dev after the follow-up commit, and the Responses API variants now stay out of the default recommended list unless Experimental Features is enabled.

`Show unrecommended` still keeps the variants reachable, so this resolves my UI-scope concern without changing the actual Responses request path. Local check, tests, and build pass, and CodeQL checks are green.

PR #1478, opened 2026-06-10
feat: support OpenAI Flex for Responses API

tasoo-oos:
## PR Checklist

- Required Checks

## Summary

Add OpenAI Flex Processing support for Responses API requests.

## Related Issues

None.

## Changes

- Share the OpenAI Flex eligibility check between Chat Completions and Responses API.
- Add `service_tier: "flex"` to eligible official Responses API requests.
- Update Flex Processing help text to mention both Chat Completion and Response API.
- Add a Responses API test for Flex Processing to mention.

## Impact

Responses API users (via Risu-official and Custom API with `api.openai.com`) can use OpenAI Flex Processing.
Non-OpenAI custom endpoints such as plugins are not affected.

## Additional Notes

Tested with `pnpm vitest run src/ts/process/request/openAI/requests.responses.test.ts` and `pnpm check`.
Manually tested, and works for Streaming / Non-Streaming


cubicj (commented):
The Flex support itself looks good, but I found one streaming failure-handling regression.

This PR changes Responses API `response.failed` / `response.error` events from a normal stream chunk into `controller.error()`, so `reader.read()` now rejects. Some streaming callers, for example the subtitle playground paths, read the stream without catching reader errors, so a failed Responses stream can bypass their existing `alertError` handling and leave the UI in an in-progress state.

Could you make sure failed Responses stream events are handled safely by all streaming callers before switching this path to `controller.error()`?

tasoo-oos:
Done! I tested Subtitle path with gemini 3 Pro and Whisper + GPT method.
I coudln't use Native OpenAI LLM to the Subtitle test since no LLM receives Audio / Video.
Also, due to some sound-related error, I had to supress it same way as https://github.com/kwaroran/Risuai/pull/1264

cubicj (commented):
Thanks for the follow-up. The subtitle and trigger reader-side handling looks better now, and the rest of the changed paths look fine to me.

One thing I still noticed is that when `getResponsesTranStream()` calls `controller.error()` for `response.failed` / `response.error`, the consumer's `reader.read()` rejects, but the upstream `response.body.pipeTo(transtream.writable)` promise also rejects separately. I reproduced a minimal Web Streams case where the reader catch runs, but the `pipeTo()` rejection is still unhandled.

Could you take a look at that remaining stream failure path?

tasoo-oos:
Done! I made it be supressed with `body.pipeTo(transtream.writable).catch(() => {})`.
Would this be the desired behavior?

cubicj (approved):
Looks good. I rechecked the latest follow-up against the remaining stream failure concern. Failed Responses events still reach callers through `reader.read()`, and the `pipeTo()` rejection is now handled at both the initial and tool-follow-up stream boundaries.

The `service_tier` override path also looks right. Additional Params can still override or remove the Flex default. `pnpm check`, `pnpm test`, and `pnpm build` pass locally, and CodeQL is green.

PR #1479, opened 2026-06-11
fix: keep checkbox switch focus within control

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Keeps the redesigned checkbox switch focus target aligned with the visible switch control instead of using an offscreen `sr-only` input.

## Related Issues

None

### `src/lib/UI/GUI/CheckInput.svelte`
- Renders the actual checkbox input as an invisible absolute overlay inside the visible switch track.
- Keeps the input focus box within the switch bounds to avoid browser scroll correction when toggles are inside nested scroll containers.
- Preserves the switch visuals, checked-state icon, accessible switch role, and existing label behavior.

## Impact

This should prevent custom prompt toggle switches in the bot sidebar from causing layout or scroll jumps when lower items are toggled in a long toggle list. Other `CheckInput` users keep the same switch-style UI.

## Additional Notes

Validated with `pnpm check`. The affected custom prompt toggle scenario was manually verified after the focus target adjustment.

### DIFF
--- a/src/lib/UI/GUI/CheckInput.svelte
+++ b/src/lib/UI/GUI/CheckInput.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
+    import { CheckIcon } from '@lucide/svelte';
 
     interface Props {
         check?: boolean;
-        onChange?: (check:boolean) => any,
+        onChange?: (check:boolean) => any;
         margin?: boolean;
         name?: string;
         hiddenName?: boolean;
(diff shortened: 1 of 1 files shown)


cubicj (approved):
I focused on whether this keeps the checkbox focus target inside the visible switch bounds without introducing a new event or layout issue. The complex custom toggle list no longer reproduces the scroll-jump problem in local dev, and the overlay input structure looks scoped to the switch track.

`pnpm check` passes locally.

SameDesu123:
After reconsidering it, I feel that this is somewhat at odds with the simple UI that Risu aims for.

For that reason, I’m closing this PR.

PR #1481, opened 2026-06-11
Add visual color palette selector

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Replace the color scheme dropdown in Display settings with visual palette cards and improve how custom color schemes are stored and edited.

The PR also adds several built-in themes and separates the user's custom palette from the currently active preset, allowing it to remain intact while switching between themes.

## Related Issues

None

## Changes

- Replaces the color scheme dropdown with a responsive card-based palette selector.
- Shows a four-color preview for each built-in theme and the user's Custom theme.
- Adds additional built-in dark and light color schemes.
- Separates customColorScheme from the currently active colorScheme.
- Preserves existing custom color schemes when loading older databases.
- Keeps Plugin API V3 color scheme updates synchronized with the stored custom scheme.
- Uses lightweight native color inputs in the Custom scheme editor instead of mounting full color picker components for every field.
- Prevents ColorInput from firing its input callback during initial component setup.
- Keeps the existing color scheme import/export and theme application behavior.

## Impact

Color schemes can now be compared visually instead of through a text-only dropdown.

Custom palettes are stored independently from built-in presets, so selecting another theme no longer replaces the palette shown under Custom. Existing databases are migrated to the new representation when loaded.

The Plugin API continues to support applying custom color schemes, with the stored Custom palette kept in sync.

The Custom editor also avoids mounting multiple heavy color picker widgets, reducing unnecessary work when opening Display settings.

No model request or prompting behavior is changed.

## Additional Notes

Validation performed:

- `pnpm check`
- `git diff --check`
- Browser rendering checks for the palette grid, selection state, responsive layout, and Custom preview stability when switching presets
- Browser interaction checks for Custom mode display-tab entry, display sub-tab navigation, Light-to-Custom selection, and custom color edit propagation

## Preview

| Fancy | 
|-------|
| ahuwtsblarewanrnlbes |

### DIFF
--- a/src/lib/Setting/Pages/Display/ColorSchemeSelect.svelte
+++ b/src/lib/Setting/Pages/Display/ColorSchemeSelect.svelte
@@ -1,19 +1,88 @@
 <script lang="ts">
     import { language } from 'src/lang';
     import { DBState } from 'src/ts/stores.svelte';
-    import { changeColorScheme, colorSchemeList } from 'src/ts/gui/colorscheme';
-    import SelectInput from 'src/lib/UI/GUI/SelectInput.svelte';
-    import OptionInput from 'src/lib/UI/GUI/OptionInput.svelte';
+    import {
+        changeColorScheme,
+        colorSchemeList,
+        colorSchemePresets,
+        type ColorScheme,
+    } from 'src/ts/gui/colorscheme';
 
-    const onSchemeInputChange = (e: Event) => {
-        changeColorScheme((e.target as HTMLInputElement).value);
+    const formatSchemeName = (name: string) => {
+        if (name === 'custom') {
+            return 'Custom';
+        }
+
+        return name
+            .split('-')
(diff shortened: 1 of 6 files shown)


cubicj (approved):
I focused on whether the new `customColorScheme` split could lose existing custom themes or leave API-applied custom schemes out of sync. Existing custom themes are carried over in `setDatabase`, `setColorScheme()` now updates `customColorScheme` as well, and the remaining `ColorInput` usage still receives picker input updates.

`pnpm check`, `pnpm test`, and `pnpm build` pass locally. GitHub checks are green.

sub-hub (approved):
I really like the new UI because it is very pretty.
It also seems to have no logical issues. In fact, I think it is good that the color I selected in the theme custom used to disappear the moment I chose a different preset, whereas now the value is saved separately in customColorScheme.
Thanks for the PR.

Issue #1425, opened 2026-04-29
Failed to fetch error after using clean unused cold storage


nainclerk:
## Environment
- Risu version: 2026.4.181
- Platform: PC (Edge) / Mobile (Chrome)

## What happened
I used Risu for several days without enabling "Save data to account".

Later, I tried to sync/save data, but I got this error:

Risu version: 2026.4.181

TypeError: Failed to fetch
    at ../../src/ts/sionyw.ts:181:22
    at ../../src/ts/sionyw.ts:61:15
    at ../../src/ts/storage/accountStorage.ts:39:23
    at ../../src/ts/storage/autoStorage.ts:82:50
    at ../../src/ts/bootstrap.ts:154:20

I also remember clicking the "unused cold storage" button a few times earlier.

## Issue
Risu only works properly on the web (PC).
On mobile, data cannot be loaded at all — neither through Google backup nor "Save data to account", so I cannot proceed.

## Question
How can I load the latest web (PC) data on the mobile version?

cubicj:
Resolved in PR #1484 and released in v2026.6.215. The account sync migration path from this stack trace was reworked — cold storage items are now validated and migrated explicitly, and a fetch failure during migration now shows a handled error instead of crashing.

Closing — please reopen if this still occurs on the current version.

cubicj:
Correction: the v2026.6.215 builds were made from a hotfix branch and do not actually include PR #1484, so the fix has not shipped yet. Reopening until it lands in the next release.

PR #1484, opened 2026-06-13
fix: preserve cold storage during sync and backup

### LINKED ISSUE #1482: TypeError: Cannot read properties of null (reading 'character') on bootstrap after Google account sync
Environment

- Risu version: 2026.6.111
- Platform: Mobile (Android 16, Chrome) / PC (Windows, Edge)

What happened

I saved data on mobile and later tried to load it on PC via Google account sync. On both platforms, the following error appears immediately on load:

```
Error
Cannot read properties of null (reading 'character')

Risu version: 2026.6.111
OS: Windows 19.0.0
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36 Edg/149.0.0.0
Risu environment: web

TypeError: Cannot read properties of null (reading 'character')
    at ../../src/ts/globalApi.svelte.ts:915:28
    at ../../src/ts/bootstrap.ts:523:32
```

The same error occurs regardless of which backup date I restore — all available Google account backups (including several days prior) produce the identical error.

Issue

Risu fails to load on both Android and PC when using Google account data. Since every saved backup triggers the same crash at bootstrap, it appears the data became corrupted at or before the point it was first synced to the Google account — meaning no usable backup remains via that method.
I suspect the root cause may be the "Skip Saving Assets on Web Sync" option being enabled by default in a recent update, though I am not certain.

Question

Is there a way to manually inspect or repair the synced Google account data (e.g. by editing the raw JSON) to remove the null `character` entry causing the crash? If so, where is the corrupted field likely located in the data structure?

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Fixes a cold storage sync/backup gap that can leave the database with cold storage pointers but no corresponding cold storage payloads.

This prevents bootstrap and character-switch paths from throwing when an existing cold storage item is missing, and ensures future account sync and backup flows carry the cold storage payloads alongside the database pointers.

## Related Issues

Fixes #1482 and #1425

## Changes

- Guard cold storage reads before dereferencing `character`, so missing payloads do not crash bootstrap or character selection.
- Add shared cold storage helpers for collecting referenced cold storage keys and recognizing backup payload names.
- During account sync migration, copy referenced cold storage items into the account cold storage endpoint before writing the migrated database.
- Include cold storage payloads in Google Drive backup uploads and restore them before applying a loaded Drive database.
- Include cold storage payloads in local and partial local backups even when the current storage backend is account sync, and restore them through the normal cold storage writer.
- Add focused tests for cold storage key collection and backup payload recognition.

## Impact

Users with already-missing cold storage payloads should no longer hit a `Cannot read properties of null (reading 'character')` crash during bootstrap. Those missing payloads are still unrecoverable unless a valid backup exists, but the app can continue loading and future sync/backup operations should no longer create pointer-only cold storage states.

Validation performed:

- `pnpm vitest run src/ts/process/coldstorageData.test.ts`
- `pnpm check`
- `git diff --check`

## Additional Notes

Existing corrupted account data may still require restoring from a backup that contains the missing cold storage payloads.

### DIFF
--- a/src/lang/cn.ts
+++ b/src/lang/cn.ts
@@ -31,6 +31,7 @@ export const languageChinese = {
         "vertexAuthError": "Vertex AI 身份验证详情缺失。",
         "coldStorageWriteFailed": "冷存储写入失败。您的聊天数据已被保留。",
         "coldStorageVerifyFailed": "冷存储验证失败。您的聊天数据已被保留。",
+        "coldStorageRestoreFailed": "无法加载冷存储数据。受影响角色的数据可能已永久丢失。",
         "coldStorageIncompleteBackupConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
             `${characterNames || "未知角色"}的冷存储数据缺失或无效。${unresolvedCount > 0 ? `有 ${unresolvedCount} 个项目无法关联到角色。` : ""}\n\n如果继续，此备份将缺少 ${unavailableCount} 个冷存储项目，受影响的角色或聊天数据可能无法从该备份恢复。\n\n仍要创建不完整的备份吗？`,
         "coldStorageIncompleteRestoreConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
--- a/src/lang/cn.ts
+++ b/src/lang/cn.ts
@@ -30,7 +30,11 @@ export const languageChinese = {
         "requestLogRemovedDesc": "当客户端刷新或加载时，该请求记录会被删除。",
         "vertexAuthError": "Vertex AI 身份验证详情缺失。",
         "coldStorageWriteFailed": "冷存储写入失败。您的聊天数据已被保留。",
-        "coldStorageVerifyFailed": "冷存储验证失败。您的聊天数据已被保留。"
+        "coldStorageVerifyFailed": "冷存储验证失败。您的聊天数据已被保留。",
+        "coldStorageIncompleteBackupConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
+            `${characterNames || "未知角色"}的冷存储数据缺失或无效。${unresolvedCount > 0 ? `有 ${unresolvedCount} 个项目无法关联到角色。` : ""}\n\n如果继续，此备份将缺少 ${unavailableCount} 个冷存储项目，受影响的角色或聊天数据可能无法从该备份恢复。\n\n仍要创建不完整的备份吗？`,
+        "coldStorageIncompleteRestoreConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
+            `无法恢复 ${characterNames || "未知角色"}的冷存储数据。${unresolvedCount > 0 ? `有 ${unresolvedCount} 个项目无法关联到角色。` : ""}\n\n如果继续，${unavailableCount} 个冷存储项目将不可用，受影响的角色或聊天数据可能会永久丢失。\n\n仍要继续不完整的恢复吗？`
     },
     "showHelp": "显示帮助",
     "help": {
--- a/src/lang/de.ts
+++ b/src/lang/de.ts
@@ -31,6 +31,7 @@ export const languageGerman = {
         "vertexAuthError": "Vertex AI-Authentifizierungsdetails fehlen.",
         "coldStorageWriteFailed": "Kaltlager-Schreibvorgang fehlgeschlagen. Ihre Chat-Daten wurden beibehalten.",
         "coldStorageVerifyFailed": "Kaltlager-Überprüfung fehlgeschlagen. Ihre Chat-Daten wurden beibehalten.",
+        "coldStorageRestoreFailed": "Kaltlagerdaten konnten nicht geladen werden. Die Daten des betroffenen Charakters können dauerhaft verloren sein.",
         "coldStorageIncompleteBackupConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
             `Kaltlagerdaten für ${characterNames || "unbekannte Charaktere"} fehlen oder sind ungültig.${unresolvedCount > 0 ? ` ${unresolvedCount} Element(e) konnten keinem Charakter zugeordnet werden.` : ""}\n\nWenn Sie fortfahren, fehlen in dieser Sicherung ${unavailableCount} Kaltlagerelement(e). Die betroffenen Charakter- oder Chatdaten können daraus möglicherweise nicht wiederhergestellt werden.\n\nTrotzdem eine unvollständige Sicherung erstellen?`,
         "coldStorageIncompleteRestoreConfirm": (characterNames: string, unavailableCount: number, unresolvedCount: number) =>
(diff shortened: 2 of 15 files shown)


cubicj (commented):
The cold storage preservation direction looks right, but I think this also makes the existing pointer-only data case more important to handle carefully.

One path I am worried about is when the database already references cold storage keys but the payload only exists locally, or no longer exists at all. `backupDrive()` currently only logs and skips missing cold storage payloads before uploading the database backup, `restoreColdStorageFromDrive()` also continues when a referenced payload is missing from Drive, and local/Drive restore calls `setColdStorageItem()` without checking whether the write actually succeeded. In those cases, the app can still end up with a restored or newly uploaded database that contains cold storage pointers without the matching payloads.

Could you take another look at whether these partial backup/restore cases should be surfaced to the user before the database is applied or uploaded?

SameDesu123:
> The cold storage preservation direction looks right, but I think this also makes the existing pointer-only data case more important to handle carefully.
> 
> One path I am worried about is when the database already references cold storage keys but the payload only exists locally, or no longer exists at all. `backupDrive()` currently only logs and skips missing cold storage payloads before uploading the database backup, `restoreColdStorageFromDrive()` also continues when a referenced payload is missing from Drive, and local/Drive restore calls `setColdStorageItem()` without checking whether the write actually succeeded. In those cases, the app can still end up with a restored or newly uploaded database that contains cold storage pointers without the matching payloads.
> 
> Could you take another look at whether these partial backup/restore cases should be surfaced to the user before the database is applied or uploaded?

That makes sense. Since cold storage payloads are required by the database references, we should not treat missing cold storage data the same way as missing image assets.

I’ll update this PR so backup/restore/account migration verifies required cold storage payloads before uploading or applying the database. If any referenced payload is missing, invalid, or fails to write, the flow should fail or surface a clear error instead of completing with a database that contains dangling cold storage pointers.

I’ll keep recovery of already-missing historical payloads out of scope, but this PR should prevent creating new pointer-only backup/sync states.

cubicj (commented):
Thanks for the follow-up. The stricter cold storage checks cover the local and Drive restore paths much better, but I still see two ordering cases that can leave partial state behind.

In account sync migration, the normal storage keys are written to `AccountStorage` before `collectColdStorageBackupPayloads()` can fail. If a referenced cold payload is missing or cannot be uploaded, the migration returns false, but the remote account can already have database/assets from the failed attempt.

For Drive sync, `risu_lastsaved` is updated before `restoreColdStorageFromDrive()` succeeds. If cold storage restore fails, the DB is not applied, but that remote sync can be skipped on the next run because it has already been marked as seen.

Could you check those ordering cases?

SameDesu123:
Thanks, good catch. I pushed `b78c30b` to cover both ordering cases.

For account sync migration, cold storage payloads are now collected and uploaded before the normal account storage keys are copied. I also skip copying `database/database.bin` in the early key-copy step, so if a referenced cold storage payload is missing or cannot be uploaded, the migration returns before writing the migrated database or other normal account data.

For Drive sync/restore, `risu_lastsaved` is now updated only after `restoreColdStorageFromDrive()` succeeds and the database has actually been written locally. So a failed cold storage restore should no longer cause that remote sync version to be marked as already seen.

This still does not try to recover historical pointer-only data, but it should prevent these backup/sync paths from creating or hiding new partial cold storage states.

SameDesu123:
And fix merge conflict from main

cubicj (commented):
Sorry for another follow-up, but I noticed one account migration case that still looks risky.

Character cold storage stores the full character object, but those payloads are uploaded before the replacement map from `AccountStorage.setItem()` exists. That can leave asset refs inside the cold payload on old local keys while the main DB is rewritten to account-renamed keys.

Could you check whether cold storage payloads should be rewritten with the same replacement map before upload? Also, since the restore path now validates cold storage payload shape, could `collectColdStorageBackupPayloads()` reject invalid payloads too before backup/account migration accepts them?

SameDesu123:
Thanks for the follow-up. I pushed it to cover this account migration case.

The migration now builds the account replacement map before uploading cold storage payloads, then rewrites character cold storage payload resource refs with that same map before sending them to account cold storage. The migrated database is still written last, so a cold storage upload failure should not leave an account database pointing at unavailable cold payload data.

I also tightened `collectColdStorageBackupPayloads()` so backup and account migration reject invalid cold storage payload shapes up front, matching the restore-side validation instead of accepting any truthy payload.

cubicj (approved):
Looks good. The remaining account migration concern looks covered now: the replacement map is built before cold payload upload, and character cold payload refs are rewritten with the same map before the account DB is written.

I also rechecked the local/Drive backup and restore paths, and missing or invalid cold storage data now fails before a partial database is applied or uploaded. `pnpm check`, `pnpm test`, and `pnpm build` pass locally.

sub-hub:
Thanks for this PR, It was really needed fix.

Small suggestion: failing the entire operation when cold storage is missing might be a bit strict. Maybe let it complete but show a warning popup before proceeding?

For restore: detect missing cold storage keys, resolve them to character names from the skeleton DB (the skeleton keeps `name`/`chaId`, so we can tell which characters are affected). Show a popup like "Bot A, Bot B — data may be missing. Continue anyway?" before applying the database. User can cancel or proceed.

For backup: same approach — warn what's missing, let the user decide whether to create a backup that lacks cold storage data.

One note on the level of detail: we can resolve missing keys to character names, but not to individual chat names — when a character is in cold storage, the real chat list is inside the cold storage payload, so the skeleton DB can't tell us which specific chats are affected for that character. Still, character-level info is already much more useful than a generic count.

What do you think about this approach of giving users a choice? This is just a suggestion, so please let me know if you have any other ideas.

SameDesu123:
Thanks, that approach makes sense. I’ve updated the implementation so missing or invalid cold storage keys are mapped to affected character names from the skeleton database.

For manual local and Drive backup/restore, the warning now shows the affected characters and lets the user cancel or continue with the incomplete operation. If a key cannot be mapped to a character, the warning also includes a generic unresolved-item count.

Automatic Drive sync and account migration still fail instead of allowing partial state, since those flows should not silently create dangling cold storage references.

I also added focused coverage for whole-character, stored-chat, and active-chat cold storage key mapping.

sub-hub (approved):
Reviewed the full diff and traced the cold storage error paths. Found one issue to fix and pushed a small follow-up:

**`coldStorageVerifyFailed` was misleading in `changeChar`**

The existing `coldStorageVerifyFailed` message says "your chat data has been preserved", which is correct in `makeColdDataForChat` (write path — the original data is still in memory when verification fails and the replacement hasn't happened yet). But it was also used in `changeChar` (read path — the cold storage payload is gone and the skeleton DB has no real data). There's nothing "preserved" in that case; the data is already lost.

Added `coldStorageRestoreFailed` with an honest message ("could not be loaded, may be permanently lost") and switched `changeChar` to use it. `coldStorageVerifyFailed` stays unchanged for the write path where it's accurate.

Everything looks satisfactory now! It is also good that you showed the user a list of characters affected by the missing data and provided an option to proceed. You implemented it well.

Again, Thanks for the PR.

PR #1485, opened 2026-06-14
refactor(docker): support build-time args for local deployments 

nevaeh5379:
## PR Checklist

- Required Checks

## Summary

I modified the docker-compose.yml file to allow building and running from source code rather than using pre-built images.

## Related Issues

None

## Changes
- Dockerfile: Added `ARG VITE_RISU_LEGAL_CONFIGURED=FALSE` and mapped it to `ENV VITE_RISU_LEGAL_CONFIGURED` in the builder stage. This allows passing the parameter down to Vite's build-time environment variables.
- docker-compose.yml: Configured the service to build locally from the source files using the `build` block instead of pulling the pre-built image, passing `VITE_RISU_LEGAL_CONFIGURED` as a build argument (defaulting to `FALSE`).

## Impact

1. Users can now build and run Risuai directly from the source code using docker-compose.
2. The build-time `VITE_RISU_LEGAL_CONFIGURED` flag can now be customized during the Docker build process. The default value is set to `FALSE` to respect developer guidelines, but users can manually edit it to `TRUE` if needed.

### Scenario 1: Using Docker Compose (Recommended)
1. Modify the argument value in `docker-compose.yml` to `TRUE`:
   ```yaml
   args:
     - VITE_RISU_LEGAL_CONFIGURED=TRUE
   ```
2. Build and run the container:
   ```bash
   docker compose up --build -d
   ```

### Scenario 2: Using Docker CLI
1. Build the image with the custom build argument:
   ```bash
   docker build --build-arg VITE_RISU_LEGAL_CONFIGURED=TRUE -t risuai:local .
   ```
2. Run the container:
   ```bash
   docker run -d -p 6001:6001 -v risuai-save:/app/save --restart always --name risuai risuai:local
   ```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,6 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --prod --frozen-l
 # ------------------------------------------------------------------------------------------
 
 FROM deps AS builder
-ARG VITE_RISU_LEGAL_CONFIGURED=FALSE
-ENV VITE_RISU_LEGAL_CONFIGURED=$VITE_RISU_LEGAL_CONFIGURED
 COPY . .
 # Install including dev deps
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,8 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --prod --frozen-l
 # ------------------------------------------------------------------------------------------
 
 FROM deps AS builder
+ARG VITE_RISU_LEGAL_CONFIGURED=FALSE
+ENV VITE_RISU_LEGAL_CONFIGURED=$VITE_RISU_LEGAL_CONFIGURED
 COPY . .
 # Install including dev deps
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
--- a/README.md
+++ b/README.md
@@ -57,9 +57,23 @@ You can get detailed information on https://github.com/kwaroran/Risuai/wiki (Wor
 
 You can also run Risuai using Docker. This method is particularly useful for web hosting.
 
+#### Using a pre-built image
+
 1. Run the Docker container:
    ```
    curl -L https://raw.githubusercontent.com/kwaroran/Risuai/refs/heads/main/docker-compose.yml | docker compose -f - up -d
    ```
+
+2. Access Risuai at `http://localhost:6001` in your web browser.
+
+#### Building from source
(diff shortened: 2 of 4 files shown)


cubicj (commented):
The local build direction makes sense for passing Vite build-time envs, but this changes the default Compose path from pulling `ghcr.io/kwaroran/risuai:latest` to requiring a source checkout.

The README install command pipes `docker-compose.yml` through `docker compose -f - up -d`; with the new `build.context: .`, that context resolves to the caller's current directory, not the RisuAI repo. This can make the documented Docker install fail unless the user already cloned the source and runs it from the repo root.

One more concern: the `FALSE` truthiness issue is not introduced by this PR, but setting `VITE_RISU_LEGAL_CONFIGURED=FALSE` as the Docker build default makes that existing behavior affect the default Docker build. Since Vite exposes env values as strings, the current frontend check treats `FALSE` as configured, which seems to undercut the intended safe default.

If we want to switch the default Docker path to source builds, the README install instructions would need to change as well. I think that part needs a bit more design discussion before changing the default Compose behavior.

SameDesu123:
@cubicj What do you think about this approach in #1507? It uses HTML injection rather than build-time arguments.

cubicj:
@SameDesu123
I haven't done a full review of #1507 yet, but directionally it looks close to #1472 to me, just a bit more complete.

I like that it keeps the existing Docker image flow and handles the runtime legal config on the Node server side. The explicit flag parsing also looks cleaner for cases like `FALSE`.

nevaeh5379:
@cubicj 
In light of the other related PRs, I removed the changes related to `VITE_RISU_LEGAL_CONFIGURED` from this PR.

I also restored `docker-compose.yml` to use the pre-built image as before. To preserve the option to build from source, I added a separate `docker-compose.build.yml` file and updated the README with instructions for both approaches.

This keeps the existing one-line installation flow working while allowing users with a source checkout to build locally.

cubicj (approved):
I went through the new revision against my earlier comment first. docker-compose.yml and the Dockerfile are byte-identical to main now, so the curl-piped install path is untouched, and the context problem is gone since the source build runs from a cloned repo root where `context: .` points at the right place. The legal flag handling is out of the diff entirely, which leaves that question to #1472.

The new docker-compose.build.yml is the existing compose file with only `image:` swapped for `build:`. Both files share the `risuai` project name and the `risuai-save` volume, so switching between the pre-built and source paths reuses the same save data and replaces the same container instead of colliding. That looks like the right behavior for an alternative path.

One thing worth noting: a source build still lands on the legal screen until #1472 merges, since nothing sets `VITE_RISU_LEGAL_CONFIGURED` at build time. Putting a `.env` with `VITE_RISU_LEGAL_CONFIGURED=TRUE` in the checkout before building does work today, the builder stage copies it in and Vite reads it at build time.

PR #1488, opened 2026-06-17
fix: decrypt local backup data before loading

cubicj:
## PR Checklist

- Required Checks

## Summary

Fixes encrypted local backup loading for account-synced web backups.

The encrypted backup path was fetching the stored crypto key correctly, but then called `encryptBuffer` while handling `database.risudat`. That re-encrypted the already encrypted database bytes instead of decrypting them, causing `decodeRisuSave` to receive invalid data and eventually fail with fflate errors such as `invalid distance`.

## Related Issues

None.

## Changes

**`src/ts/drive/backuplocal.ts`**
- Import `decryptBuffer`
- Use `decryptBuffer(db, key)` when loading an encrypted `database.risudat`
- Keep the save path unchanged, where `encryptBuffer(dbData, key)` is still used to create encrypted backups

## Impact

- Encrypted local backups created from account-synced `risuai.xyz` web sessions can be loaded correctly.
- Non-encrypted local backups are unaffected because they do not enter the `encryption.risudat` path.
- Partial local backups are unaffected because they are not encrypted by this path.
- The save format does not change.

## Additional Notes

Validated with:

- `pnpm check`
- `pnpm test`
- `pnpm build`

### DIFF
--- a/src/ts/drive/backuplocal.ts
+++ b/src/ts/drive/backuplocal.ts
@@ -6,7 +6,7 @@ import { isTauri } from "src/ts/platform"
 import { decodeRisuSave, encodeRisuSaveLegacy } from "../storage/risuSave";
 import { getDatabase, setDatabaseLite } from "../storage/database.svelte";
 import { relaunch } from "@tauri-apps/plugin-process";
-import { encryptBuffer, sleep } from "../util";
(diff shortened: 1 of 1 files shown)


mcpejs (approved):

sub-hub (approved):
It seems like it is clearly an error. Thank  you!

PR #1486, opened 2026-06-17
refactor: split alert UI components and redesign overlay alert

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions?
    - [x] Have you tested your changes?
    - [x] Have you checked that it won't break any existing features?
- [ ] If your PR uses models[^1], check the following:
    - [ ] Have you checked if it works normally in all models?
    - [ ] Have you checked if it works normally in all web, local, and node-hosted versions? If it doesn't, have you blocked it in those versions?
- [x] If your PR is highly AI generated[^2], check the following:
    - [x] Have you understood what the code does?
    - [x] Have you cleaned up any unnecessary or redundant code?
    - [x] Is it not a huge change?
       - We currently do not accept highly AI generated PRs that are large changes.

[^1]: Modifies the behavior of prompting, requesting, or handling responses from AI models.
[^2]: Over 80% of the code is AI generated.

## Summary

Refactors the shared alert UI into focused components and adds schema-driven settings alert callouts.

## Related Issues

None

## Changes

- Split alert rendering into shared host, toast, callout, severity icon, and dialog components.
- Added an alert descriptor model with focused tests for surface routing and severity defaults.
- Registered a `SettingRenderer` alert type and moved advanced/language setting warnings to schema-defined alert blocks.
- Added localized labels for settings alert variants.

## Impact

Alert dialogs, request log dialogs, and toast behavior should remain functionally equivalent while the implementation becomes easier to share and extend. Settings pages can now render inline alert callouts through the existing renderer path.

## Additional Notes

Validation:
- `git diff --check upstream/main...HEAD`
- `git merge-tree --write-tree upstream/main HEAD`
- `pnpm check`
- `pnpm test src/ts/alertModel.test.ts`

PR #1491, opened 2026-06-20
feat: add setting renderer alert type

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Adds a reusable `alert` setting item type for inline informational callouts in data-driven settings pages.

## Related Issues

Part of the replacement split for #1486.

## Changes

- Adds `SettingType = 'alert'` and alert-specific setting options.
- Registers `SettingAlert.svelte` in the settings renderer registry.
- Supports info, warning, error, success, and neutral callout variants.

## Impact

Existing settings behavior is unchanged unless a settings data file opts into the new `alert` type.

## Additional Notes

Validation:

- `pnpm check`

### DIFF
--- a/src/lib/Setting/Wrappers/SettingAlert.svelte
+++ b/src/lib/Setting/Wrappers/SettingAlert.svelte
@@ -1,0 +1,71 @@
+<script lang="ts">
+    import type { Component } from 'svelte';
+    import type { SettingItem, SettingContext } from 'src/ts/setting/types';
+    import { language } from 'src/lang';
+    import { getLabel } from 'src/ts/setting/utils';
+    import { CircleAlert, CircleCheck, CircleX, Info, TriangleAlert } from '@lucide/svelte';
+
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I mainly checked whether registering the new type could affect any existing settings behavior. Nothing else reads `options.title` or `options.variant`, nothing iterates the registry, and no data file defines a `type: 'alert'` item yet, so current rendering paths are untouched until a follow-up PR actually uses it. The wrapper follows the existing renderer conventions and falls back to the info variant on unknown data.

I also traced where alert items will surface once the later split PRs define them. The settings search and the custom sidebar both go through `getFullSettingsData` and `SettingRenderer`, so they render consistently there as well. `pnpm check`, `pnpm test`, and `pnpm build` pass locally.

SameDesu123:
A few months have passed since I opened this PR, and I think it would be better to review the changes again.

For that reason, I’m closing this PR for now.

PR #1492, opened 2026-06-20
refactor: split alert ui components

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Splits the large alert component into smaller alert UI components while preserving the existing alert behavior.

## Related Issues

Part of the replacement split for #1486.

## Changes

- Adds a typed alert descriptor model for legacy alert types.
- Moves modal/dialog/toast rendering into `AlertHost` and extracted dialog components.
- Keeps `AlertComp.svelte` as a compatibility wrapper.
- Adds localized action labels used by the extracted alert UI.

## Impact

This is intended as a behavior-preserving refactor. It keeps the legacy toast alert path so follow-up toast changes can be reviewed separately.

## Additional Notes

This is PR 2 of 4. Later PRs build on this alert host split.

Validation:

- `pnpm check`
- `pnpm exec vitest run src/ts/alertModel.test.ts`


SameDesu123:
Actual diff for this PR: +1,369 -1,148

cubicj (approved):
I mainly checked the behavior-preserving claim, since the whole modal chain moved into `AlertHost`. A whitespace-normalized diff against the old `AlertComp` shows the modal markup is unchanged apart from the newly localized labels, and the `getAlertDescriptor` surface routing resolves every legacy alert type to the same branch the old type-exclusion gates did. The state that moved into the extracted dialogs is also fine. Since `App.svelte` only mounts this component while an alert is open, unmount already covers what the old `$effect.pre` resets covered.

One small behavior change worth noting is the 1.2s fallback timer in `AlertToast`. It fixes toasts that never close when animation events don't fire, but it also caps the toast lifetime for anyone extending the `.toast-anime` duration through custom CSS. I don't think that needs to block anything, just flagging it since the PR is billed as behavior preserving. `pnpm check`, `pnpm test`, and `pnpm build` pass locally.

SameDesu123:
A few months have passed since I opened this PR, and I think it would be better to review the changes again.

For that reason, I’m closing this PR for now.

PR #1497, opened 2026-06-22
feat: Alert user with fix options when chat fmIndex is out of range

sub-hub:
## PR Checklist

- Required Checks

## Images
image
fmindex is completely out of range, but it shows the bot's first fm.

image

Send a chat also works very well.

image

## Summary
The first message index in the chat data is stored separately from the bot, so the first message index in the chat can be larger than the size of the first message the bot has when the chat put in to the another bot.

So places that use fmindex should have exceptions, but they don't have them now.
I Added `?? firstMessage` fallback to all `alternateGreetings[fmIndex]` lookups across DefaultChatScreen, cbs.ts, index.svelte.ts, scripts.ts, and triggers.ts

## The Popup

When `fmIndex` goes out of bounds, the user is shown a popup with two options:

- **A.** Use the default first message
- **B.** Use an empty first message

The popup check triggers in two scenarios:

1. **Before sending a chat** — validates `fmIndex` right before building the prompt. If out of bounds, the popup blocks sending until the user fixes it.
2. **After importing chat data** — validates `chat index 0` right after import. For `risuAllChats`, only the first item is checked, but since the validation repeats on send, this isn't a major issue.

## Concern
There are similar `fmIndex` issues that were **not** fixed in this PR, as they are far removed from the original problem:
- `HypaV3Modal/utils.ts` — returns `null` instead of `firstMessage`
- `branches.ts` — uses `?? 0` instead of `?? firstMessage`
- `characters.ts` (exportChat HTML mode) — uses `?? 0` instead of `?? firstMessage`

## Reproduction
```json
{
    "type": "risuChat",
    "ver": 2,
    "data": {
        "message": [],
        "note": "",
        "name": "New Chat 1",
        "localLore": [],
        "fmIndex": 999,
        "id": "eded059d-5f9e-49db-b51a-bd1228b2369d"
    },
    "folders": []
}
```
Try to import the above json into any bot. You will get an error if you try to send a chat.

## Details
To be more specific, I have **configured the bot to use the first fm if there is no data corresponding to fmindex**.
This applies to the backend, frontend, cbs, triggers, and even @@repeat_back.

**We must decide whether to use the first fm or the last fm when a value does not exist in the data using fmindex going forward.**

If my approach (Use first fm) differs from yours, please let me know.


sub-hub:
Oh, and if fmindex is out of range and you want to switch to a different fm, simply pressing the right arrow key returns fmindex to the beginning.

So I didn't change this behavior.

cubicj (commented):
The fallback direction looks right. The six `?? firstMessage` guards are the same change as #1353, which I approved before it was closed, and they are still missing on main. `pnpm check`, `pnpm test`, and `pnpm build` pass locally on this head.

On what to fall back to, I think the first message is the better default. An empty string goes straight through `{{previouscharchat}}`, `@@repeat_back`, and `v2GetFirstMessage` to consumers that cannot tell it apart from a greeting that is genuinely empty. A wrong greeting is at least visible to the user. An empty one is not.

Where I would push back is Option B. Adding an entry to `alternateGreetings` repairs one chat by changing the character. The blank greeting shows up in the character editor, adds a page to the first message swipe count on every other chat of that character, and goes out with the card on export. It also fails at the thing this function is for. Once `fmIndex` is normalized to `-1`, the three files you listed as out of scope recover on their own, since `getFirstMessage()` in `HypaV3Modal/utils.ts`, the HTML export in `characters.ts`, and `branches.ts` all check `fmIndex === -1` before touching the array, and the page counter recovers with them. Option B does not get you that. It points the chat at an empty string, and `getFirstMessage()` tests the greeting for truthiness, so that chat reads back as `null` there instead.

The `.length` reads in the same block of `DefaultChatScreen.svelte` are still unguarded, and that is what creates the case the popup is really for. With no alternate greetings at all, `altGreeting` is false, the arrows never render, and a stale `fmIndex` cannot be cleared by hand. With at least one greeting the right arrow already clears it, as you said, at least for values above the range.

A few smaller things. Every return path in `validateAndFixFmIndex` yields `true`, so the `if (!fmValid)` branch in `sendChat` never runs and the doc comment describes a failure that cannot happen. The range check accepts a negative `fmIndex` other than `-1`, and it accepts a non-integer, so both survive the popup. The `else` also makes Option B the fallthrough for anything that is not `'0'`, which puts the destructive choice on the unexpected path. The new import into `process/index.svelte.ts` creates a cycle with `characters.ts`, which already imports `doingChat` from there. When the visible chat is not already index 0, the popup can appear for a chat the user is not looking at. `git diff --check` also flags trailing whitespace in the new doc comment.

One more thing, outside this PR. The same mismatch happens without going out of range at all. Deleting an alternate greeting only resets `fmIndex` on the currently open chat, reordering does not touch it, and the MCP character write replaces the whole array. Other chats keep an index that is still in range but now points at different text. Nothing catches that today, and no fallback can. I think it belongs in its own issue rather than here.

How would you like to handle the popup, given that normalizing `fmIndex` covers the remaining paths on its own?

sub-hub:
### Option B staying, here's why

When you import a chat from another character, the imported chat's fmIndex can point at a greeting that doesn't exist on the current character. That mismatch is the root problem this PR fixes. The fallback to firstMessage handles the silent case; Option B is for when the user deliberately wants no greeting, common enough when the existing chat history already carries all the context. It's an opt-in choice with a clear description in the popup, not a hidden side effect.

### The import cycle fix (biggest one)

To resolve this, the number of modified files ended up being quite large—a total of 11 files. However, since the changes are limited to just the import paths for `doingChat`, it is actually very simple.
`doingChat` was the only reason characters.ts had to pull in process/index.svelte.ts in this PR, I checked with git blame and it's been sitting in process/index.svelte.ts since literally the **Risuai 0.6.3 first commit** three years ago. Moved it to stores.svelte.ts where it actually belongs, next to `DBState` and `selectedCharID`. Cycle gone.

### Everything else

All the smaller things you caught are fixed.

The `.length` accesses in DefaultChatScreen are now guarded with `?.length ?? 0`. Negative and non-integer fmIndex values get rejected before the popup. Cancelling the popup returns false, so the `if (!fmValid)` branch in sendChat isn't dead anymore. The else-fallthrough is replaced with an explicit `else if (choice === '1')`. All import sites call `changeChatTo` before showing the popup. Doc comment matches what the function actually does.

### The array mutation problem you mentioned

Agreed this is real, but it needs its own ticket. It looks like resolving this issue will require quite a few modifications.

---

@cubicj Sorry for the mess. I treated this like a easy fix and shipped a bunch of careless mistakes that never should've left draft. Really appreciate the thorough review.

sub-hub:
@cubicj Realized the "use empty first message" option didn't make the permanent effects clear enough, someone could reasonably think it's temporary change.
Added a confirmation popup that explicitly warns the empty greeting will be included when exporting the bot. Should be good now.

cubicj (commented):
I checked the follow-ups on b2ba6fb6c and the small ones are all good. The four `.length` reads in `DefaultChatScreen` are guarded, negative and non-integer values now fall through to the popup, `if (!fmValid)` is reachable through the confirm decline, and all five `importChat` sites switch the chat first. Moving `doingChat` into `stores.svelte.ts` is worth doing on its own. It does not leave the graph acyclic, since `process/index.svelte.ts` still reaches `characters.ts` and comes back through `globalApi.svelte` and `hotkey.ts`, but that loop already exists on main without `characters.ts` in it. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

Then I went looking for how the codebase already says "do not send a first message", since that is what Option B is for. It turns out there is a mechanism. `makeMs()` inside `sendChat` sets `msReseted` as soon as it hits a message with `disabled === 'allBefore'`, and the block that appends the first message only runs while `msReseted` is false. The scissors button in the message menu, `language.disableAbove`, is what sets that flag. It lives on the message, so it is chat scoped, and it never touches the character.

Option B does not set that flag. It points `fmIndex` at an empty string, so `sendChat` still builds the assistant turn and still pushes it, just with empty content. A user who picks it does not end up with a chat that has no greeting. They end up with an empty greeting turn in the prompt, plus a greeting entry on the character that shows in the editor, counts toward `totalPages` on every other chat of that character, and ships with the card on export. The new confirmation makes that second part visible, which is a real improvement, but the first part is the thing the option was supposed to deliver.

The other half is what I raised last time. Setting `fmIndex` to -1 also fixes the three files you listed as out of scope, because `getFirstMessage()` in `HypaV3Modal/utils.ts`, the HTML export in `characters.ts`, and `branches.ts` each check `fmIndex === -1` before they index the array, and the page counter recovers along with them. Option B reaches none of them. `getFirstMessage()` tests the greeting for truthiness, so an empty entry reads back as `null` there instead.

Two smaller things. `validateAndFixFmIndexEmptyConfirm` states that a new empty greeting will be added, but it runs before the `indexOf('')` reuse check, so it also appears on the path where an existing empty entry is reused and nothing is added. And `importChat` discards the return value and calls `alertNormal(language.successImport)` either way, so declining that confirmation during import reports a successful import while `fmIndex` is still out of range.

Thanks for turning these around so quickly. Does `disableAbove` look to you like it already covers the case Option B is for, or is there something about it that does not fit?

sub-hub:
Ah, I also got my hopes up for a moment, thinking I had missed a good method. I looked into it, but disableAbove won't work for first messages.

The first message is never stored in `chat.message`. sendChat reads it straight from the character — `currentChat.fmIndex === -1 ? nowChatroom.firstMessage : nowChatroom.alternateGreetings?.[currentChat.fmIndex]`. It's a plain string, not a `Message` object with a `disabled` field. makeMs() only walks `chat.message`, so there is nothing to attach the `'allBefore'` flag to. The two data paths are completely separate — msReseted is the only thing connecting them, and it only gets set by makeMs() hitting a flagged entry inside the message array.

Frontend side, the scissors button accesses `message[idx]` directly. The first message display passes `idx={-1}`, so `message[-1]` is `undefined` and `.disabled` access throws. Even if we guarded that, setting disabled on `message[-1]` overwrites the last array element, not the first message.

So the idea was reasonable but the mechanism just doesn't reach first messages at all. First message blocking would need its own chat-scoped flag, which is a separate feature from what this PR is trying to fix.

The import return value thing is fixed too — the five importChat sites now only call alertNormal when validateAndFixFmIndex returns true, so declining the popup during import no longer reports success with a broken fmIndex.

Honestly, I think adding disableAbove support for first messages would get complicated fast. This is because the task requires modifications to both the front and backend.
Fixing the other issues that come up when using an empty first message feels more like straightforward bug fixes, so I think it's fine to keep Option B as is and deal with those as they come up.

sub-hub:
Anyway, regarding the `disableAbove` option you mentioned—while it turned out not to work on the backend either, I hadn't even checked that aspect because the option isn't visible on the frontend.
That’s why I find conversations like this so helpful. Thanks.

cubicj (approved):
You are right about `disableAbove`, it never reaches the first message. Checking that sent me back to my own claim about the empty assistant turn, and my claim was wrong. I'm sorry for pushing back on Option B across two rounds based on it. sendChat does push the greeting turn, but it never reaches the request. The `unformated.chats` build filters out empty content, and `pushPrompts` skips empty prompts again when the sections are assembled. So an empty greeting ends up in the same state as the existing `{{none}}`/`{{blank}}` convention, which the chat screen already renders as a blank first message. Option B lands in a state the app already supports.

One leftover from last time. `validateAndFixFmIndexEmptyConfirm` still runs before the `indexOf('')` reuse check, so it warns about adding a greeting on the path where an existing empty entry is reused and nothing is added. I don't think that needs to block anything, just flagging it.

The success alert is now gated on the validator result at all five `importChat` sites, and `pnpm check`, `pnpm test`, and `pnpm build` are clean on 893ca78e5.

sub-hub:
Please don't feel bad about it. `disableAbove` was something that definitely needed to be checked, and I was only able to catch it because you pointed it out—thank you.
Also, the other corrections you suggested addressed areas that genuinely had issues.
It really makes me realize just how many things I had overlooked. Next time, I’ll make sure to take the time to review everything thoroughly before open the PR. Thank you.

PR #1502, opened 2026-06-24
fix: update openrouter attribution headers

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This PR updates the OpenRouter attribution headers to use the current OpenRouter-specific header name and reflects the recent project name change from `RisuAI` to `Risuai`.

## Changes

* Replaced the legacy `X-Title` OpenRouter attribution header with `X-OpenRouter-Title`
* Updated the OpenRouter app title value from `RisuAI` to `Risuai`
* Added `X-OpenRouter-Categories` for clearer OpenRouter app categorization

## Notes

This PR does not change internal Risu proxy identification headers, since those may be used for compatibility outside of OpenRouter attribution.

### DIFF
--- a/src/ts/process/request/openAI/requests.ts
+++ b/src/ts/process/request/openAI/requests.ts
@@ -553,8 +553,9 @@ export async function requestOpenAI(arg:RequestDataArgumentExtended):Promise<req
         headers["Authorization"] = "Bearer " + db.OaiCompAPIKeys[arg.modelInfo.keyIdentifier]
     }
     if(aiModel === 'openrouter'){
-        headers["X-Title"] = 'RisuAI'
+        headers["X-OpenRouter-Title"] = 'Risuai'
         headers["HTTP-Referer"] = 'https://risuai.xyz'
+        headers["X-OpenRouter-Categories"] = 'roleplay,general-chat'
     }
     if(aiModel === 'nanogpt' && db.nanogptProvider){
         headers["X-Provider"] = db.nanogptProvider


SameDesu123:
OpenRouter Docs: https://openrouter.ai/docs/app-attribution#x-openrouter-title

cubicj (approved):
I checked the header names against the current OpenRouter app attribution docs first. `X-OpenRouter-Title` is the current name and `X-Title` only remains for backwards compatibility, so nothing was broken before, this just moves off the legacy name. `roleplay` and `general-chat` are both recognized categories, and app identity on their side is keyed by `HTTP-Referer`, which this PR keeps, so the rankings entry carries over with the display name updated to match the e6ee605cd rename.

Then I traced whether the renamed headers survive every transport, since this is the only `X-Title` usage in the repo and the same `headers` object feeds both the streaming and non-streaming paths. The web proxy and the node server pass header names through `risu-header` without an allowlist, and the Tauri side accepts arbitrary names through `HeaderName::from_bytes`. The one path that could actually break is plain fetch on web, where the browser preflights openrouter.ai directly. I probed that preflight, and their `access-control-allow-headers` already lists `X-Openrouter-Title` and `X-Openrouter-Categories`, so that path is fine too.

One thing worth noting is that categories are additive on OpenRouter's side. Once sent they merge into the app entry and stay visible on their apps page. The chosen values fit the app, so I don't see a problem, just flagging it since it is publicly visible.

`pnpm check`, `pnpm test`, and `pnpm build` pass locally on this head.

cubicj:
I found this while going through #1510, and it is something I should have caught when I approved this at b8fccdc35. Renaming `X-Title` to `X-OpenRouter-Title` also changes what users can reach through Additional Parameters, since `X-Title` was both the header RisuAI sent and the only key a user could set to override it or remove it.

Two things change for anyone on the built in openrouter provider with `applyAdditionalParamsToAll` enabled. `header::X-Title={{none}}` used to strip the app title from the request, and it now deletes a key that is no longer being sent while `X-OpenRouter-Title: Risuai` goes out unchanged, so there is no way left to opt out. And `header::X-Title=MyApp` used to replace the app title through an ordinary same key assignment, where it now leaves both `X-OpenRouter-Title: Risuai` and `X-Title: MyApp` on the request.

I reproduced both against this head and against main through `requestOpenAI` with `previewBody`. The inline block in requests.ts runs before `applyAdditionalParameters`, so after the rename the user's value lands next to the app value rather than over it.

`HTTP-Referer` and `X-OpenRouter-Categories` are not affected, since both keep names a user would type as is, so overriding or removing either still works the way it did.

The rename itself looks right to me. The problem is only that the name users have already configured is no longer the one being sent. #1510 removes this block entirely and carries the same header names, and it covers the `{{none}}` case there but not the override, so the same gap exists in a different shape. Since both are yours, either place works from my side, whether this one carries the fix or #1510 takes the headers over and handles it there. How would you like to sequence them?

SameDesu123:
Thanks for catching this. I agree that #1510 is the better place to handle the compatibility behavior first, since it centralizes OpenRouter attribution.

I’ll update #1510 so existing `X-Title` Additional Parameters continue to work for both `{{none}}` removal and value overrides, then adjust this PR to follow that shared behavior. That should keep the migration in one place and avoid duplicating the alias handling here.

PR #1503, opened 2026-06-24
fix: include presetRegex in edittrans translation regex application

miladiashe:
## PR Checklist

- Required Checks

## Summary

applyEdittransRegex was only collecting module and character scripts, missing db.presetRegex (preset/prompt-side regex scripts). This caused edittrans regex placed in the prompt preset to have no effect on any translation type, including LLM translation. Aligns collection order with processScriptFull: presetRegex → module → character.

## Related Issues

None

## Changes

기존에 번역문 수정 정규식이 원래부터 프롬프트 정규식 스크립트를 무시하고 적용되고 있었고, 제가 LLM 번역에도 번역문 수정 정규식이 작동되게 수정할 때 그걸 단순히 가져다 써서 누락되어 있었습니다. 누락된걸 뒤늦게 확인해 적용했습니다.

## Impact

프롬프트 소속의 작동하지 않던 정규식이 이제 작동할 뿐입니다.

### DIFF
--- a/src/ts/translator/edittransRegex.test.ts
+++ b/src/ts/translator/edittransRegex.test.ts
@@ -12,9 +12,14 @@ vi.mock("../process/modules", () => ({
     moduleUpdate: () => {},
 }));
 
+const parserCalls: { data: string, chatID: number }[] = [];
+
 vi.mock("../parser/parser.svelte", () => ({
     applyMarkdownToNode: () => {},
-    risuChatParser: (data: string) => data.replaceAll("{{char}}", "Risu"),
+    risuChatParser: (data: string, arg: { chatID?: number } = {}) => {
+        parserCalls.push({ data, chatID: arg.chatID ?? -1 });
+        return data.replaceAll("{{char}}", "Risu");
+    },
 }));
 
 import { applyEdittransRegex } from "./translator";
@@ -29,13 +34,14 @@ const script = (v: Partial<customscript>): customscript => ({
 
 const alwaysExistChar = { customscript: [] } as unknown as character;
 
-const apply = (text: string, scripts: customscript[]) => {
+const apply = (text: string, scripts: customscript[], chatID = -1) => {
     database.presetRegex = scripts;
-    return applyEdittransRegex(text, "chatid", alwaysExistChar);
+    return applyEdittransRegex(text, "chatid", alwaysExistChar, chatID);
 };
 
 beforeEach(() => {
     database.presetRegex = [];
+    parserCalls.length = 0;
     vi.spyOn(console, "error").mockImplementation(() => {});
 });
 
@@ -82,6 +88,14 @@ describe("applyEdittransRegex", () => {
         expect(apply("Risu said hi", [
             script({ in: "{{char}}", out: "Seia", ableFlag: true, flag: "g" }),
         ])).toBe("Risu said hi");
+        expect(parserCalls).toEqual([]);
+    });
+
+    it("forwards the chatID to the parser so message scoped syntaxes resolve", () => {
+        apply("Risu said hi", [
+            script({ in: "{{char}}", out: "Seia", ableFlag: true, flag: "g<cbs>" }),
+        ], 7);
+        expect(parserCalls).toEqual([{ data: "{{char}}", chatID: 7 }]);
     });
 
     it("sorts scripts by the order flag, higher first", () => {
--- a/src/ts/translator/edittransRegex.test.ts
+++ b/src/ts/translator/edittransRegex.test.ts
@@ -1,0 +1,129 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { character, customscript } from "../storage/database.svelte";
+
+const database: { presetRegex: customscript[] } = { presetRegex: [] };
+
+vi.mock("../storage/database.svelte", () => ({
+    getDatabase: () => database,
+}));
+
+vi.mock("../process/modules", () => ({
+    getModuleRegexScripts: () => [] as customscript[],
+    moduleUpdate: () => {},
+}));
+
+vi.mock("../parser/parser.svelte", () => ({
+    applyMarkdownToNode: () => {},
+    risuChatParser: (data: string) => data.replaceAll("{{char}}", "Risu"),
+}));
+
+import { applyEdittransRegex } from "./translator";
+
+const script = (v: Partial<customscript>): customscript => ({
+    comment: "",
+    in: "",
+    out: "",
+    type: "edittrans",
+    ...v,
+} as customscript);
+
+const alwaysExistChar = { customscript: [] } as unknown as character;
+
+const apply = (text: string, scripts: customscript[]) => {
+    database.presetRegex = scripts;
+    return applyEdittransRegex(text, "chatid", alwaysExistChar);
+};
+
+beforeEach(() => {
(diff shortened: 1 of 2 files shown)


cubicj (commented):
I traced the fix itself and it checks out. `presetRegex` edittrans scripts were never collected anywhere before, `processScriptFull` filters them out by type, and the new collection reaches all three translation paths through the shared helper. `pnpm check`, `pnpm test`, and `pnpm build` pass locally on this head.

What worries me is that this activates a population of preset scripts that has never been compiled before. `executeScript` in scripts.ts skips scripts with an empty `in` and wraps each script in its own try/catch, so a broken edittrans regex sitting in a shared preset is harmless today. `applyEdittransRegex` has neither guard, so after this merges the same regex compiles on every translated message, and a `new RegExp` throw propagates out of `translateHTML` into the chat display path. Character and module scripts already carry this risk, but presets travel between users, so the exposed surface grows quite a bit here.

How do you think a broken script should behave once these are live? I'd like to hear your take before this activates them as is.

miladiashe:
> I traced the fix itself and it checks out. `presetRegex` edittrans scripts were never collected anywhere before, `processScriptFull` filters them out by type, and the new collection reaches all three translation paths through the shared helper. `pnpm check`, `pnpm test`, and `pnpm build` pass locally on this head.
> 
> What worries me is that this activates a population of preset scripts that has never been compiled before. `executeScript` in scripts.ts skips scripts with an empty `in` and wraps each script in its own try/catch, so a broken edittrans regex sitting in a shared preset is harmless today. `applyEdittransRegex` has neither guard, so after this merges the same regex compiles on every translated message, and a `new RegExp` throw propagates out of `translateHTML` into the chat display path. Character and module scripts already carry this risk, but presets travel between users, so the exposed surface grows quite a bit here.
> 
> How do you think a broken script should behave once these are live? I'd like to hear your take before this activates them as is.

Thanks for pointing this out. I agree that a broken "edittrans" script should fail open rather than propagate into the translation display path.

I reused the existing behavior when adding preset scripts and missed that it did not have the same safeguards as "processScriptFull". I’ll update this PR so that it:

- skips scripts with an empty "in"
- normalizes regex flags before compilation
- catches errors per script and continues with the remaining scripts

That way, an invalid shared preset regex will only be skipped, while the translated message and other valid scripts will continue to work.

miladiashe:
@cubicj 

Pushed the safeguards. `applyEdittransRegex` now fails open, so a broken script is
skipped and the translated message plus every other valid script still applies —
same posture `executeScript` already has for these scripts.

The three things I said I'd do:

- scripts with an empty `in` are skipped
- flags are normalized before compiling: unsupported characters dropped
  (`[^dgimsuvy]`), duplicates dropped, falling back to `u` when nothing is left
- each script runs in its own `try`/`catch` and logs, and the loop continues

While tracing the flag normalization I found something that pushed the scope a bit,
and I'd like your read on it. The FLAGS panel in `RegexData.svelte` is gated on
`ableFlag` only, never on `value.type`, so an `edittrans` script can be given
`<cbs>`, `<order n>`, `<move_top>` or `<move_bottom>` straight from the GUI. Those
are parsed out in `processScriptFull`, whose `ScriptMode` has never included
`edittrans`, so nothing removed them here and they reached `new RegExp` verbatim as
`Invalid flags`. That is the same crash you were worried about, except it is
reachable by clicking a button rather than by a malformed regex.

Normalization alone would have turned it into silent nonsense — `<cbs>` reduces to a
bare `s`, quietly enabling dotAll. So I made the three that are meaningful here
actually work, reusing the logic from `processScriptFull`: `<cbs>` parses curly
braced syntaxes in IN, `<order n>` sorts the scripts, `<move_top>`/`<move_bottom>`
move the match. Only flags, not `@@` prefixes in OUT, so existing OUT strings keep
their literal meaning.

Wiring `<cbs>` up turned out to need one more thing. `translateHTML` already takes
`chatID` and all three `applyEdittransRegex` call sites are inside it, but it was
never passed down, so `risuChatParser` fell back to `-1` and message scoped syntaxes
in IN would have silently resolved to their defaults — unlike `processScriptFull`,
which passes the caller's `chatID`. Threaded it through.

Three are parsed out and ignored instead, because they have no sound meaning on this
path:

- `<inject>` writes back into `chat.message[chatID].data`. The translation display
  path should not mutate stored chat data. It has no GUI button either.
- `<repeat_back>` matches against the untranslated source of the previous same-role
  message and appends that to the translated output. There is no translated
  equivalent to carry forward.
- `<no_end_nl>` is the off switch for the `\n` suffix in `scripts.ts`, which
  `applyEdittransRegex` never applied. Honouring it would mean adding that suffix
  first and changing existing scripts.

None of these ever worked here, so ignoring them is not a regression — they threw
before.

Two pre-existing issues in `scripts.ts` I noticed while tracing this, neither one
touched here:

- the `$<name>` expansion in the move branch runs `parseInt` on the group name, so it
  is always `NaN` and named groups never resolve. I mirrored it verbatim rather than
  fixing it on my side — the same script behaving differently depending on whether it
  is an `edittrans` script or not seemed worse than the feature being uniformly
  broken, and it would hide the divergence from whoever fixes `scripts.ts` later.
  There is a comment marking it so both can be fixed in one change. Happy to do that
  separately if you confirm the current behaviour is unintended.
- `<repeat_back>` reads `r[0]` without a null check, so when the previous message does
  not match either it throws and the outer `try`/`catch` silently drops the whole
  script.

I left this without a cache, unlike `processScriptFull`. `translateHTML` is gated in
`ChatBody.svelte` so it runs about once per message rather than per render, the
surrounding path already pays for a network call or a cache read, and a correct key
would have to run `risuChatParser` for `<cbs>` scripts anyway — `generateScriptCacheKey`
does exactly that. Happy to add one if you'd rather have the parity.

Added `src/ts/translator/edittransRegex.test.ts`, 15 cases covering the fail-open
behaviour, each supported flag, and the ignored ones. `applyEdittransRegex` is
exported so the test can reach it. `pnpm check` (0 errors, 0 warnings), `pnpm test`
(195 passed, 3 skipped) and `pnpm build` pass locally.

Happy to cut the flag support back to plain normalization if you'd rather keep this
PR to the collection fix — it's a separate commit.

cubicj (approved):
I checked the safeguards and the flag support at 4ac399ed4 against processScriptFull, and the mirror holds, including the quirks you kept on purpose. The FLAGS panel really is gated on ableFlag alone with no type check, so these buttons already reach edittrans scripts from the GUI. That settles the scope for me: keep the flag support.

On the cache, your reasoning holds, a correct key would run risuChatParser anyway, so leave it out. And parseInt on a group name cannot be intended, matched.groups[NaN] never resolves, so a separate PR fixing both sites in one change sounds good. Worth a line there that literal $<name> outputs will change.

One thing to note: with ableFlag on and every flag toggled off, the old code replaced only the first match and the new g fallback replaces all of them, same as processScriptFull. The state is reachable since toggleFlag can empty the field. I doubt anyone relies on it, just disclosing the delta. pnpm check, pnpm test and pnpm build are clean on this head with the fifteen new tests.

sub-hub (approved):
The number of changed files is small and I don't expect any side effects. It also adds flag and indicator detection while processing presetRegex, which wasn't there before. Looks good to me.
Thanks for the PR.

PR #1504, opened 2026-06-24
feat(nai): support v4/v4.5 per-character prompts via Lua module API

miladiashe:
## PR Checklist

- Required Checks

## Summary

Wire NAI v4/v4.5 multi-character prompting through `generateAIImage` and expose it to Lua modules so they can supply per-character prompts.

- `generateAIImage` now accepts an optional `naiOptions` param carrying per-character prompts (prompt, negative, x, y).
- Build `v4_prompt` / `v4_negative_prompt` `char_captions` with centers from those options instead of hardcoding empty arrays, and send `parameters.characterPrompts` alongside them.
- Enable `use_coords` automatically when any character has a non-auto center. `use_order` stays `true`, matching every reference client.
- `generateImage` Lua API gains an optional 4th JSON options arg, parsed via `parseLuaOptions` and passed through. Fully backward compatible.

Claude-Session: https://claude.ai/code/session_01Ko5GAw8P9N7548yUaaakgY

## Related Issues

NONE

### `generateImage` options argument

This PR adds an optional, backward-compatible options argument that lets modules pass **per-character prompts** for NovelAI v4 / v4.5 multi-character prompting.

#### Lua API

```
generateImage(id, prompt, negative, options)
```

| Arg | Type | Required | Description |
|-----|------|----------|-------------|
| `id` | string | ✓ | Low-level access trigger id |
| `prompt` | string | ✓ | Base (scene-level) positive prompt |
| `negative` | string | ✗ | Base negative prompt |
| `options` | string (JSON) | ✗ | JSON-encoded options object (see below). Omit, or pass `""`/invalid JSON, to fall back to the legacy single-prompt behavior |

The 4th argument is a **JSON string** (consistent with the existing `LLM` / `axLLM` convention, parsed via `parseLuaOptions`). On the TypeScript side it maps to `NAIImageGenOptions` and is passed to `generateAIImage(...)`.

#### Options schema (`NAIImageGenOptions`)

```jsonc
{
  // Per-character prompts, max 6 (NAI's limit — exceeding it is rejected by the API).
  // Order always matters: the model places characters in listed order
  // (first prompt → left). Coordinates are an additional nudge on top,
  // so keep the order consistent with the positions you set.
  "characters": [
    {
      "prompt":   "1girl, smile",  // required — this character's positive prompt
      "negative": "blurry",        // optional — this character's negative prompt (default "")

      // Character center, 0–0.9. Both optional.
      // 0 (or omitted) means automatic placement. The web UI's position
      // picker is a 5×5 grid over 0.1–0.9. Out-of-range values are clamped,
      // non-numeric values fall back to 0.
      "x": 0.3,
      "y": 0.5
    }
  ],

  // Optional. Forces coordinate placement on or off.
  // Omitted: enabled automatically when any character has a non-auto center.
  // Only an actual boolean is honoured; anything else falls back to auto-detection.
  "use_coords": true
}
```

#### Field semantics

| Field | Default | Notes |
|-------|---------|-------|
| `characters` | `[]` | Empty/absent → legacy single-prompt request (`char_captions` stays empty, no `characterPrompts` key). |
| `characters[].prompt` | `""` | Mapped to `v4_prompt.caption.char_captions[].char_caption`. SD-style `()` emphasis is converted to NAI-style `{}` (same as the base prompt); use `\(` `\)` for literal parentheses. |
| `characters[].negative` | `""` | Mapped to `v4_negative_prompt.caption.char_captions[].char_caption`. Passed through verbatim (no emphasis conversion, matching base-negative behavior). |
| `characters[].x` / `.y` | `0` (auto) | Written to `centers: [{ x, y }]`. `(0,0)` is NAI's `AUTO` placement, not top-left. Values are clamped to `0`–`0.9`; non-numeric values fall back to `0`. |
| `use_coords` | auto | `true` → NAI places characters by `centers`. Auto-detection enables it when any character has a center other than `(0,0)`. Only an actual boolean overrides the auto-detection. |

> **Note:** `use_coords` is global to the request. If any character provides a position, coordinate mode is enabled for **all** characters. When using coordinates, provide positions for every character to avoid overlap.

`use_order` is always sent as `true` and is not derived from `use_coords`. Order and position are reinforcing signals rather than alternatives — this matches novelai-python, Aedial/novelai-api, SillyTavern and the ComfyUI NAI nodes, none of which ever turn it off, and the NAI docs' guidance that positions "should not conflict with the order of your character prompt boxes".

`parameters.characterPrompts` is sent alongside `v4_prompt`, as every reference client does. It is omitted entirely when no characters are supplied, so a plain generation is byte-identical to before.

Example
```Lua
local result = generateImage(id, basePrompt, negPrompt, json.encode({
    characters = {
        { prompt = "1girl, smile", x = 0.3, y = 0.5 },
        { prompt = "1boy",         x = 0.7, y = 0.5 },
    }
})):await()
-- returns "{{inlay::...}}" on success, or "Error: Image generation failed"
```

### Error handling

Malformed options degrade instead of propagating. A wrong-typed field falls back to its documented default, and anything that does throw is caught and returned to the script as `'Error: Image generation failed'` — the same contract a failed generation always had. A bad options object never aborts the script run.

### Structure and tests

The multi-character field building lives in `src/ts/process/naiCharacterPrompts.ts`, separate from `stableDiff.ts` so it can be asserted without pulling in the app's stores. `src/ts/process/tests/naiCharacterPrompts.test.ts` covers the option handling, including malformed input.

## Impact

Existing callers using `generateImage(id, prompt, negative)` are unaffected — the options argument is optional, and when absent the request is identical to before this change: `use_coords: false`, `use_order: true`, empty `char_captions`, and no `characterPrompts` key. Non-NAI providers are untouched.

## Additional Notes

Module used for test: https://drive.proton.me/urls/5ZC3KMR860#BYEluVwBfLgV

How to test: Type /test. Then it generates image using Prompts, negative prompt, character prompt, character negative prompt(Optional), character location(optional) in lorebook

### DIFF
--- a/src/ts/process/naiCharacterPrompts.ts
+++ b/src/ts/process/naiCharacterPrompts.ts
@@ -1,0 +1,74 @@
+/** NAI clamps character centers to 0~0.9, and treats (0, 0) as automatic placement. */
+const NAI_POSITION_MIN = 0
+const NAI_POSITION_MAX = 0.9
+
+export interface NAICharacterPrompt{
+    /** Positive prompt for this character */
+    prompt: string
+    /** Optional negative prompt for this character */
+    negative?: string
+    /** Optional horizontal position (0~0.9). 0 or omitted means automatic placement */
+    x?: number
+    /** Optional vertical position (0~0.9). 0 or omitted means automatic placement */
+    y?: number
+}
+
+export interface NAIImageGenOptions{
+    /** Per-character prompts for NAI v4/v4.5 multi-character prompting */
+    characters?: NAICharacterPrompt[]
+    /** Force coordinate-based placement. Defaults to true when any character supplies a non-zero x/y */
+    use_coords?: boolean
+}
+
+const naiPosition = (value:unknown):number => {
+    if(typeof value !== 'number' || !Number.isFinite(value)){
+        return NAI_POSITION_MIN
+    }
+    return Math.min(NAI_POSITION_MAX, Math.max(NAI_POSITION_MIN, value))
+}
+
+/** Convert SD-style () emphasis to NAI-style {} emphasis, preserving escaped \( \) */
+export const naiEmphasis = (s:string) => s
+    .replaceAll('\\(', "♧")
+    .replaceAll('\\)', "♤")
+    .replaceAll('(','{')
+    .replaceAll(')','}')
+    .replaceAll('♧','(')
+    .replaceAll('♤',')')
+
+/**
+ * Builds the NAI v4/v4.5 multi-character fields from options supplied by modules via Lua.
+ * Kept out of stableDiff.ts so the shape can be asserted without pulling in the app's stores.
+ */
+export function buildNAICharacterPrompts(naiOptions?:NAIImageGenOptions){
+    const naiCharacters = naiOptions?.characters ?? []
+    const charCenter = (c:NAICharacterPrompt) => ({
+        x: naiPosition(c?.x),
+        y: naiPosition(c?.y)
+    })
+    // (0, 0) is NAI's "auto" center, so coordinates only kick in once one of them is placed.
+    // Only an actual boolean overrides that, so a wrong-typed option falls back to auto-detection.
+    const useCoords = typeof naiOptions?.use_coords === 'boolean'
+        ? naiOptions.use_coords
+        : naiCharacters.some(c => charCenter(c).x !== NAI_POSITION_MIN || charCenter(c).y !== NAI_POSITION_MIN)
+
+    return {
+        useCoords,
+        posCharCaptions: naiCharacters.map((c) => ({
+            char_caption: naiEmphasis(c?.prompt ?? ''),
+            centers: [charCenter(c)]
+        })),
+        negCharCaptions: naiCharacters.map((c) => ({
+            char_caption: c?.negative ?? '',
+            centers: [charCenter(c)]
+        })),
+        // The web client and novelai-python both send characterPrompts alongside v4_prompt.
+        // Left undefined without characters so the request stays identical to a plain generation.
+        characterPrompts: naiCharacters.length > 0 ? naiCharacters.map((c) => ({
+            prompt: naiEmphasis(c?.prompt ?? ''),
+            uc: c?.negative ?? '',
+            center: charCenter(c),
+            enabled: true
+        })) : undefined,
+    }
+}
--- a/src/ts/process/scriptings.ts
+++ b/src/ts/process/scriptings.ts
@@ -7,7 +7,7 @@ import { get } from "svelte/store";
 import { DBState, ReloadChatPointer, ReloadGUIPointer, selectedCharID } from "../stores.svelte";
 import { alertSelect, alertError, alertInput, alertNormal, alertConfirm } from "../alert";
 import { HypaProcesser } from "./memory/hypamemory";
-import { generateAIImage } from "./stableDiff";
+import { generateAIImage, type NAIImageGenOptions } from "./stableDiff";
 import { writeInlayImage, getInlayAsset } from "./files/inlays";
 import type { OpenAIChat, MultiModal } from "./index.svelte";
 import { requestChatData, type StreamResponseChunk } from "./request/request";
@@ -360,11 +360,12 @@ export async function runScripted(code:string, arg:{
                 }
             })
 
-            declareAPI('generateImage', async (id:string, value:string, negValue:string = '') => {
+            declareAPI('generateImage', async (id:string, value:string, negValue:string = '', optionsStr:string = '') => {
                 if(!ScriptingLowLevelIds.has(id)){
                     return
                 }
-                const gen = await generateAIImage(value, char as character, negValue, 'inlay')
+                const options = parseLuaOptions(optionsStr) as NAIImageGenOptions
+                const gen = await generateAIImage(value, char as character, negValue, 'inlay', options)
                 if(!gen){
                     return 'Error: Image generation failed'
                 }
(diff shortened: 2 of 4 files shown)


cubicj (commented):
I went through the options path with a small Node repro before anything else, since a JSON boundary fed directly by modules is where this gets exercised hardest. The compatibility claims hold up. With the options argument absent the built request is structurally identical to main, the char_captions shape with centers matches what novelai-python and the web client send, leaving per-character negatives unconverted matches the base negative handling, and syncing use_coords into both parameters and v4_prompt is what the reference SDK does too. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

What I am worried about is the options boundary. parseLuaOptions guarantees an object and nothing more, so valid JSON with a wrong-typed field gets past the documented fallback. `{"characters":"abc"}` throws at `naiCharacters.some`, and `{"characters":[{"prompt":1}]}` throws inside the emphasis conversion at `replaceAll`. Both rejections escape the Lua call into the mode runner's catch, which only logs to the console, so the whole script run dies with nothing shown to the user. That lands worse than the legacy single-prompt behavior, where a failed generation returned 'Error: Image generation failed' into the script and the script kept running. There are quieter shapes too. `{"use_coords":"false"}` is truthy, so it enables coordinate mode and serializes the string into both use_coords fields, and a string x or y falls back to 0.5 without counting toward the auto-enable, so coordinate placement silently disappears. I will leave the shape of the fix to you, but as written the fallback promise in the body only holds for input that fails to parse.

On the position fields, the reference implementations read them differently from the body. novelai-python constrains x and y to the 0 to 0.9 range where 0 means auto placement, and the web UI's position picker is a 5x5 grid over 0.1 to 0.9, so (0,0) is auto rather than top-left and (1,1) is a value the official client never produces. The code passes coordinates through without clamping, so a module following the body's description would send that.

Two things I could not settle from documentation, since I have no API key to test against. The SDK keeps use_order true even when coordinates are enabled, while this sends use_order as the negation of use_coords. And both the web client and the SDK send parameters.characterPrompts alongside v4_prompt, while this sends the v4_prompt captions alone. I could not find another client that does multi-character with v4_prompt only, so your live test is the best evidence available here. Could you check how these two behaved in your tests when you get a chance?

miladiashe:
@cubicj 

Thanks for the thorough pass — the repro was spot on, and I've confirmed every case you raised.

## Error containment

You're right that this landed worse than the legacy behavior. Some context I found while
fixing it: it isn't specific to the options path. `LLMMain` does `JSON.parse(promptStr)`
with no try/catch at all, so a malformed prompt array dies exactly the same way — the
`catch (error) { console.error(error) }` in the mode runner swallows all of it.

So I went with containment rather than a validation layer, to stay consistent with the
rest of the Lua bridge. `generateImage` is now wrapped, and any throw returns
`'Error: Image generation failed'` — the same contract a failed generation always had —
with the script still running.

`use_coords` needed a separate fix since a wrong type there doesn't throw. `"false"` is
truthy, so it enabled coordinate mode and serialized the string into both fields. It now
only accepts an actual boolean and otherwise falls back to auto-detection, matching how
the other `parseLuaOptions` consumers compare against `true` instead of validating up front.

## Positions

You were right and the PR body was wrong — my mistake. Confirmed against novelai-python:

```python
x: Union[float, int] = Field(0, le=0.9, ge=0)
AUTO = CharacterPosition(x=0, y=0)
```

Centers are now clamped to 0–0.9, and the fallback for a missing axis is `0` (auto)
rather than `0.5`. `0.5` isn't "unspecified" — it's the center cell of the 5×5 grid, so
supplying only `x` was silently pinning the character to the middle row. Auto-enable now
triggers on a non-auto center instead of on "is a number". PR body updated to match.

## use_order

Checked four clients; none of them ever turn it off on the positive prompt:

| client | value |
|---|---|
| novelai-python | `Field(True)` — only `use_coords` is synced into `v4_prompt` |
| Aedial/novelai-api | `"use_order": True`, hardcoded |
| SillyTavern | `use_order: true` |
| ComfyUI_RS_NAI_API_Request | `use_order=True` default |

They also don't appear to be mutually exclusive. The docs say the AI follows the order the
character prompts are listed in (first prompt → left), and then that positions are "a
nudge" which "should not conflict with the order of your character prompt boxes" — that
only makes sense if both signals stay live. `!use_coords` was switching the ordering
signal off exactly when coordinates were on, the opposite of reinforcing it. Changed to
`true`, which is also what this file did before the PR.

## characterPrompts

Multi-character did work in my live tests with `v4_prompt` alone. But since
novelai-python, SillyTavern and the web client all send `parameters.characterPrompts`
as well, I've switched to sending both — no reason to be the only client doing it
differently. For what it's worth, `v4_negative_prompt` staying at `caption` + `legacy_uc`
does match novelai-python, so I left that as is.

## Tests

Your repro cases are in the suite now — `src/ts/process/tests/naiCharacterPrompts.test.ts`.
The field building moved into `naiCharacterPrompts.ts` so it can be asserted without
pulling in the app's stores. `pnpm check` / `test` / `build` are clean, and with no options
supplied the request is unchanged from main.

cubicj (approved):
I went through 9a8561e4f against the three points from the last round. The wrong-typed shapes from my repro now stay inside the script, the handler catch returns 'Error: Image generation failed' with the run still alive, and `{"use_coords":"false"}` falls back to auto-detection instead of enabling coordinate mode. Centers match novelai-python now, 0 is auto, values clamp to 0 to 0.9, and the auto-enable keys off a non-auto center, so a string coordinate no longer flips coordinate mode on. The client survey settles the two payload questions for me, keeping `use_order` true and sending `characterPrompts` alongside `v4_prompt` is what every reference I checked does too.

With no options the request is still key-identical to main. `characterPrompts` is undefined in that case and every `globalFetch` transport serializes the body through `JSON.stringify`, which drops the key, the same way the existing `sm` and `image` fields already work. The branch also merges clean against current main, including the Lua engine cache changes on `scriptings.ts`. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head with the ten new tests.

PR #1509, opened 2026-06-27
feat: redesign model settings tab

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions? (Updated the `ModelList` prop type for the new full-width mode.)
- [ ] If your PR uses models[^1], check the following: Not applicable; this reorganizes the model settings UI, but does not change prompting, request handling, or model responses.
       - The scope is limited to the model settings UI and related labels.

## Summary

This redesigns the Bot Settings model tab so each model role has its own grouped section. The main model, sub model, and task-specific auxiliary models now share the same layout pattern, which makes model-related API keys and provider options easier to find next to the model they belong to.

## Related Issues

None.

## Changes

- Reworked the model settings area into reusable model role sections for the main model, sub model, and auxiliary model roles.
- Moved task-specific auxiliary model selection into Bot Settings, so the separate memory/translation/emotion/other model controls live with the rest of the model setup.
- Removed the old advanced toggle that controlled whether auxiliary model settings appeared under Model Settings.
- Added a full-width display mode to `ModelList` so model selectors fit cleanly inside the new grouped layout.
- Updated the related language strings to describe task-specific auxiliary models more clearly.

## Impact

This should make the settings page easier to scan without changing how chat requests are built. Existing saved values are reused, so users keep their current main, sub, and auxiliary model selections. The main visible change is that auxiliary model controls now live directly in Bot Settings instead of being split between settings pages.

## Additional Notes

Validated with:

- `pnpm check`

Still in testing...


SameDesu123:
There is an issue where ModelGrid values for the same provider share a single global value. This can be problematic, especially when using task-specific auxiliary models.

However, since addressing this would require a lot of changes, I won’t touch it for now. Maybe later.

SameDesu123:
| Looks like this |
|-----------------|
| gbaklktrnsertewt |

SameDesu123:
Also, I will change parameters options to match this in the future.

SameDesu123:
It could significantly affect the user experience, so I think we need to approach this more carefully.

PR #1510, opened 2026-06-28
feat: apply OpenRouter attribution headers globally

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions? No public types were changed.
    - [x] Have you checked if it works normally in all models? The policy is limited by destination hostname, and non-OpenRouter hosts remain unchanged.
    - [x] Have you checked if it works normally in all web, local, and node-hosted versions? The shared request paths used by these targets are covered; packaged platform runs were not performed.

## Summary

This centralizes default OpenRouter app-attribution headers in the shared request layer.

Requests whose destination is `openrouter.ai` or one of its subdomains now receive `HTTP-Referer: https://risuai.xyz` and `X-OpenRouter-Title: RisuAI` when they pass through `globalFetch` or `fetchNative`, unless the caller already supplied attribution values.

## Related Issues

None.

## Changes

- Added a shared, URL-based OpenRouter attribution-header helper.
- Applied the helper to both `globalFetch` and `fetchNative`, covering non-streaming and streaming request paths.
- Reused the same helper in the OpenAI-compatible request builder so provider-body previews match real requests.
- Preserved caller-provided headers case-insensitively, including the legacy `X-Title` alias.
- Added focused coverage for exact hosts, subdomains, misleading hosts, invalid URLs, and custom headers.

## Impact

Built-in OpenRouter requests, plugin requests, and custom or reverse-proxy requests that target an OpenRouter host through the shared request APIs receive consistent app attribution.

Requests to other hosts remain unchanged. OpenRouter model and provider discovery currently use raw `fetch` directly, so those non-generation requests remain outside this shared policy.

## Additional Notes

Validated with:

- `pnpm exec vitest run src/ts/process/request/openAI/requests.responses.test.ts src/ts/network/openRouterHeaders.test.ts` (28 tests)
- `pnpm check`
- conflict-free merge-tree check against the latest `upstream/main`

### DIFF
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -43,7 +43,7 @@ import type { AccountStorage } from "./storage/accountStorage";
 import { getColdStorageItem, makeColdData } from "./process/coldstorage.svelte";
 import { isTauri, isNodeServer } from "./platform";
 import { isLocalNetworkUrl } from "./network/localNetwork";
-import { withOpenRouterAttributionHeaders } from "./network/openRouterHeaders";
+import { isOpenRouterUrl, withOpenRouterAttributionHeaders } from "./network/openRouterHeaders";
 import { decodeProxyJobWsChunk, formatProxyStreamErrorMessage, parseProxyJobWsEvent } from "./network/proxyJobWs";
 import { getNodeServerProxyAuth } from "./storage/nodeStorage";
 
@@ -681,9 +681,11 @@ export function addFetchLog(arg: {
  */
 export async function globalFetch(url: string, arg: GlobalFetchArgs = {}): Promise<GlobalFetchResult> {
     try {
-        arg = {
-            ...arg,
-            headers: withOpenRouterAttributionHeaders(url, arg.headers),
+        if(isOpenRouterUrl(url)){
+            const headers = withOpenRouterAttributionHeaders(url, arg.headers)
+            if(headers !== arg.headers){
+                arg = { ...arg, headers }
+            }
         }
         const db = getDatabase();
         if (arg.abortSignal?.aborted) { return { ok: false, data: 'aborted', headers: {}, status: 400 }; }
@@ -1728,9 +1730,11 @@ export async function fetchNative(url: string, arg: {
     networkRoute?: 'auto' | 'local_network'
 } = {}): Promise<Response> {
 
-    arg = {
-        ...arg,
-        headers: withOpenRouterAttributionHeaders(url, arg.headers),
+    if(isOpenRouterUrl(url)){
+        const headers = withOpenRouterAttributionHeaders(url, arg.headers)
+        if(headers !== arg.headers){
+            arg = { ...arg, headers }
+        }
     }
     const useInterceptor = !!arg.interceptor
     console.log(arg.body, 'body')
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -43,6 +43,7 @@ import type { AccountStorage } from "./storage/accountStorage";
 import { getColdStorageItem, makeColdData } from "./process/coldstorage.svelte";
 import { isTauri, isNodeServer } from "./platform";
 import { isLocalNetworkUrl } from "./network/localNetwork";
+import { withOpenRouterAttributionHeaders } from "./network/openRouterHeaders";
 import { decodeProxyJobWsChunk, formatProxyStreamErrorMessage, parseProxyJobWsEvent } from "./network/proxyJobWs";
 import { getNodeServerProxyAuth } from "./storage/nodeStorage";
 
@@ -680,6 +681,10 @@ export function addFetchLog(arg: {
  */
 export async function globalFetch(url: string, arg: GlobalFetchArgs = {}): Promise<GlobalFetchResult> {
     try {
+        arg = {
+            ...arg,
+            headers: withOpenRouterAttributionHeaders(url, arg.headers),
+        }
         const db = getDatabase();
         if (arg.abortSignal?.aborted) { return { ok: false, data: 'aborted', headers: {}, status: 400 }; }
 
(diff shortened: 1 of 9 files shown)


cubicj (commented):
I traced the injection from both entry points through every transport. All send paths and the request log use the post-injection headers, re-application is idempotent, and plugin traffic is covered since risuFetch and nativeFetch map onto these two functions. pnpm check, pnpm test, and pnpm build pass on this head.

The helper only handles plain objects, but v3 nativeFetch takes options?: RequestInit. A Headers instance spreads to {}, so a plugin sending new Headers({ Authorization: ... }) to openrouter.ai silently loses its Authorization. Tuple-array headers get mangled the same way.

header::X-Title={{none}} in Additional Parameters also no longer strips the attribution, the helper refills it after the deletion. There is no way left to reach openrouter.ai without attribution. Is that intentional?

requestOpenAIResponseAPI's previewBody also misses the attribution the real request gets, the parity requests.ts got in this PR.

How do you think the helper should handle headers that arrive as a Headers instance?

SameDesu123:
Thanks for the detailed review. 

I updated the helper to accept every standard HeadersInit form and normalize Headers instances and tuple arrays into a new plain header record before adding the defaults, so Authorization and other caller headers are preserved without mutating the input. 

I also kept the existing explicit opt-out behavior: header::X-Title={{none}} now removes both title aliases case-insensitively, and a later pass through the shared fetch layer will not add the title back. 

Finally, Responses API previews now apply the same attribution policy as the real request. Regression coverage was added for all three cases. 

The full test suite passes with 195 tests passed and 3 skipped, pnpm check reports no errors or warnings, and git diff --check passes.

cubicj (commented):
Thanks for the follow-ups. I reran the three points from 7bc6b9641 and they all hold. A `Headers` instance and a tuple array both survive the helper now with `Authorization` intact and the input left unmodified, `header::X-Title={{none}}` reaches the request with neither title header set, and the Responses preview carries the same attribution the real request does. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One regression to flag. With a reverse proxy or a custom model pointed at openrouter.ai, `header::X-Title=MyApp` now puts both `X-OpenRouter-Title: RisuAI` and `X-Title: MyApp` on the request, where main sends the user value alone. The referer behaves the same way when the case differs, `header::http-referer=https://example.com` leaves both `HTTP-Referer: https://risuai.xyz` and `http-referer: https://example.com` on the request, and a duplicate referer costs the attribution since that is the header OpenRouter keys the app on. The root of the title half is the rename this PR shares with #1502, so I wrote that part up over there rather than arguing it in two places. What is specific to this PR is that the deletion path in `shared.ts` knows the two title names are the same header while the assignment path does not.

Then there are two builders in `requests.ts` the helper never reaches. `requestOpenAILegacyInstruct` applies Additional Parameters to headers that carry no attribution yet and hands them straight to `globalFetch`, so `header::X-Title={{none}}` deletes nothing there and the transport adds the defaults afterward, which leaves that path with attribution and no way off it. The Mistral branch returns its preview before the transport call, so the preview shows `Authorization` alone while the request that follows picks up attribution inside `globalFetch`, the same gap the Responses preview had at 7bc6b9641. Both are what the marker is meant to prevent, and nothing in the current code makes a builder that skips the helper fail in a visible way.

Four smaller things, none of them blocking. `toHeaderRecord` runs tuple arrays through `Object.fromEntries`, so a repeated header name loses the earlier value where `Headers` would have joined them, and the test for that case only covers a single entry tuple. The `arg` copy at the top of `globalFetch` and `fetchNative` happens for every host rather than only OpenRouter ones, so caller visible mutations like `arg.method ??= 'POST'` no longer reach the caller's object anywhere. The `{{none}}` header deletion in `shared.ts` is shared by 15 call sites across every provider, so moving it to case insensitive matching changes behavior well outside OpenRouter. And the title value here is `RisuAI` where #1502 uses `Risuai`, which is the spelling e6ee605cd standardized on, so if the headers end up in one place it is worth picking one.

How do you read the two builders? I could not tell whether the transport level injection is meant to cover them or whether they should get the same treatment the other two did. On the title and referer duplicates, let's settle the direction on #1502 and I will follow it here.

SameDesu123:
Thanks for the follow-up review. I rebased the branch onto the current main and addressed the remaining points.

Additional Parameters assignments on OpenRouter requests now replace existing title aliases and differently-cased copies of the same header, so X-Title=MyApp and a custom http-referer no longer produce duplicates. This compatibility handling is limited to headers already marked as OpenRouter headers, so deletion behavior for other providers remains unchanged.

Legacy Instruct now applies attribution before Additional Parameters, which preserves explicit opt-outs, and Mistral previews now receive the same attribution headers as the real request. Repeated tuple-array header values are normalized through the Headers API, request argument copying is limited to OpenRouter URLs, and the title spelling is now consistently Risuai.

I added regression coverage for the override, opt-out, repeated-header, Legacy Instruct, and Mistral preview paths. The focused suite passes with 46 tests, the full suite passes with 234 tests and 3 skipped, pnpm check reports no errors or warnings, and all new GitHub Actions checks pass.

cubicj (approved):
I went through 2b3bf968b against the points from fff7dd753. The assignment path now knows what the deletion path knew, so on an OpenRouter request `header::X-Title=MyApp` replaces both title aliases and sends the user value alone again, a differently-cased referer override replaces instead of duplicating, and everything outside OpenRouter keeps its exact-key semantics. Legacy Instruct applies attribution ahead of Additional Parameters so `{{none}}` opts out there too, Mistral previews now carry the same headers the real request sends, repeated tuple values survive through the Headers API, the argument copy is scoped to OpenRouter URLs, and the title reads `Risuai` throughout. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, which sits on current main.

One thing I noticed while tracing the override, not blocking. The non-streaming transports re-add an exact-case `Content-Type` default after the builder runs, so a case-variant `header::content-type` override leaves both spellings on the request there, while streaming and the preview carry the clean replacement. Non-streaming behaves as main already did, so the override just cannot fully land on that path.

PR #1513, opened 2026-06-29
fix: avoid duplicate module creation

cubicj:
## PR Checklist

- Required Checks

## Summary

Fixes module creation adding the same new module twice.

The module settings page already adds a new module to `DBState.db.modules` when the plus button is clicked, so closing the settings page, switching tabs, or using hotkeys while editing still keeps the newly created module. The create-mode footer button also pushed the same `tempModule` again, which could leave two copies after pressing it.

This PR keeps the initial create behavior and changes the footer button into a close action instead of a second create action.

## Related Issues

None.

## Changes

**`src/lib/Setting/Pages/Module/ModuleSettings.svelte`**
- Removed the second `DBState.db.modules.push(tempModule)` from the create-mode footer button
- Changed the create-mode footer button label from `Create Module` to `Close`

**`src/lang/en.ts`**
- Added the English `close` label

**`src/lang/ko.ts`**
- Added the Korean `close` label

## Impact

- Creating a module no longer adds a duplicate when the footer button is pressed.
- The existing behavior where a new module is added immediately after pressing the plus button is preserved.
- Exiting module creation through settings close, hotkeys, or tab changes still keeps the newly created module.
- Other locales fall back to the English `Close` label through the existing language merge behavior.
- No model behavior, request handling, storage format, database schema, or platform-specific logic is changed.

## Additional Notes

Validated with:

- `pnpm check`
- `pnpm test`
- `pnpm build`
- `git diff --check`

### DIFF
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -959,6 +959,7 @@ export const languageEnglish = {
     account: "Account",
     remove: "Remove",
     able: "Able",
+    close: "Close",
     assetWidth: "Asset Images Max Width",
     animationSpeed: "Animation Speed",
     screenshot: "Screenshot",
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -944,6 +944,7 @@ export const languageEnglish = {
     account: "Account",
     remove: "Remove",
     able: "Able",
+    close: "Close",
     assetWidth: "Asset Images Max Width",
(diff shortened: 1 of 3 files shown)


sub-hub (approved):
I confirmed that two are created when generating modules.
Thanks for the PR!

PR #1516, opened 2026-07-01
fix: preserve reasoning effort options with Responses API

tasoo-oos:
## PR Checklist

- Required Checks
    - [x] Type definitions were not needed
    - [x] Have you checked that it will not break any existing features?
- [x] If your PR uses models, check the following:
    - [x] Request body generation is covered for supported Responses reasoning markers
    - [x] This only changes OpenAI Responses API request construction and is platform-independent

## Summary

Preserve OpenAI Responses API reasoning capability markers when building request parameters.

## Related Issues

https://arca.live/b/characterai/175449760

image

For some models that supports `none` reasoning effort, mapping is not getting applied correctly in response API.
It resulted in the error above.

Similar bug happened for `xhigh` reasoning effort, result in downgrade to `high` effort.

## Changes

- Keep `reasoning_effort_none`, `reasoning_effort_min_medium`, and `reasoning_effort_xhigh` markers when mapping Responses API parameters.
- Add regression tests for `none`, min-medium, and `xhigh` reasoning effort mapping.

## Impact

Responses API requests now send the selected supported reasoning effort instead of falling back to default mappings such as `minimal` or `high`.

### DIFF
--- a/src/ts/process/request/openAI/requests.responses.test.ts
+++ b/src/ts/process/request/openAI/requests.responses.test.ts
@@ -190,6 +190,7 @@ describe('OpenAI Responses API helpers', () => {
         mocks.db.nanogptProvider = ''
         mocks.db.nanogptRequestModel = 'nanogpt-model'
         mocks.db.nanogptUseSubscriptionEndpoint = false
+        mocks.db.reasoningEffort = 2
         mocks.db.simplifiedToolUse = false
         mocks.db.autofillRequestUrl = false
     })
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I traced why only the Responses path lost the markers. The chat completions path hands modelInfo.parameters to applyParameters whole, while buildResponsesBody filtered through a whitelist that predates the capability markers, so the (Response API) model variants, which inherit the same parameters array, mapped -1 to minimal and 3 to high. Widening the whitelist is the right shape of fix, and the flipped filter direction changes nothing since applyParameters reads the markers with includes and skips them during iteration, so they never leak into the request body.

The separate-parameters branch computes its effort flags from the same list, so per-mode and per-model overrides get fixed by this too. pnpm check, pnpm test, and pnpm build pass on this head, with the three new regression tests pinning exactly the mappings that were broken.

Approving - thanks for the fix!

sub-hub (approved):
The change looks correct against the current Responses API spec, `none`, `medium`, and `xhigh` are all valid `reasoning.effort` values for the affected models. Your regression tests for none/min-medium/xhigh mapping are solid, and the fix also brings the Responses path in line with the Chat Completions and Anthropic paths.

Before merging, I made one small refactor: `buildResponsesBody` now reuses `isReasoningCapabilityParameter` from `shared.ts` instead of hardcoding the capability marker list, so future markers won't need to be added in two places.

Thanks for the PR.

PR #1517, opened 2026-07-02
chore: migrate pnpm config to v11

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you checked that it won’t break any existing features?

## Summary

Migrate pnpm configuration to the pnpm v11 layout and pin project tooling to pnpm 11.9.0.

## Related Issues

None.

## Changes

- Move pnpm settings from `package.json` and `.npmrc` to workspace configuration files.
- Pin pnpm 11.9.0 across the root and Hono workspaces, Docker, release CI, and self-host scripts.
- Let the PR check workflow use the root `packageManager` version instead of a conflicting explicit pnpm 10 pin.
- Require Node.js 22.13 or newer for pnpm 11.9.0 compatibility.
- Set `minimumReleaseAge` to 4320 minutes (3 days).
- Preserve build-script behavior by allowing `core-js` and keeping `sharp` blocked.

## Impact

Node.js 20 and 22.12 are no longer supported for pnpm-based development and self-host setup. Newly published dependency versions must be at least 3 days old unless explicitly excluded. Application runtime behavior is otherwise unchanged.

## Additional Notes

Validated with:

- `pnpm install --lockfile-only --frozen-lockfile`
- `pnpm --dir server/hono install --lockfile-only --frozen-lockfile`
- `svelte-check --tsconfig ./tsconfig.json` (0 errors and 0 warnings)
- `vitest run` (228 passed, 3 skipped)
- `git diff --check`


cubicj (commented):
I ran the migration locally at b80323fc0, with a pnpm 10.34.1 install on main as the baseline. Under 11.9.0, `pnpm install --lockfile-only` leaves both lockfiles byte-identical and `lockfileVersion` stays at `'9.0'`, frozen installs work in root and `server/hono`, and `pnpm check`, `pnpm test`, and `pnpm build` pass on this head. Worth noting that 10.34.1 on main already warns that the `pnpm` field in `package.json` is no longer read, so the `sharp` build block currently survives only through the settings section of the lockfile.

The one that actually breaks CI is `.github/workflows/pr-check.yml`. It landed on main after this branch's base, so the PR never saw it, and it still pins pnpm `10.34.1` through `pnpm/action-setup@v4`. That action throws "Multiple versions of pnpm specified" when its `version` input disagrees with `packageManager`, so once this merges, every PR check dies at the install step until that file gets the same bump.

The other thing is Node 20. `engines.node` still allows `^20.19.0`, but pnpm 11.9.0 wants Node 22.13 or newer. I reproduced it on Node 20.19.0, the install dies on the missing `node:sqlite` builtin, through corepack as well, and `server.sh` / `server.bat` users on Node 20 hit the same wall. Node 20 is EOL so I think raising `engines` is fine, but as written the PR changes the supported range without saying so.

Two smaller notes. pnpm 11 turns on `minimumReleaseAge` 1440 by default, which frozen installs pass, but same-day dependency bumps get refused for a day unless the setting is declared. And `core-js` ran its postinstall under 10.34.1 while `allowBuilds` now sets it to `false`. That one seems like a harmless tightening, just noting it since the PR describes the policy as preserved.

Could you check the `pr-check.yml` pin when you get to the rebase? On `engines.node` and `minimumReleaseAge` I don't have a strong preference, I'd just like to hear which way you want to take them.

SameDesu123:
Thanks for the thorough review. I rebased the branch onto the current `main` and addressed the points you raised:

- Removed the explicit pnpm 10 version from `pr-check.yml`, so `pnpm/action-setup` now follows the root `packageManager` pin (`pnpm@11.9.0`).
- Raised the Node.js requirement to `>=22.13.0` in both the root and Hono manifests.
- Explicitly set `minimumReleaseAge: 4320` (3 days).
  - Default value is 1440 (1 day), but usually these security attacks go far, so I think we need to set it to 4320.
- Restored the previous `core-js` behavior with `core-js: true`, while keeping the existing `sharp` build block as `sharp: false`.
- Added the pnpm 11.9.0 pin to the standalone Hono workspace as well.

Both lockfiles pass frozen pnpm 11.9.0 validation and the 3-day supply-chain policy. Svelte check reports 0 errors and 0 warnings, Vitest reports 228 passed and 3 skipped, and `git diff --check` passes.

cubicj (commented):
Thanks for the follow-ups. I reran the migration at 6b81da1e5. Frozen installs pass in root and server/hono under 11.9.0 with both lockfiles untouched, the `allowBuilds` map behaves as declared in a real install, `core-js` runs its postinstall again while `sharp` stays blocked, and the 4320 policy check passes against the current lockfile. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head. Since the PR check workflow runs from the PR's own branch, the green run on this PR also confirms `pnpm/action-setup@v4` resolves 11.9.0 from `packageManager` without the version input. I also simulated the Docker runtime stage locally, and a prod-only install plus `pnpm runserver` under `NODE_ENV=production` goes straight into the script without pnpm 11's verify-deps default forcing a reinstall.

Two leftovers. README.md and AGENTS.md still document Node.js 20.19+ or 22.12+, and with `engines` now at `>=22.13.0` an install following that guidance fails hard. And `server/hono/pnpm-workspace.yaml` makes the Hono workspace its own workspace root, so it does not inherit the root `minimumReleaseAge` and falls back to pnpm 11's default of 1440 there, which means the 3-day policy from the PR body currently holds for the root workspace only.

How do you want to take these two?

SameDesu123:
I fixed those issues now. Thanks for keeping review.

cubicj (approved):
I went through 2692fc82a against the two leftovers. README and AGENTS now say Node.js 22.13+, matching the engines fields, and the Hono workspace got its own minimumReleaseAge: 4320 entry. pnpm config get reports 4320 in both roots, a frozen install in server/hono passes a fresh supply-chain verification under the new policy, and both lockfiles are still byte-identical to main. The delta since 6b81da1e5 is the two doc lines and the one workspace entry, so the check, test, and build results from that head carry over.

One correction to my last comment. I wrote that Hono falls back to a 1440 default, but pnpm 11 has no default there at all. With the key absent the resolver skips the age check entirely, so the Hono workspace had no age gate rather than a 1-day one. The explicit 4320 was even more necessary than I made it sound.

SameDesu123:
Nightly build actions is added, so I will update this for it soon

PR #1518, opened 2026-07-03
feat: add bot icon thumbnails

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Add low-resolution thumbnails for bot icons shown in the sidebar while keeping the original images unchanged for full-size use.

## Related Issues

None.

## Changes

- Generate content-addressed thumbnails in the existing flat `assets/` directory.
- Encode thumbnail data as WebP when available, while using `.png` filenames so the existing asset backup paths continue to include them without changing backup selection behavior.
- Load and generate sidebar thumbnails lazily, track their source image, and ignore stale asynchronous writes after a character or icon changes.
- Preserve referenced thumbnail files during asset cleanup and keep thumbnail metadata synchronized when asset references are replaced.
- Add focused coverage for thumbnail selection, generation, fallback behavior, source races, group icons, removal, and persistence failures.

## Impact

The sidebar can display smaller image assets instead of repeatedly decoding full-size bot icons. Missing or stale thumbnails fall back to the original image, and the primary icon update remains intact if thumbnail generation or persistence fails. Existing local and Drive backup asset enumeration remains unchanged. This does not change model or prompt behavior.

## Tests

- `pnpm exec vitest run src/ts/characters.test.ts src/ts/process/coldstorageData.test.ts`
- `pnpm check`
- `git diff --check upstream/main`

## Additional Notes

Thumbnails use a maximum dimension of 192px with WebP quality 0.8. Images smaller than that are not upscaled. If WebP encoding is unavailable, thumbnail generation falls back to PNG data.


sub-hub:
Great work! So, it automatically displays compressed image thumbnails in the sidebar. I think this is a great optimization, especially when dealing with a large number of bots.

However, there is one issue that concerns me.

Up until now, all assets have been stored under the `assets/` directory; adding another subdirectory named `thumbnail` in this PR seems like a significant change.
```js
# bootstrap.ts
async function cleanChunks(options:{
    cleanColdStorage?: boolean
} = {}) {
...
    const uncleanable = new Set(await getUncleanables(db))
    if (isTauri) {

        // This line, readDir is not recursive
        const assets = await readDir('assets', { baseDir: BaseDirectory.AppData })

        console.log(assets)
        for (const asset of assets) {
            try {
                const n = getBasename(asset.name)
                if (!uncleanable.has(n)) {
                    await remove('assets/' + asset.name, { baseDir: BaseDirectory.AppData })
                }
            } catch (error) {
                console.log('error', asset.name)
            }
        }
```
For instance, since Tauri only reads files from the `assets` folder (as mentioned above), the assets inside the `thumbnail` folder would never be processed or included.
Aside from the code mentioned above, if I search for `readDir('assets`, I find that files are being read in five different locations—including files related to backups. I am wondering if this poses any issues.

Aside from that issue, I’m also wondering if it’s okay to frequently reset the thumbnails like this:
```js
char.imageThumbnail = ''
char.imageThumbnailVersion = undefined
DBState.db.characters[charIndex] = char
```

SameDesu123:
@sub-hub Thanks for clarifying, and no worries at all — your suggestion was helpful. I reconsidered the storage layout and changed the implementation to store content-addressed thumbnails directly in the existing flat `assets/` directory. The `thumbnail/` prefix and the recursive cleanup and path-preserving backup/restore support that existed only for that subdirectory are now gone.

I kept the source-image tracking, stale asynchronous write protection, lazy sidebar loading, and shared thumbnail update path because those changes protect the feature itself. For backups, I did not simply remove the `.png` check across every stored key; local and Drive backups now select only assets referenced by the database, which includes the WebP thumbnails without pulling in unrelated storage entries. Unavailable references are reported and skipped instead of aborting the rest of the backup.

The updated implementation is in `65355535`, with the focused thumbnail and backup tests, `pnpm check`, and `git diff --check` passing. Thanks again for suggesting the simpler layout.

G4tsby (commented):
Took a close look at the reworked version — sharing what I verified in case it helps the review.

The storage-layout rework holds up well. The new backup entry names carry the `assets/` prefix, but `LocalWriter.writeBackup` normalizes names with `getBasename()` and restore writes `'assets/' + name`, so the round-trip stays compatible with existing backups — I verified this directly. Dropping the old `.png`-only filter is also a real improvement beyond the PR's scope description: webp/jpeg assets were previously never included in web or Drive backups at all, and the reference-based list fixes that. The race guards (chaId + image source matching, stale-write rejection), the `characterImageThumbnailVersion` gate, the `toBlob` webp→png fallback, and the fallback-to-original behavior all look solid, and orphaned thumbnails get collected by the existing cleanup. Pushing both the live stub and the cold character in `getUncleanables` (instead of replacing) is also the right call now that thumbnails live on the stub.

A few observations, the first one being the reason I'd hold off on merging as-is:

1. A cold storage read failure now fails silently, and `cleanChunks` can permanently delete a cold character's assets

    The `coldData.character` → `coldData?.character` change in `getUncleanables` is the issue. `getColdStorageItem` returns null on read failure or corruption on every platform, and the live cold stub only keeps `image` — emotion images and other asset references exist solely in the cold-side full character.

    - Before: a null cold read threw a TypeError inside `getUncleanables`, which aborted `cleanChunks` — accidental, but fail-safe: nothing got deleted.
    - After: the failure is swallowed, the cold character's asset references are missing from the uncleanable set, and `cleanChunks` deletes those files with no warning. If the cold file is corrupted rather than transiently unreadable, this reproduces on every cleanup run.
    - The same null also silently shrinks the asset list for `SaveLocalBackup` / `backupDrive`. The #1484 confirmation gate does fire for the missing cold payload, but what the user consents to there is "cold data missing" — not "the asset files that cold data references are also excluded". Under the old full-directory enumeration those files still made it into the backup.

    Suggestion: when `cha.coldstorage` is set but the cold read returns null, either throw (restoring the old fail-safe behavior) or surface the failed key so callers can decide — abort cleanup in `cleanChunks`, and route it into the missing-assets / confirmation flow for backups. A test for this path would be valuable too, since it's the riskiest interaction in the PR.

2. On account storage, `assetExists` is a full download rather than an existence check

    The non-Tauri branch uses `forageStorage.getItem(data)`, which fetches the entire asset from the server on account storage. Since `ensureCharacterSidebarImageThumbnail` (existence check) and `getCharacterSidebarImage` → `getCharImage` (actual load) run back to back, this can mean two fetches per avatar. A cached key list or a lighter existence probe might be worth considering.

3. `SavePartialLocalBackup` re-reads cold items it already has

    `collectColdStorageBackupPayloads` at the top of the function has already read every cold item, but the cold-thumbnail collection calls `getColdStorageItem` again per character. On account storage each of those is a server round-trip, so reusing the collected payloads would be cheaper.

4. Group thumbnails are generated but never displayed

    `makeGroupImage` now generates and persists a thumbnail for groups, but only the `char.type === 'normal'` tiles and group members were switched to the lazy path — the group's own tile still uses the original image. If that's intentional, fine; otherwise it's dead data that still gets backed up and synced.

5. Test gap on the riskiest path

    `characters.test.ts` covers the thumbnail logic well, but there's no coverage for the `getUncleanables` cold-null path or for what the reference-based backup list does when a cold read fails — which is exactly where point 1 lives.

The thumbnail feature itself (generation, race handling, lazy loading) looks well built, and the `.png` filter removal is a nice hidden fix. Point 1 is the one thing I'd consider a blocker, since it turns a previously fail-safe path into a silent permanent-deletion path.

SameDesu123:
@G4tsby Thanks for the detailed review. I went back through the cold-storage paths after the rebase and found that the conflict resolution had brought back the older behavior where the stored character replaced the live cold stub during asset collection. That meant a thumbnail generated on the stub after the character was moved to cold storage could be left out, while a failed cold read could still hide emotion images and other references that only exist in the stored character. The asset scan now keeps both sides: the live stub preserves its current profile image and thumbnail, and the cold payload contributes the rest of the character assets. If the payload is missing or belongs to a different character, automatic cleanup stops before it removes anything.

I also changed local, partial, and Drive backups to reuse the same cold-storage payload snapshot that was already read and validated for the confirmation step. This removes the second account-storage read in partial backups and prevents a transient second failure from silently changing the asset list after confirmation. If the user explicitly continues with an unavailable cold item, the live stub's profile image and thumbnail are still included. I added regression coverage for the live-stub plus cold-character combination, missing and mismatched payloads, and snapshot-based partial backup collection; the focused suite now passes 33 tests, along with `pnpm check` and `git diff --check`.

I rechecked the group-thumbnail observation as well. The sidebar's local `type: "normal"` value is a presentation type assigned to every string entry in `characterOrder`, including database group characters, so the group's own tile already goes through the thumbnail-aware path. I left `assetExists` unchanged in this pass since that is a separate account-storage performance optimization rather than part of the data-safety fix. The follow-up is in `e61ff0d6` and `9017bd6f`.

PR #1520, opened 2026-07-05
feat: add cbs parser API for Plugin V3

seto-sama:
## PR Checklist

- Required Checks

## Summary

Add parseRisuChat API for parsing CBS text to Plugin V3 API

## Related Issues

None

## Changes

Receives text containing {{cbs}} from within the plugin, and returns the text with {{cbs}} parsing completed.  
It can be called in the form of risuai.parseRisuChat(text, options), and by passing characterIndex, chatIndex, messageIndex, etc., it minimizes errors during parsing.

By using the processRegex option, you can also apply new regex scripts (Request Modification).

## Impact

No modifications to existing APIs, no model calls, only returning CBS-parsed text.

## Additional Notes

None

### DIFF
--- a/src/ts/plugins/apiV3/risuai.d.ts
+++ b/src/ts/plugins/apiV3/risuai.d.ts
@@ -1299,6 +1299,45 @@ interface RisuaiPluginAPI {
      * @returns Chat object or null if not found
      */
     getChatFromIndex(characterIndex: number, chatIndex: number): Promise<any|null>;
+
+    /**
+     * Parses Risu CBS macros in text with the selected/current character context.
+     * Set processRegex to also apply editprocess regex scripts after CBS parsing.
+     * @param text - Text containing CBS macros
+     * @param options - Parser context options
+     * @returns Parsed text
+     *
+     * @example
+     * ```typescript
+     * const text = await risuai.parseRisuChat('Hello, {{user}}', {
+     *   role: 'user',
+     *   processRegex: true,
+     * });
+     * ```
+     */
+    parseRisuChat(text: string, options?: {
+        /** Character index to use. Defaults to the selected character. */
+        characterIndex?: number;
+        /** Chat index to use as the current chat context. Defaults to the character's current chat. */
+        chatIndex?: number;
+        /** Message index used by chat-aware CBS macros. Defaults to -1. */
+        messageIndex?: number;
+        /** Message role used by role-aware CBS macros. */
+        role?: string;
+        /** Apply editprocess regex scripts after CBS parsing. */
+        processRegex?: boolean;
+        /** Enable chat variable writes from CBS macros. */
+        runVar?: boolean;
+        /** Remove unresolved variable macros. */
+        rmVar?: boolean;
+        /** Use tokenizer-accurate parsing behavior where supported. */
+        tokenizeAccurate?: boolean;
+        /** Additional CBS condition flags. */
+        cbsConditions?: {
+            firstmsg?: boolean;
+            chatRole?: string;
+        };
+    }): Promise<string>;
     
 
     /**
--- a/src/ts/plugins/apiV3/risuai.d.ts
+++ b/src/ts/plugins/apiV3/risuai.d.ts
@@ -1301,8 +1301,10 @@ interface RisuaiPluginAPI {
     getChatFromIndex(characterIndex: number, chatIndex: number): Promise<any|null>;
 
     /**
-     * Parses Risu CBS macros in text with the selected/current character context.
-     * Set processRegex to also apply editprocess regex scripts after CBS parsing.
+     * Parses Risu CBS macros using the currently selected character and active chat.
+     * Set processRegex to also run the editprocess script pipeline after CBS parsing.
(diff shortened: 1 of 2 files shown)


SameDesu123:
Thanks for adding this API. Exposing CBS parsing to Plugin V3 would be useful, but I found a context-handling issue that should be resolved before merging.

`parseRisuChat` builds `parserChar` from `characterIndex` and `chatIndex`, but many chat-aware CBS implementations do not read that value. They still resolve their data through the global `selectedCharID` and the selected character's current `chatPage`. This includes `{{char}}`, `{{history}}`, `{{lastmessage}}`, message-time macros, and the chat-variable helpers used by `{{getvar}}` / `{{setvar}}`.

For example, if character A is currently selected and a plugin calls:

```ts
await risuai.parseRisuChat('{{setvar::status::ready}} {{history}}', {
  characterIndex: characterBIndex,
  chatIndex: 1,
  runVar: true,
});
```

the history can come from character A's current chat, and `status` can also be written to character A's current chat instead of character B's chat 1. This makes the new index options especially risky because the call appears explicitly scoped while it can read or modify a different conversation.

Could you please pass an explicit character/chat context through the CBS matchers and chat-variable helpers, or initially limit this API to the currently selected chat and remove the unsupported index options? Temporarily changing `selectedCharID` during parsing would probably not be safe because parsing calls can overlap and the store is reactive.

There are two related cases worth addressing as part of that contract:

- An invalid `characterIndex` silently falls back to the selected character, and an invalid `chatIndex` falls back to that character's current chat. Explicit invalid indices should produce a clear error rather than parse a different conversation.
- `processRegex` calls `processScriptFull`, which does more than apply regex replacements: it invokes Plugin V2 `editprocess` hooks, performs another CBS pass without `parserChar`, and action scripts such as `@@inject` can modify the currently selected message. If `processRegex` is intended to be a pure transformation option, it should use a narrower path; otherwise the API documentation should describe these side effects and the operations must honor the requested context.

Tests with character A selected while parsing character B—including `{{history}}`, `runVar`, invalid indices, and `processRegex`—would make the intended behavior much clearer. The PR passes `pnpm check`; these are runtime correctness issues rather than type errors.

seto-sama:
Thanks for the detailed review. I removed characterIndex and chatIndex and restricted parseRisuChat to the currently selected character and active chat.  
The API now throws when there is no active character/chat or when messageIndex is invalid. I kept processRegex because applying request data regex scripts is part of the intended use case, and updated the documentation to clarify that it runs the full editprocess pipeline and may invoke plugin handlers or action scripts.

I also manually tested CBS parsing while switching between characters, including {{char}}, {{history}}, {{lastmessage}}, messageIndex, invalid messageIndex, and runVar/getvar.

cubicj (commented):
I reviewed the head after the follow-up commit. The restriction to the selected character and active chat resolves the context mismatch SameDesu123 raised, and the new error paths reach the plugin as rejected promises through the sandbox bridge. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One behavior I want to confirm first. With processRegex the text is CBS-parsed twice, by parseRisuChat and again inside processScriptFull. Raw-value macros expose this. If a variable value contains {{char}}, {{getvar::x}} returns the expanded name under processRegex, while the request pipeline parses once and sends the literal braces. No variable writes happen in the second pass, but the output differs from the request path for the same text, and plugins will depend on whichever depth ships. Is the extra expansion depth intended as part of the processRegex contract?

seto-sama:
I tested this several times on the current head but couldn’t reproduce the single-pass behavior. With x containing the literal {{char}}, sending {{getvar::x}} through the normal chat flow also expanded it to the character name, which I confirmed in the request log.

The existing request path calls risuChatParser before processScriptFull, which parses it again internally, so this behavior predates the PR and processRegex currently matches it. Changing the shared pipeline seems beyond the scope of this PR. I may be missing another request path, so please let me know if there is a different reproduction I should check. Sorry for the confusion

cubicj (approved):
I traced your reproduction through the code before anything else, and you are right. My claim that the request pipeline parses once and sends the literal braces was wrong, and I'm sorry for holding this PR on a contract question built on that claim. The message loop pre-parses each message with `risuChatParser` before handing it to `processScriptFull`, whose internal pass runs unconditionally, so the normal path was already two passes at exactly the point I compared. It is actually three. `runCurrentChatFunction` runs every message through the parser with `runVar` at the start of request building and writes the result back to the chat, which is what your request log picked up. `parseRisuChat` with `processRegex` never exceeds the pipeline's depth. For macro-shaped variable values it expands one level less and it does not write back into the chat, and I think both are the right calls for a parser API, not deltas worth documenting.

The head is unchanged since my last pass, so everything I verified there still holds. The API stays restricted to the selected character and active chat, the three error paths throw and reach plugins as rejected promises through the sandbox bridge, and the `messageIndex` guard still rejects out-of-range indices. `pnpm check`, `pnpm test`, and `pnpm build` are clean on 3470cc776, and it merges clean against current main.

sub-hub (approved):
Although various side effects are expected, I believe the plugin developer can adjust them as desired because there are various settings within the function to control those effects.
Thanks for the PR.

PR #1521, opened 2026-07-05
fix: make partial edit use the translation cache for translated messages

seto-sama:
## PR Checklist

- Required Checks

## Summary

When editing a chat message, if the message is currently translated, partial edit now searches the translation cache instead of the original text and updates the translation cache.

## Related Issues

None

## Changes

Pass the translated state from Chat.svelte to PartialEditController.svelte.

PartialEditController.svelte uses that state to match against either the original text or the translation cache.

Refresh the displayed message in Chat.svelte after the translation cache is updated.

## Impact

No impact except for editing the translation cache.

## Additional Notes

None


sub-hub:
I am the creator of this feature—thank you! I am positive about this feature being extended.

I have completed the code review. Although functions like `findAndProcessMatches` have been changed to `async`, they appear to work well since they originally had no return value.

I have also confirmed that the feature for making partial edits to translated results works correctly when using the Echo model to translate with `Ax.Model`. However, since I don't use LLM translation very often, I have only confirmed that it works well for now, and it is difficult to test this feature in various environments in the long term.

If an issue arises when not using LLM translation after this PR is merged, I might attempt to fix it, but it is difficult to focus on it if the problem occurs when using LLM translation.

So, I am curious if you intend to continue using this feature yourself.
After all, if a feature you use breaks, you will easily feel the inconvenience and be able to fix it :D

seto-sama:
Thank you for reviewing the code.  
I am using RisuAI by receiving English output and translating it into Korean via LLM, but since I don't chat that much, I think the frequency of using the feature is probably average.

If any issues arise regarding the LLM translation, I will do my best to fix them!

sub-hub (approved):
Thank you for your contribution. Although I do not frequently use LLM translation, I intend to approve this PR for the following reasons:

1. I tried using it myself and analyzed the code, and it appears to have no errors.
2. Even if an error occurs in the feature, it will not have a critical impact. Users of LLM translation are currently unable to use the partial edit feature anyway.
Furthermore, even if an error occurs during the cache modification process, the original message remains intact while only the translation cache is updated; therefore, the translation cache can be recreated at any time.
3. Since there are people who need this feature, I believe there will be active reporting if any issues arise.

Therefore, I think it would be great if this feature were added!
Thanks for the PR.

G4tsby (commented):
Verified the key design decision here: the translation cache key derives from the original message (which the translation-edit path never mutates), so an edited translation persists under the same key across re-renders and retranslation lookups — I cross-checked the key derivation against ChatBody's cache lookup and they agree. Also checked the `findAndProcessMatches` async conversion: all four call sites are fire-and-forget with the proceed callbacks running inside the async function, so there's no ordering dependency. Snapshot semantics look right too — resets null out `selectedRange`, so a stale source can never be combined with a fresh range on save.

One defensive suggestion: in `handlePartialEditSave`, an event with `target === 'translation'` but an `undefined` `translationKey` falls through into the original-overwrite branch — i.e., the edited *translation* text would replace the *original* message. It's unreachable today thanks to the `!key` guard added in 3357223d, but since the failure mode is destructive, an explicit early `return` for that combination would be cheap insurance.

cubicj (commented):
I traced the cache key end to end before looking at anything else, since the whole feature hangs on it. The key Chat.svelte writes under matches the key `translateLLM` looks up in all three translator configurations, including the CBS role condition, and the translated view reads the edited entry back on the next render. The original message and DB are never touched on the translation path. `pnpm check`, `pnpm test`, and `pnpm build` pass on this head.

What breaks is the view for manual translation users. Saving a partial translation edit bumps `ReloadChatPointer`, and that pointer is part of the message hash in `Chats.svelte`, so the whole `Chat` instance is unmounted and remounted, not just the `{#key}` block inside it. The fresh instance starts with `translated = false`, so with `autoTranslate` off the message flips back to the untranslated original right after saving. The edit is in the cache and reappears after pressing translate again, but the immediate result reads as the edit having failed. With `autoTranslate` on the remount re-enters translation and hits the cache, which is likely why it looked fine in testing. The existing whole-message translation edit avoids this because it never bumps the pointer, it remounts `ChatBody` inside the same instance and `translated` survives.

Deleting the entire translation content also does not stick. The delete path can write an empty string to the cache, and `translateLLM` checks the cached value with `if (cacheMatch)`, so an empty string reads as a miss and the next render fires a fresh LLM request that resurrects the translation.

One smaller thing from the async conversion. `matchingState.mode` is set before the await, and the match selection modal renders on mode alone, so a slow cache lookup can show an empty selection modal first, and cancelling during that window gets revived when the pending lookup resolves into the proceed callback.

Could you take a look at these points?

seto-sama:
1. Thank you for the fix and approval.

2. Thank you for the review. I added a defensive guard for events where target === 'translation' but translationKey is undefined.

3. Thank you for the detailed review. I fixed the issues you identified. Preserving the translated view during manual translation edits also required adjusting the ChatBody initialization and rendering flow, so I updated that path as well.
I manually tested partial and full edits of both translated and original messages with auto-translation enabled and disabled, including saving an empty translation. I’ve pushed the fixes in commit c2a71c29.

cubicj (commented):
Sorry, one claim in my last comment was wrong — the whole-message translation edit had the same reset on main for manual translation users, since exiting the editor remounted `ChatBody`. Your restructure keeping it mounted behind `class:hidden` fixes that path too.

I went through c2a71c29 and the three issues are all fixed. The revision counter keeps the view on the same `ChatBody` instance, deleted translations stay deleted with the `!== null` check, and the reworked match flow removes both the empty modal flash and the cancel revival. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One new issue the restructure opens up. The edit button still works while the translation editor is open, so both editors can be up at once. Editing the original updates `msgDisplay` live, while `saveTranslationEdit` recomputes its cache key at save time, so saving the translation afterwards stores the old translation under the edited original's key, and the next translated render shows it as a stale cache hit. The partial edit path is not affected, since it carries the key it captured at match time.

seto-sama:
1. Thank you again for the additional review! I updated the edit-state handling so that original-message editing and translation editing cannot be active at the same time. The translation button is also intentionally disabled while the original message is being edited to prevent accidental translation-state changes during source editing.

2. This was not a functional bug, but I noticed a minor visual inconsistency: the bottom margin of the last paragraph changed only while a partial-edit modal was open. I fixed that issue in the separate commit 7db697d. If you consider it outside the scope of this PR, that commit can be omitted.

cubicj (commented):
I went through ed1babcb and 7bd697d5. The mutual exclusion closes the dual-editor path, and capturing the key in `loadTranslationForEdit` fixes the stale-key issue, since saving no longer recomputes the key. The UI commit is worth keeping. Besides the margin fix, the new bodyRoot check stops hover from treating the whole chattext span as a block, and the relocation keeps the match modal's own elements out of detection. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One issue in the new save guard. An empty message with a real index isn't `blankMessage`. A `/send` or `/sendas` with empty text inserts one, and editing a message down to empty produces one too. On such a message the translation editor still opens, and its cache key normally comes out empty as well. `saveTranslationEdit`'s `!editTranslationKey` check treats '' the same as null, so saving silently does nothing and the editor never closes, and there is no cancel path. The guard covers the missing-key case, but the empty key falls into it as well.

There's also a smaller edge in the spinner suppression. When a translated parse throws once and the retry succeeds, the outer call's finally overwrites `lastParsed` with its own empty queue after the inner call filled it, which is pre-existing. But that finally now also records `lastRenderedRevision`, so the next rerender after a cache edit counts as preservable, skips the spinner, and shows the empty `lastParsed`. The message blanks out during that reparse instead of keeping the old content. The cache itself stays correct; only the pending frame is affected.

Could you take a look at these two points?

seto-sama:
Sorry for making you play bug whack-a-mole on this PR 😅 I've fixed both issues you pointed out.  
For the second one, I considered refactoring the recursive retry into a loop, but that would significantly expand the scope and potential impact of the change, so I went with a minimal fix that preserves the existing behavior.  
Thanks again for the incredibly thorough review!

cubicj (approved):
Thanks for the quick fixes. I went through 9b3fa1f03 and both points from the last round hold. Allowing '' as a valid key closes the soft-lock, since saving now writes the entry and closes the editor, and assigning the retry result to the outer queue fixes the blank pending frame along with the pre-existing clobber it came from. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head. The failing CI job is a main-side test mock gap that #1608 fixes, unrelated to this PR.

While checking what an empty key means once it becomes savable, I found that the LLM cache is keyed by text alone, so a translation stored under '' is shared by every message whose translation key comes out empty, including ones a display regex intentionally blanks, and with cached-only auto-translation it can surface without any user action. Entries like that already appear on main, since auto-translating an empty message sends the empty content to the model and caches whatever comes back.

The fix is mechanical, so I pushed f6c67d66 onto this branch, an empty-input guard at the top of `translateHTML` that stops both the empty-content requests and the display of '' entries. The editor path reads the cache directly, so editing an existing entry still works.

PR #1524, opened 2026-07-06
feat: add MiniMax M3 and M2.7 models

octo-patch:
## Summary

Add MiniMax as a built-in provider with MiniMax-M3 and MiniMax-M2.7 model entries for OpenAI-compatible and Anthropic-compatible requests.

## Changes

- Register the provider and its display name.
- Add both target model IDs with a shared API key setting.
- Expose global and CN endpoint variants for both OpenAI and Anthropic protocols.
- Append /v1/messages to the configured Anthropic base URL in the request adapter.
- Enable image and video input for MiniMax-M3 while keeping MiniMax-M2.7 text-only.
- Encode MiniMax-M3 image and video attachments for both protocol formats.
- Expose Off and Adaptive thinking control for MiniMax-M3 while preserving each protocol's default behavior.
- Keep request-level coverage for endpoints, payloads, modalities, and thinking modes.

## Verification

- `pnpm exec vitest run src/ts/process/request/openAI/requests.responses.test.ts`: 28 passed.
- `pnpm check`: 0 errors and 0 warnings.
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test`: 196 passed, 3 skipped.
- `git diff --check`.


src/ts/model/types.MiniMax.test.ts
tasoo-oos, line 1: This file seems redundant. Is there a particular reason for making it?

octo-patch:
Updated the PR to expose both target models through global and CN OpenAI and Anthropic endpoints. The Anthropic adapter now appends /v1/messages to the configured base URL and preserves MiniMax-M3 image and video blocks. I ran the targeted request suite, pnpm check, the full Vitest suite, and git diff --check.

octo-patch:
Thanks for the review. I removed the redundant registry test and kept the coverage in the request-level suite.

octo-patch:
Thanks for the review. I merged the latest main to resolve the merge conflict in `src/ts/model/types.ts`, keeping the MiniMax M3/M2.7 model entries, regional OpenAI and Anthropic endpoints, and adaptive thinking handling intact. I bumped the new `adaptiveThinking` flag to a free slot to avoid colliding with main, updated the request test mock to match, and ran `vitest run` (232 passed) and `svelte-check` (0 errors).

PR #1526, opened 2026-07-06
Fix empty system role replacement

tammy3211:
## PR Checklist

- Required Checks

## Summary

If `db.systemRoleReplacement` is blank, fallback to 'user'

## Related Issues

If `db.systemRoleReplacement` is blank, request sends role with empty string.

## Changes

If `db.systemRoleReplacement` is blank, fallback to 'user'

### DIFF
--- a/src/ts/process/request/request.ts
+++ b/src/ts/process/request/request.ts
@@ -368,7 +368,7 @@ export function reformater(formated:OpenAIChat[],modelInfo:LLMModel|LLMFlags[]){
         for(let i=0;i<formated.length;i++){
             if(formated[i].role === 'system'){
                 formated[i].content = db.systemContentReplacement ? db.systemContentReplacement.replace('{{slot}}', formated[i].content) : `system: ${formated[i].content}`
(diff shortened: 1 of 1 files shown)


sub-hub (approved):
src/ts/storage/database.svelte.ts
`data.systemRoleReplacement ??= 'user'`

It seemes fallback with a user is a consistent choice.
It's already been two years since systemRoleReplacement was added, but there's still no reason not to approve it.

It must have been a difficult problem to find. Thank you for your PR as well.

cubicj (approved):
Verified. The empty role breaks differently per provider, and the worst case is silent: the Anthropic converter's `switch(chat.role)` has no default case, so mid-conversation system messages are dropped from the request without any error. Falling back to `'user'` matches the existing `??= 'user'` load default and the preset apply fallback, and this line is the only place the value is consumed at request time, so the one-line guard covers the whole issue.

`pnpm check`, `pnpm test`, and `pnpm build` pass locally.

PR #1529, opened 2026-07-11
feat: show detailed request context token breakdown

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This PR makes it easier to understand why a request is using a certain number of input tokens. The request information dialog now shows how much of the context came from the prompt, character, lorebook, modules, previous messages, the current message, long-term memory, and anything that could not be classified elsewhere.

## Related Issues

None.

## Changes

Context is categorized while the request is being assembled, and that attribution is preserved when system messages are merged. The final category total is kept in sync with the existing input-token count, then stored with the message generation metadata so it can be inspected later. The internal attribution fields are removed before the request reaches the provider, so the prompt and request body themselves are not changed.

The request information dialog presents the result as localized token counts, percentages, and compact bars. Each category has its own muted CSS color, with separate light and dark palettes, and very small non-zero categories still receive a visible 2px bar. The token distribution, fallback category, and provider handoff behavior are covered by focused tests.

## Impact

The breakdown is available for newly generated messages. Existing messages remain compatible and continue to show their existing input and output totals as before. Provider request bodies and prompt content are unchanged; the only additional cost is a small amount of local token accounting and generation metadata.

## Additional Notes

I tested the request-context logic with `pnpm exec vitest run src/ts/process/requestContext.test.ts` (3 tests passed), ran `pnpm check`, and verified the final PR diff with `git diff --check upstream/main...HEAD`.

|Preview|
|-----|
| IMG_1823 |


SameDesu123:
I thought this was a good idea when I originally wrote it, but looking back at it now, I don't think this UI design is particularly reasonable.

While explicitly displaying the token count is useful from an informational perspective, I don't think it's desirable from a UI perspective.

PR #1532, opened 2026-07-12
Fix searchMatch key mutation in regex lorebook

### LINKED ISSUE #1530: Regex lorebook keys silently fail on later scanned messages (`arg.keys` mutated during scan)
## Summary

When a regex lorebook key does not match the first (oldest) scanned message, the key stored in `arg.keys` has already been mutated, and later messages are matched against a corrupted key — typically failing with a caught `SyntaxError`. In practice, regex entries fail to activate on triggers that appear in recent messages, which is the common case.

The regex matching loop mutates `arg.keys[0]` while scanning messages. After the first unsuccessful message, the stored regex key no longer contains its trailing delimiter and flags, so the next iteration parses the pattern body as flags. The resulting `SyntaxError` is caught and the function returns `false`; the lorebook entry is silently skipped, with no error logged by this branch.

The same block has three additional defects:

- the opening `/` is kept as part of the regex pattern;
- regexes with no flags, such as `/studio/`, are never tested;
- the parsed pattern is always written to `arg.keys[0]`, even while processing a later key, so with multiple regex keys the first key's original value is destroyed and replaced by a mangled copy of whichever key was processed last.

## Environment

- Checked against `main` (`src/ts/process/lorebook.svelte.ts`), 2026-07-11.

## Location

- File: `src/ts/process/lorebook.svelte.ts`
- Function: `searchMatch`
- Branch: `if (arg.regex)`

## Current behavior

```ts
if(arg.regex){
    for(const mText of mList){
        for(const regexString of arg.keys){
            if(!regexString.startsWith('/')){
                return false
            }
            const regexFlag = regexString.split('/').pop()
            if(regexFlag){
                arg.keys[0] = regexString.replace('/'+regexFlag,'')
                try {
                    const regex = new RegExp(arg.keys[0],regexFlag)
                    const d = regex.test(mText.data)
                    if(d){
                        matchLog.push({
                            prompt: mText.prompt,
                            source: mText.source,
                            activated: regexString
                        })
                        return true
                    }
                } catch (error) {
                    return false
                }
            }
        }
    }
    return false
}
```

## Root cause

This line mutates the shared key array during iteration:

```ts
arg.keys[0] = regexString.replace('/' + regexFlag, '')
```

`mList` is iterated oldest → newest, and the inner loop re-reads `arg.keys` for every message. For a key such as:

```
/(?!)|\bstudio\b/i
```

- **1st message:** `regexFlag = 'i'`, the key compiles correctly, and the message is tested. If it does not match, the loop moves on — but `arg.keys[0]` now holds the stripped string `/(?!)|\bstudio\b`.
- **2nd message:** `regexString` is the stripped string. `regexString.split('/').pop()` now returns the **pattern body** instead of the flags, and the code effectively calls `new RegExp('', '(?!)|\\bstudio\\b')`, which throws `SyntaxError: Invalid flags`.
- The exception is caught with `return false`, so the remaining messages and keys are never tested and the entry is silently skipped.

As a result, a regex lorebook entry only activates reliably if the match happens to be in the **oldest** message of the scan window (or if `searchDepth` is 1); later messages are evaluated against a corrupted key. Because the failure is intermittent — it works when the trigger happens to sit in the oldest scanned message — it is easy to misdiagnose as a configuration problem.

## Minimal reproduction

```js
function searchMatchRegex(key, messages) {
  const arg = { keys: [key] };

  for (const message of messages) {
    for (const regexString of arg.keys) {
      if (!regexString.startsWith('/')) return false;

      const regexFlag = regexString.split('/').pop();

      if (regexFlag) {
        arg.keys[0] = regexString.replace('/' + regexFlag, '');

        try {
          if (new RegExp(arg.keys[0], regexFlag).test(message)) {
            return true;
          }
        } catch (error) {
          return `CAUGHT: ${error.name}`;
        }
      }
    }
  }

  return false;
}

// The "(?!)|" prefix works around the leading-slash defect
// described below, so this example isolates the mutation bug.
const key = '/(?!)|\\bstudio\\b/i';

console.log(searchMatchRegex(key, ['the studio']));
// true

console.log(searchMatchRegex(key, ['no match', 'the studio']));
// CAUGHT: SyntaxError
```

### 1. Opening slash remains in the pattern

Removing only `'/' + regexFlag` leaves the opening `/` in place. For example:

```
/유즈|yuzu/i
```

is compiled with the pattern `/유즈|yuzu`, so the **first alternative requires a literal `/`** in the text and effectively never matches. (Prepending a non-matching `(?!)` alternative works around this defect: the leading slash then attaches to the dead alternative instead of a usable one.) With a single-alternative key such as `/\bstudio\b/i`, the whole pattern becomes `/\bstudio\b` and the key cannot match ordinary text at all — even in the first scanned message.

### 2. Regexes without flags are skipped

For `/studio/`, `regexString.split('/').pop()` returns an empty string. Since the regex is only evaluated inside `if (regexFlag)`, the key is never tested.

### 3. Multiple regex keys clobber `arg.keys[0]`

The parsed value is always written to `arg.keys[0]`, even while the loop is processing a later key. With more than one regex key, the first key's original value is destroyed after the first message and replaced by a mangled copy of the last-processed key.

## Proposed fix

Parse the pattern and flags into local variables without mutating `arg.keys`:

```diff
         if(arg.regex){
             for(const mText of mList){
                 for(const regexString of arg.keys){
-                    if(!regexString.startsWith('/')){
+                    const parsed = regexString.match(/^\/([\s\S]*)\/([dgimsuvy]*)$/)
+                    if(!parsed){
                         return false
                     }
-                    const regexFlag = regexString.split('/').pop()
-                    if(regexFlag){
-                        arg.keys[0] = regexString.replace('/'+regexFlag,'')
-                        try {
-                            const regex = new RegExp(arg.keys[0],regexFlag)
-                            const d = regex.test(mText.data)
-                            if(d){
-                                matchLog.push({
-                                    prompt: mText.prompt,
-                                    source: mText.source,
-                                    activated: regexString
-                                })
-                                return true
-                            }
-                        } catch (error) {
-                            return false
+
+                    const [, pattern, flags] = parsed
+
+                    try {
+                        const regex = new RegExp(pattern, flags)
+                        if(regex.test(mText.data)){
+                            matchLog.push({
+                                prompt: mText.prompt,
+                                source: mText.source,
+                                activated: regexString
+                            })
+                            return true
                         }
+                    } catch (error) {
+                        return false
                     }
                 }
             }
             return false
         }
```

This fixes all four defects at once: no mutation, the opening `/` is excluded from the pattern, flagless regexes are accepted (`([dgimsuvy]*)` matches an empty flag string), and each key is evaluated independently.

An equivalent alternative is to split at `lastIndexOf('/')` and let the `RegExp` constructor validate the flags, which avoids hardcoding the flag list; either approach removes the mutation.

**Notes on behavior changes:**

- Existing keys using the `(?!)|` prefix workaround will continue to work after this change, because `(?!)` is a non-matching alternative.
- Entries whose keys previously could not activate — flagless regexes, and matches in any message after the first scanned one — will begin activating. This is the intended fix.
- Keys that (likely unintentionally) relied on the retained leading slash may change behavior; the new parsing aligns with the apparent `/pattern/flags` key format.

## Expected behavior

| Case | Current behavior | Expected behavior |
| --- | --- | --- |
| Match in first scanned message | Activates in some cases (leading-slash defect can still block it) | Activates |
| Match only in a later scanned message | Caught `SyntaxError`, returns `false` | Activates |
| Flagless regex `/studio/` | Skipped | Activates |
| Multiple regex keys | `arg.keys[0]` clobbered after the first message | Each key evaluated independently |

renamerdis-wq:
## PR Checklist

- Required Checks
    - [ ] Have you added type definitions? (N/A, no new types introduced; the change only uses locals inferred from existing types)

## Summary

Fixes the `arg.keys[0]` mutation bug in regex lorebook key matching, which caused regex lorebook entries to silently fail on all messages after the first non-matching one.

## Related Issues

Fixes #1530

## Changes

- Removed the mutation of `arg.keys[0]` during regex key iteration in `searchMatch` (`src/ts/process/lorebook.svelte.ts`).
- Each key is now parsed into local `pattern` and `flags` variables using `^\/([\s\S]*)\/([dgimsuvy]*)$`.
- _**Updated**_: Keys are parsed and compiled once into a `{ regex, source }[]` array before the message scan instead of being re-parsed and re-compiled for each message; `lastIndex` is reset before each `.test()` call for `g`/`y` flagged regexes.
- This fixes all four reported defects:
  - matches in later scanned messages failing;
  - the opening `/` being retained as part of the pattern;
  - flagless regexes never being evaluated;
  - multiple regex keys overwriting `arg.keys[0]`.

## Impact

- Regex lorebook entries now activate correctly regardless of which scanned message contains the match.
- ~~Error behavior is preserved: keys that fail to parse or compile still return false, same as before.~~
  _**Updated**_: regex keys are validated up front, so any malformed key makes the entry inactive. See the verification comment below for details.
- Pure string/regex parsing change — model-agnostic, no platform-specific code, so behavior is identical across web, local, and node-hosted versions.

## Additional Notes

Manual verification on a local node-hosted build, with a test lorebook covering each reported defect:

- a match in the first scanned message (`/studio/i` — also covers the leading-slash defect);
- a match only in the last scanned message (the mutation defect);
- a flagless regex (`/flagless/`);
- multiple comma-separated regex keys where only the second key matches.

All four entries activate correctly on the fixed build (verified via the lore monitor and the actual prompt sent).

Minor edge case: an empty pattern key (`//`) previously was silently skipped and now compiles to an empty regex (matches everything). This follows the parsing approach proposed and approved in #1530.

<details>
<summary>Verification screenshots (lore monitor + actual request body)</summary>

All four test entries activate — TEST1 (match only in the last scanned message), TEST2 (`/studio/i`, the leading-slash case from the issue), TEST3 (flagless `/flagless/`), TEST4 (second of two comma-separated keys):

</details>

### DIFF
--- a/src/ts/process/lorebook.svelte.ts
+++ b/src/ts/process/lorebook.svelte.ts
@@ -143,28 +143,33 @@ export async function loadLoreBookV3Prompt(){
             }))    
 
         if(arg.regex){
+            const regexes: { regex: RegExp, source: string }[] = []
+            for(const regexString of arg.keys){
+                const parsed = regexString.match(/^\/([\s\S]*)\/([dgimsuvy]*)$/)
+                if(!parsed){
+                    return false
+                }
+                const [, pattern, flags] = parsed
+
+                try {
+                    regexes.push({
+                        regex: new RegExp(pattern, flags),
+                        source: regexString
+                    })
+                } catch (error) {
+                    return false
+                }
+            }
             for(const mText of mList){
-                for(const regexString of arg.keys){
-                    if(!regexString.startsWith('/')){
-                        return false
-                    }
-                    const regexFlag = regexString.split('/').pop()
-                    if(regexFlag){
-                        arg.keys[0] = regexString.replace('/'+regexFlag,'')
-                        try {
-                            const regex = new RegExp(arg.keys[0],regexFlag)
-                            const d = regex.test(mText.data)
-                            if(d){
-                                matchLog.push({
-                                    prompt: mText.prompt,
-                                    source: mText.source,
-                                    activated: regexString
-                                })
-                                return true
-                            }
-                        } catch (error) {
-                            return false
-                        }
+                for(const {regex, source} of regexes){
+                    regex.lastIndex = 0
+                    if(regex.test(mText.data)){
+                        matchLog.push({
+                            prompt: mText.prompt,
+                            source: mText.source,
+                            activated: source
+                        })
+                        return true
                     }
                 }
             }


cubicj (commented):
I reproduced all four defects from #1530 on current main and ran the same cases against this branch, a match only in the last scanned message, the leading slash case, a flagless key, and the second of two keys matching. All four activate here, the `(?!)|` workaround keys people use today keep matching, and the error cases behave the same as before, non-`/` keys, unclosed regexes, and bad flags all still return false. The parse also handles slashes inside the pattern correctly, `/a\/b/i` and `/[/]/g` both split at the right place. `pnpm check`, `pnpm test`, and `pnpm build` pass on this head.

One change I would like to ask for. The parse and the `new RegExp` construction sit inside the message loop, but the keys never change during a scan, so every message re-parses and re-compiles every key, and recursive scanning repeats all of it each round. I would rather see the keys parsed and compiled once before the message loop. Two things to watch if you take this: a reused regex makes `.test()` stateful under the `g` and `y` flags, which your current version avoids by constructing a fresh regex each time, so the compiled regexes need their `lastIndex` reset or the flags handled some other way. And validating all keys up front means an entry with any malformed key is always inactive, where the current code stops validating later keys once an earlier one matches the first scanned message. I think the stricter behavior is the right one, but it is a small semantic change so I am calling it out.

Two smaller notes, neither blocking. A key of exactly `/` used to be skipped with the remaining keys still evaluated, and now it fails the parse and deactivates the entry. Nobody realistically has that key, I am noting it only because the PR body covers `//` but not this case. And in regex mode `@@exclude_keys_all` acts as ANY rather than ALL, since this branch returns on the first key match before `arg.all` is read. That is pre-existing rather than something this PR introduces, but regex exclude keys only start working with this fix, so this is where it will start to show.

Does hoisting the compilation look right to you, or is there a reason to prefer the per-iteration construction that I am missing?

renamerdis-wq:
Thanks for the thorough review :)
There wasn’t any particular reason for keeping the parsing and `RegExp` construction inside the message loop.

I’ll move them before the scan as you suggested, and I’ll make sure to reset `lastIndex` before each `.test()` call so the behavior of the `g` and `y` flags remains unchanged.

I’m currently traveling, so I’ll update the branch as soon as I’m back.

renamerdis-wq:
Sorry for the delayed update, and thanks for your patience 🙇🏻‍♀️

Updated the branch with the requested changes!

I moved regex parsing/compilation before the message scan and reset `lastIndex` before each `.test()` call.
The branch is also rebased onto the current upstream `main`, so the PR now contains a single commit on top of it.

I also manually verified the relevant cases in the RisuAI UI, including `/apple/g`, malformed regex handling, and the exact `/` case.

`pnpm check`, `pnpm test`, and `pnpm build` all pass.

Thanks again for the review :)

<details>
<summary>Verification screenshots (lore monitor + actual request body)</summary>

TEST1 (match only in the last scanned message)  
→ Activated correctly when the trigger appeared only in a later scanned message.

TEST2 (`/studio/i`, the leading-slash case from the issue)  
→ Activated correctly without treating the leading `/` as part of the pattern.

TEST3 (flagless `/flagless/`)  
→ Activated correctly with no regex flags.

TEST4 (second of two comma-separated keys)  
→ Activated correctly from the second key (`/포도밭/i`).

TEST5 (`/apple/g`)  
→ Activated correctly with the compiled global `RegExp` reused during the message scan.

TEST6 (`/foo/i,/(/`)  
→ Stayed inactive as expected because the malformed second regex fails upfront validation.

TEST7 (exact `/`)  
→ Stayed inactive as expected because the key does not match the regex literal format.

I also checked the actual request body to confirm that only TEST1-5 were inserted into the prompt, while TEST6-7 were not.

image
image
image
image
image
image

</details>

cubicj (approved):
I went through 4fa72fb5c against the two asks from the last round. The parse and the `RegExp` construction now run once before the message scan, and `lastIndex` is reset before each `.test()` call. I compared the reused compiled regexes against fresh per-iteration construction across the constructible flag combinations, including repeated `g` and `y` rounds, and found no mismatches. The four defect cases from #1530 and the `(?!)|` workaround keys give the same results as my pass on 970cd9827, and the upfront validation delta is now disclosed in the PR body.

`pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, and the branch sits on current main as a single commit.

PR #1534, opened 2026-07-14
fix: expand node storage write errors

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Improve Node storage write diagnostics so failed `setItem` calls expose actionable context instead of always reporting only `setItem Error`.

## Related Issues

None.

## Changes

The Node storage client now reports the logical storage key, payload size, HTTP status, and a sanitized, length-limited server response. It parses structured JSON errors, extracts useful text from HTML responses as a fallback, and distinguishes transport failures from HTTP failures.

The Node server now returns structured write errors with the filesystem error code and message. It logs the affected logical key, payload size, and full server-side error, while paths returned to the client are converted from absolute filesystem paths to relative `./save/...` paths.

Disk-full and quota errors return HTTP 507. Oversized `/api/write` payloads now return a structured HTTP 413 response with the `PAYLOAD_TOO_LARGE` code and the 100 MiB storage limit instead of relying on Express's HTML error page. Other write failures return HTTP 500.

Focused tests cover structured disk errors with relative paths, structured payload-limit errors, HTML error fallback behavior, and network failures.

## Impact

Successful writes and the existing save retry behavior are unchanged. When a write fails, users and server operators can distinguish request-size limits, network failures, insufficient disk space, quota exhaustion, and file permission errors without logging stored payload contents or authentication tokens. Absolute server filesystem paths are no longer exposed in client-facing write errors.

## Additional Notes

Validated with:

- `pnpm exec vitest run src/ts/storage/nodeStorage.test.ts` (4 tests)
- `pnpm check`
- `node --check server/node/server.cjs`
- `git diff --check`
- A live Node-server request exceeding 100 MiB, which returned the expected structured HTTP 413 response

### DIFF
--- a/server/node/server.cjs
+++ b/server/node/server.cjs
@@ -1,5 +1,7 @@
 const express = require('express');
 const app = express();
+const requestBodyLimit = '100mb';
+const requestBodyLimitLabel = '100 MiB';
 if (process.env.TRUST_PROXY) {
     app.set('trust proxy', Number(process.env.TRUST_PROXY) || process.env.TRUST_PROXY);
 }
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I went through 3cc8bfebe on this. I ran the head and main servers side by side with real and injected write failures behind the same authenticated client. A write into a directory comes back as a 500 with `EISDIR` and the `./save/...` path where main returned the Express HTML page, absolute path included outside production. `ENOSPC` and `EDQUOT` come back as 507, a payload one byte over the limit returns the `PAYLOAD_TOO_LARGE` JSON while exactly 100 MiB still goes through, and an oversized body on any other route keeps the default page. I also pointed the real `setItem` at that server and the messages come out with the key, size, status, and server line as described, and a transport failure comes back with its own cause line instead.

Just one thing to flag, not blocking. The `trim()` on the write path does not do anything for ordinary values, since Node already strips the surrounding spaces and tabs and `isHex` trims on its own. The one value it does change is a header padded with a non-breaking space, and there the write lands on the trimmed name while `/api/read` and `/api/remove` still look up the original. The built-in client only ever sends hex, so nothing reaches it in practice.

PR #1535, opened 2026-07-14
chore: update Vite to 8.1.5

SameDesu123:
## PR Checklist

- Required Checks
- [ ] If your PR is highly AI generated, check the following:

## Summary

Updates Vite from 8.0.3 to 8.1.5 and refreshes the pnpm lockfile.

## Related Issues

None.

## Changes

The Vite development dependency now targets 8.1.5. The lockfile includes the matching Vite dependency graph, including the updated Rolldown release used by Vite.

## Impact

This keeps the project on the latest Vite 8.1 release without changing the existing Node.js requirement or application behavior.

Validation completed with `pnpm check`, `NODE_OPTIONS=--no-experimental-webstorage pnpm test`, `pnpm exec vite --version`, and `git diff --check`.

## Additional Notes

Vite is now more stable, with several packages having graduated from RC to stable releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <img alt="text" src="https://raw.githubusercontent.com/kwaroran/Risuai/refs/heads/main/public/logo_typo_small.avif" width="400"/>
 </picture>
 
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I went through c4e65c087 with the full lockfile delta rather than the summary. Everything stays inside the vite subtree, rolldown moves from 1.0.0-rc.12 to stable 1.1.5 with its platform bindings, and the postcss, picomatch, and tinyglobby bumps are vite's own dependencies. No unrelated top-level changes. The `libc` fields showing up on sharp and rollup entries are serializer churn from a full lockfile rewrite, a frozen install under the pinned 10.34.1 passes, and a plain re-install leaves the file untouched.

The validation list in the PR body has no build, so I ran one. It passes with one new warning, `[SOURCEMAP_BROKEN]` on `@tailwindcss/vite`. The plugin is unchanged at 4.2.2 and never emitted sourcemaps for its transform, rolldown 1.1.5 just started reporting it, so nothing regressed there. I also checked vite.config.ts against the 8.1.0 to 8.1.5 changelog, none of the renames or deprecations reach this config and the new features are all opt-in. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, and it merges clean against current main.

PR #1536, opened 2026-07-14
fix: restore SafeMutationObserver payloads across the iframe bridge

poroyo:
## PR Checklist

- Required Checks

---

## Summary

The callback passed to `createMutationObserver` receives a bare `{__classType: 'REMOTE_REQUIRED'}` shell instead of the documented `SafeClassArray<SafeMutationRecord>`, so the example in plugins.md fails with `length is not a function`. Cause: the `INVOKE_CALLBACK` path was missing the REMOTE_REF conversion (both directions) that the request/response path already has.

Also wires up `removedNodes`: the observer already computes it (`removedNodes: elementMapHelper(mutation.removedNodes)`) but never passes it into `SafeMutationRecord`, so the value was silently dropped.

## Related Issues

None.

## Changes

- `factory.ts`: serialize `INVOKE_CALLBACK` args on the host side, deserialize in the guest bridge (2 lines)
- `v3.svelte.ts`, `risuai.d.ts`: add `getRemovedNodes()`, symmetric with `getAddedNodes()`
- `plugins.md`, `migrationGuide.md`: document `getRemovedNodes()`; the migration guide example still used the old plain-object API (updated to the current async-getter API)

## Impact

- Verified with a test plugin: all four getters return correct payloads across childList add/remove, attribute, and text mutations.
- The bug lives in the argument path shared by all plugin callbacks, so the fix applies to all of them, not just mutation observers. `serialize()` passes previously-working args (plain data, `AbortSignal`) through unchanged; verified with a published v3 provider plugin: generation and mid-generation abort still work.

## Additional Notes

Out of scope, noted while working on this:

- `attributeName` / `oldValue` not exposed; `disconnect()` is pending in #1307
~~- `instanceRegistry` entries are freed only by `release()` or iframe teardown, so a busy observer accumulates them. Will address in a separate PR.~~ **Update:** per review discussion, registry lifetime
  management now lands in this PR: guest-side `WeakMap` +
  `FinalizationRegistry` auto-release with batched messages, monotonic ids,
  and fail-closed handling of released refs. Details and measurements in the
  review thread below.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -1462,13 +1462,13 @@ Add clear comments and metadata:
 
     // Monitor DOM changes
     let changeCount = 0;
-    const observer = Risuai.createMutationObserver(async (mutations) => {
-      changeCount += mutations.length;
+    const observer = await Risuai.createMutationObserver(async (mutations) => {
+      changeCount += await mutations.length();
       await indicator.setTextContent(`Changes: ${changeCount}`);
     });
 
     if (body) {
-      observer.observe(body, {
+      await observer.observe(body, {
         childList: true,
         subtree: true,
         attributes: false
--- a/plugins.md
+++ b/plugins.md
@@ -512,7 +512,7 @@ await observer.observe(body, {
 });
 ```
 
-Note: `getAddedNodes()`/`getRemovedNodes()` contain HTML element nodes only.
+Note: `getAddedNodes()`/`getRemovedNodes()` contain HTML element nodes only; text nodes and SVG elements are excluded.
 
 ### SafeClassArray
 
--- a/plugins.md
+++ b/plugins.md
@@ -493,6 +493,12 @@ const observer = await Risuai.createMutationObserver(async (mutations) => {
     for (const node of addedNodes) {
       console.log(`Node added: ${await node.nodeName()}`);
     }
+
+    // getRemovedNodes() returns SafeClassArray<SafeElement>
+    const removedNodes = await Risuai.unwarpSafeArray(await mutation.getRemovedNodes());
+    for (const node of removedNodes) {
+      console.log(`Node removed: ${await node.nodeName()}`);
+    }
   }
 });
 
@@ -505,6 +511,8 @@ await observer.observe(body, {
   attributes: true
 });
 ```
+
+Note: `getAddedNodes()`/`getRemovedNodes()` contain HTML element nodes only.
 
 ### SafeClassArray
 
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -455,6 +455,7 @@ export class SandboxHost {
     private csp = `connect-src 'none'; script-src 'nonce-${this.nonce}' 'wasm-unsafe-eval'; frame-src 'none'; object-src 'none'; style-src * 'unsafe-inline'; default-src 'none'; img-src * data: blob:; font-src * data: blob:; media-src * data: blob:; base-uri 'none';`;
 
     private instanceRegistry = new Map<string, any>();
+    private refIdCounter = 0;
     private abortControllers = new Map<string, AbortController>();
     private messageHandlerRef: ((event: MessageEvent) => void) | null = null;
     private callbackWrapperCache = new Map<string, Function>();
@@ -533,7 +534,7 @@ export class SandboxHost {
             if (Array.isArray(val)) return val;
 
 
-            const id = 'ref_' + Math.random().toString(36).substring(2);
+            const id = 'ref_' + (++this.refIdCounter);
             this.instanceRegistry.set(id, val);
             return { __type: 'REMOTE_REF', id } as RemoteRef;
         }
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -333,7 +333,7 @@ await (async function() {
                         if (a.aborted) { controller.abort(); }
                         return controller.signal;
                     }
-                    return a;
+                    return deserializeResult(a);
                 });
                 const result = await fn(...deserializedArgs);
                 response.result = result;
@@ -584,7 +584,7 @@ export class SandboxHost {
                                 }
                                 return ref;
                             }
-                            return arg;
+                            return this.serialize(arg);
                         });
 
                         const message = {
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -40,7 +40,14 @@ await (async function() {
     const pendingRequests = new Map();
     const callbackRegistry = new Map();
     const callbackIdByFunction = new WeakMap();
-    const proxyRefRegistry = new Map();
+    const proxyRefRegistry = new WeakMap();
+    const releaseRegistry = new FinalizationRegistry((id) => {
+        try {
+            send({ type: 'RELEASE_INSTANCE', id: id });
+        } catch (_) {
+            // Frame is tearing down; host clears its registry anyway
+        }
+    });
     const abortControllers = new Map();
 
     function serializeArg(arg) {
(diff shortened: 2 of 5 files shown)


G4tsby (commented):
Not a maintainer, but sharing what I verified in case it helps the review.

The root-cause diagnosis checks out. `SafeClassArray` keeps its data in private fields (`#items`), so structured clone strips everything except the enumerable `__classType`, leaving exactly the bare `{__classType: 'REMOTE_REQUIRED'}` shell described in the PR. The two-line fix mirrors the existing request/response path: host-side `serialize()` registers the instance and emits a `REMOTE_REF`, guest-side `deserializeResult()` restores the async method proxy. Since `AbortSignal` is handled before the `serialize()` call and plain data passes through unchanged, the claim that previously-working args are unaffected holds. The `removedNodes` wiring, `getRemovedNodes()`, and the doc updates (`await mutations.length()`, `await observer.observe(...)`) all match the actual runtime behavior as far as I can tell.

A few observations, the first being the one I'd most want a second opinion on:

1. The instanceRegistry accumulation gets materially worse with this fix — and now pins detached DOM

    The PR notes registry growth as out of scope, but the fix changes its character a bit:

    - Before the fix, class-instance callback args were broken, so this path never registered anything. After the fix, every mutation batch registers an entry even if the plugin never touches the records — and `childList` + `subtree` on `body` makes this a high-frequency path.
    - Each registered `SafeClassArray` → `SafeMutationRecord` → `SafeElement` strongly references real DOM nodes. `removedNodes` are by definition detached, so subtrees that should be GC'd stay pinned until iframe teardown. Since `removedNodes` exposure is new in this PR, this part seems like a new risk rather than a pre-existing one.
    - `unwarpSafeArray` amplifies it: each `.at(i)` mints a fresh `ref_` id for the same element (`serialize()` has no dedup), and the guest-side `proxyRefRegistry` is a strong Map that grows in parallel.

    One possible direction for the plan: track the ids registered while serializing `INVOKE_CALLBACK` args per `reqId` and auto-release them on `CALLBACK_RETURN` (documenting that records are callback-scoped), or use a `FinalizationRegistry` on guest proxies to send `RELEASE_INSTANCE`.

2. "Applies to all callbacks" holds for top-level args only

    `serialize()` / `deserializeResult()` don't recurse, so a class instance nested in a plain object (e.g. `{records: safeArray}`) would still arrive as a bare shell. No current call site does this, so nothing to change — just noting the boundary of the fix for future readers.

3. Response/stream args still can't cross this path

    A `Response` arg now becomes `CALLBACK_STREAMS` with `value: val.body` (a `ReadableStream`), but `collectTransferables()` doesn't collect `ReadableStream`, and the `INVOKE_CALLBACK` path has no `replaceStreamsWithPorts()` like the response path does — so it still throws `DataCloneError` at `postMessage`. Not a regression (it was equally broken before); just noting that routing args through `serialize()` doesn't make stream args work.

4. Doc nit: "HTML element nodes only" also excludes SVG

    `elementMapHelper`'s `instanceof HTMLElement` filter drops SVG elements as well as text nodes. Pre-existing `addedNodes` behavior, so out of scope here, but a one-line mention in the docs might save SVG-touching plugins some surprise.

5. Test coverage

    apiV3 has no tests, and serialization round-trips are regression-prone. Even a minimal round-trip test for the `INVOKE_CALLBACK` arg path would help protect this fix.

Overall the core fix looks correct and minimal to me — hope this is useful context for the review.

poroyo:
@G4tsby Thanks for the detailed review!

1. Yes. This is the known trade-off I flagged in Additional Notes. I have a few approaches sketched, each with different guarantees, and I'm still weighing them. Planned as a separate PR to keep this one minimal.

2. That one's deliberate: `serialize()` is top-level on the existing RESPONSE path too, so I kept the callback path at parity rather than deepening both.

3. Right; not a regression, and no current callback call site passes a Response/stream, so leaving it out of scope here.

4. Agreed. The `instanceof HTMLElement` filter does exclude SVG elements, and that's not obvious, so it's worth stating explicitly. I've updated the note.

5. I did verify this change with a local test plugin, but as you say, having real test coverage there would be much more reassuring.

cubicj (commented):
I went through the fix in more depth. The two changed lines mirror what the RESPONSE path already does, and `pnpm check`, `pnpm test`, and a production build are clean at aeb27fb47. Since this branch predates the stream port bridge and the observer `disconnect()` from #1307 on current main, I also checked the merge result directly, and both sides compose correctly.

On the registry growth discussed above, I think it belongs in this PR rather than a follow-up. This fix is what makes the observer payloads cross the bridge at all, so the growth it enables is DOM-driven rather than plugin-driven. A `body` observer with `childList` and `subtree` registers the records array plus an entry per `at()`, `getTarget()`, `getAddedNodes()`, and `getRemovedNodes()` call on every mutation batch, on both sides of the bridge, and each retained `SafeElement` pins its DOM node, so the removed message subtrees a streaming chat keeps replacing stay alive until the plugin unloads. The only cleanup today is undocumented. `release()` works at runtime but is declared nowhere in `risuai.d.ts`, frees one host entry at a time, and leaves the guest `proxyRefRegistry` entry behind, the documented examples never call it, and `disconnect()` stops new batches without freeing anything already registered. For as long as the plugin stays loaded, the accumulation is the documented default usage.

You mentioned you have some approaches sketched. Would you be open to landing one of them here rather than in the separate PR? The shape is entirely up to you.

poroyo:
@cubicj Thanks for the thorough review; checking the merge result against current main ahead of time was especially helpful, since this update includes exactly that rebase. Took you up on landing it here: went with the FinalizationRegistry direction (one of the two @G4tsby sketched). Four commits, all in `src/ts/plugins/apiV3/factory.ts`.

### What changed

1. 0b4eb929 — guest `proxyRefRegistry` becomes a `WeakMap`, and a `FinalizationRegistry` sends the existing `RELEASE_INSTANCE` message when a proxy is garbage-collected (held value is the id string). The method-forwarding closures returned by the get trap now also capture their proxy, so an extracted method (`const at = arr.at`) keeps its instance alive.
2. 89efce43 — registry ids switch from `Math.random()` to a per-host monotonic counter. "Ids are never reused within a host lifetime" becomes structural, which makes a late or duplicate release a provable no-op.
3. bb49b30c — `deserializeArgs` rejects a `REMOTE_REF` whose id is no longer registered ("Instance not found or released", matching the method-call path) instead of passing the raw ticket object into the API, where it previously surfaced much later as a confusing `element.setAttribute is not a function`.
4. 1309ee29 — finalizer releases are batched into one `RELEASE_INSTANCE { ids: [...] }` per GC burst. Manual `release()` keeps the single-id form; the host accepts both, with an `Array.isArray` guard since the payload crosses the sandbox boundary.

### Why this shape

The main constraint was that existing plugins keep working with zero changes. `release()` is not part of the documented API, so any design that asks plugin authors to manage memory manually was off the table, and the deterministic alternatives (callback-scoped lifetimes, refcounting) all require opt-in contracts or protocol changes. Tying the host entry's lifetime to the guest proxy's reachability needs neither: no new message type, no new API surface, and the resulting lifetime rule is the one JS developers already expect: an object stays alive exactly as long as something can still call it.

### Characteristics

Reclamation timing is nondeterministic, as is inherent to GC-based resource management, so this bounds the in-session high-water mark rather than providing deterministic cleanup, and the existing deterministic paths (iframe teardown, manual `release()`) are unchanged and remain the floor. Message ordering: calls and releases travel through the same `postMessage` channel (FIFO per source), and the host resolves the instance synchronously before its first `await`, so a release cannot overtake or invalidate an in-flight call; batching only ever delays a release, which is the harmless direction. One invariant worth naming: GC-driven release relies on `serialize()` minting a fresh id per crossing, so releasing one proxy's id can never invalidate another proxy; adding identity dedup on top would require per-id refcounting first.

### Test setup and results

(Chromium; a local test plugin flips one attribute per tick, so every registry entry comes from the callback delivery path; forced-GC runs used `--js-flags=--expose-gc`)

- Before: the registry grows ~+1 per callback with nothing reclaiming it; a heap snapshot comparison after a few minutes showed +1008 live `SafeClassArray`, retained via `instanceRegistry`.
- After, drop-everything workload, 10.5-minute run with no forced GC: six natural GC cycles, registry oscillating between a floor of 3–8 (the plugin's live proxies) and peaks of ~25–142, each cleanup arriving as a single batched message. Heap snapshot delta for payload classes: 0 new.
- After, hold-everything workload: forced GC releases nothing while 206 payloads are pinned; dropping the pins then releases exactly 206 (no over-release).
- Extracted methods survive GC while held, and the entry is released once both the method and its proxy are dropped.

**Out of scope**: declaring `release()` in `risuai.d.ts` (documenting it permanently reserves `release`/`then` as method names on every remote object, so that call should be yours) and identity dedup (see the invariant above).

cubicj (approved):
I went through the four lifecycle commits at 1309ee29, mostly under Node with forced GC. Dropped proxies release as one batched message, held proxies survive, and an extracted method pins its owning instance until the method itself is dropped. Removing the `void proxy` capture reproduces a release while the extracted method is still callable, so that line is what makes extraction safe. The deterministic side, no id reuse after release, both release forms, the fail-closed rejection, checks out against the real `SandboxHost`, and `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One note, not a request. `deserializeResult` is shared with the RESPONSE path, so the reachability lifetime applies to every remote proxy, not just callback arguments. I could not construct a case where a plugin loses an entry it can still legitimately reach. Also, bb49b30c makes top-level data shaped like the internal ref marker reject at the call where main passed it through silently. I doubt any plugin does either, just disclosing the deltas.

On `release()` in `risuai.d.ts`, I would keep it undeclared. Manual release is now an optimization escape hatch, and declaring it would permanently reserve `release` and `then` as method names on every remote object.

PR #1540, opened 2026-07-16
feat: add Vercel AI Gateway support

SameDesu123:
## PR Checklist

- Required Checks
- [x] If your PR uses models, check the following:

## Summary

This pull request adds Vercel AI Gateway as a built-in model provider. It supports both Chat Completions and Responses API formats and exposes the related model and provider routing controls directly at the bottom of the Model settings tab.

## Related Issues

None.

## Changes

The Vercel AI Gateway provider is integrated into the existing OpenAI-compatible request pipeline with dedicated API endpoints, authentication, model discovery, streaming support, and request format selection. The Model settings tab now provides controls for selecting a Vercel model, choosing the API format, configuring provider preference and exclusion rules, selecting routing and service-tier options, and enabling the supported privacy and caching options. These settings are persisted in both the database and presets, with localized labels and help text for all supported languages.

Provider metadata is loaded from Vercel and cached to avoid unnecessary repeated requests. Provider exclusion settings are converted into the Gateway-supported allowlist format when requests are assembled. Saved provider preference entries are also validated against the providers currently available for the selected model, preventing stale provider identifiers from being sent after changing models.

## Impact

The new behavior is limited to the Vercel AI Gateway provider. Existing providers continue through their previous request paths, and the Vercel-specific request options are applied only when Vercel AI Gateway is selected. Existing saves receive safe defaults for the newly added settings, while older Vercel allowlist data remains compatible.

## Additional Notes

Validation was performed with `pnpm check` and focused Vitest coverage for Vercel model discovery, provider routing options, Chat Completions integration, Responses API integration, and non-Vercel request isolation. All 39 focused tests passed.
Also tested manually with GPT-5-nano.


SameDesu123:
I think this PR is too large for just adding a single provider, so I’m closing it for now.
I might come back with a smaller, more focused PR later.

PR #1543, opened 2026-07-16
feat: expose cache points to V3 plugins

### LINKED ISSUE #1538: Plugin API: expose cachePoint in V3 OpenAIChat messages
## Summary

RisuAI already supports `cachePoint` on its internal `OpenAIChat` messages, and prompt-template cache cards use this field internally.

However, the public Plugin API v3 `OpenAIChat` interface does not expose it. As a result, a V3 plugin using `beforeRequest` cannot set cache boundaries through the supported TypeScript API without using `any` or an unsafe cast.

Please expose `cachePoint?: boolean` as an officially supported V3 message field.

## Current gap

The internal message type contains:

```ts
interface OpenAIChat {
    // ...
    cachePoint?: boolean
}
```

The public V3 type in `src/ts/plugins/apiV3/risuai.d.ts` currently only exposes:

```ts
interface OpenAIChat {
    role: 'system' | 'user' | 'assistant' | 'function';
    content: string;
    name?: string;
    function_call?: {
        name: string;
        arguments: string;
    };
}
```

Both `addRisuReplacer('beforeRequest', ...)` and `ProviderArguments.prompt_chat` use this public type, even though the runtime pipeline and custom providers can already carry cache-point metadata.

## Requested API

```ts
interface OpenAIChat {
    role: 'system' | 'user' | 'assistant' | 'function';
    content: string;
    name?: string;
    function_call?: {
        name: string;
        arguments: string;
    };

    /**
     * Marks the end of a reusable prompt prefix.
     * The cache boundary is immediately after this message,
     * including this message in the prefix.
     *
     * Providers that do not support explicit prompt caching may ignore it.
     */
    cachePoint?: boolean;
}
```

Example plugin usage:

```ts
await risuai.addRisuReplacer('beforeRequest', async (messages) => {
    return messages.map((message, index) => ({
        ...message,
        cachePoint: shouldCacheAfter(index),
    }));
});
```

## Use case

A provider-independent plugin can analyze the final request on every turn and mark stable prefix boundaries, such as:

- after stable system/persona instructions
- before long-term memory
- before chat history
- at a repeatedly observed prompt prefix

A custom provider can then translate the same logical boundary into its native caching mechanism, for example:

- Anthropic `cache_control`
- OpenAI GPT-5.6+ `prompt_cache_breakpoint`
- a Gemini explicit cached-content boundary

The planner plugin should not need to know or modify the provider-specific request body.

## Required runtime semantics

Exposing the type alone is useful, but the boundary also needs predictable runtime semantics:

1. A `cachePoint` returned by `beforeRequest` should be preserved through `ProviderArguments.prompt_chat`.
2. The boundary should remain immediately after the marked message.
3. Message normalization must not merge later content into a message with `cachePoint: true`, because that would silently move the boundary.
4. Providers that do not support the field may safely ignore it.

For example, these messages:

```ts
[
    { role: 'user', content: 'stable prefix', cachePoint: true },
    { role: 'user', content: 'changing suffix' },
]
```

must not become:

```ts
[
    {
        role: 'user',
        content: 'stable prefix\nchanging suffix',
        cachePoint: true,
    },
]
```

The latter changes the cache boundary from “after stable prefix” to “after changing suffix”.

## Scope

For the initial API, only the existing provider-neutral boolean field is requested:

```ts
cachePoint?: boolean
```

Provider-specific TTLs, cache resource management, preset access, and provider request-body APIs are intentionally out of scope.

## Why this should be low-risk

- The internal field and prompt-template cache-point behavior already exist.
- `beforeRequest` already allows plugins to transform the final message list.
- Custom providers already receive the transformed list through `prompt_chat`.
- Unsupported providers can simply ignore the optional field.

The main changes should be public typing/documentation, preservation of the boundary during normalization, and tests for the `beforeRequest` → custom provider path.

I would be happy to submit a PR after confirming the intended semantics.

pnya2021:
## Summary

- expose the existing `cachePoint?: boolean` field on the public Plugin API V3 `OpenAIChat` type
- prevent shared message normalization from merging later content past a cache boundary
- preserve cache boundaries while folding leading system messages
- add focused regression coverage for same-role and leading-system normalization

## Scope

This exposes provider-neutral cache metadata to `beforeRequest` replacers and custom providers through `ProviderArguments.prompt_chat`. Provider-specific request-body translation remains unchanged, and unsupported providers may ignore the optional field.

## Testing

- `pnpm vitest run src/ts/process/request/tests/reformater.test.ts` — 4 passed
- `pnpm test` — 196 passed, 3 skipped
- `pnpm check` — 0 errors, 0 warnings
- `pnpm build` — passed

Closes #1538

### DIFF
--- a/src/ts/plugins/apiV3/risuai.d.ts
+++ b/src/ts/plugins/apiV3/risuai.d.ts
@@ -238,6 +238,11 @@ interface OpenAIChat {
         name: string;
         arguments: string;
     };
+    /**
+     * Marks the end of a reusable prompt prefix, including this message.
+     * Providers that do not support explicit prompt caching may ignore it.
+     */
+    cachePoint?: boolean;
 }
 
 /**
--- a/src/ts/process/request/request.ts
+++ b/src/ts/process/request/request.ts
@@ -354,9 +354,12 @@ export function reformater(formated:OpenAIChat[],modelInfo:LLMModel|LLMFlags[]){
 
     if(!flags.includes(LLMFlags.hasFullSystemPrompt)){
         if(flags.includes(LLMFlags.hasFirstSystemPrompt)){
-            while(formated[0].role === 'system'){
+            while(formated[0]?.role === 'system' && !systemPrompt?.cachePoint){
                 if(systemPrompt){
                     systemPrompt.content += '\n\n' + formated[0].content
+                    if(formated[0].cachePoint){
+                        systemPrompt.cachePoint = true
+                    }
                 }
                 else{
                     systemPrompt = formated[0]
@@ -382,7 +385,7 @@ export function reformater(formated:OpenAIChat[],modelInfo:LLMModel|LLMFlags[]){
                 continue
             }
 
-            if(newFormated[newFormated.length-1].role === m.role){
+            if(newFormated[newFormated.length-1].role === m.role && !newFormated[newFormated.length-1].cachePoint){
             
                 newFormated[newFormated.length-1].content += '\n' + m.content
 
--- a/src/ts/process/request/tests/reformater.test.ts
+++ b/src/ts/process/request/tests/reformater.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+    db: {
+        systemContentReplacement: '',
+        systemRoleReplacement: '',
+    },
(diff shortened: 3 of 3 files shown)


sub-hub (approved):
Overall, it is a stable fix. The reformatter, in particular, seems to have been modified well.
Thanks for the PR.

PR #1545, opened 2026-07-18
fix: support explicit cache points for OpenAI GPT-5.6

seto-sama:
## PR Checklist

- Required Checks

## Summary

Adds explicit prompt cache breakpoint support for OpenAI GPT-5.6 models.

## Related Issues

Related to #1533 and #1543.

## Changes

- Converts prompt preset cache points into OpenAI `prompt_cache_breakpoint` content blocks.
- Supports both Chat Completions and Responses API request formats.
- Supports built-in and OpenAI-compatible custom GPT-5.6 models.
- Keeps cache point fields out of requests for earlier models.
- Adds type definitions and focused regression tests.

## Impact

Prompt preset cache points now use GPT-5.6 explicit prompt caching. Existing behavior for earlier models remains unchanged.

## Additional Notes

Manually tested Responses API GPT-5.6 Sol, Terra, and Luna twice each using the node-hosted version. Local builds were not tested because I do not currently have a local build environment configured.

### DIFF
--- a/src/ts/process/request/openAI/promptCache.test.ts
+++ b/src/ts/process/request/openAI/promptCache.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from 'vitest'
+
+import { addPromptCacheBreakpoint, applyGPT56ChatCachePoints, responsesCacheableContentTypes } from './promptCache'
+
+describe('OpenAI GPT-5.6 explicit prompt caching', () => {
+    it('converts a GPT-5.6 chat cache point into a text content breakpoint', () => {
+        expect(applyGPT56ChatCachePoints([
+            { role: 'system', content: 'Stable instructions', cachePoint: true },
+            { role: 'user', content: 'Changing request' },
+        ], 'gpt-5.6-sol')).toEqual([
+            {
+                role: 'system',
+                content: [{
+                    type: 'text',
+                    text: 'Stable instructions',
+                    prompt_cache_breakpoint: { mode: 'explicit' },
+                }],
+            },
+            { role: 'user', content: 'Changing request' },
+        ])
+    })
+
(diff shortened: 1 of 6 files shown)


G4tsby (commented):
Ran the new `promptCache.test.ts` locally — all 5 pass. I also traced `cachePoint` through the request paths: it never reaches the wire (the Mistral early-return, OpenRouter `useInstructPrompt`, and tool-call rebuild paths all construct fresh objects), and removing the old `delete formatedChat[i].cachePoint` actually fixes a pre-existing leak for `function`-role messages. The cacheable block-type whitelists also match the documented sets exactly for both APIs. Nice work.

Two notes:

1. The [prompt caching guide](https://developers.openai.com/api/docs/guides/prompt-caching) says that on GPT-5.6+ you must set `prompt_cache_key` for reliable cache matching. The PR adds breakpoints but never sets `prompt_cache_key`, so explicit caching may still miss more than expected. Deriving it from `arg.chatId` seems like a natural fit — worth adding here, or as a follow-up?
2. Minor: a `cachePoint` message dropped by the `newOAIHandle` empty-content filter (or the `item.content.length > 0` guard in Responses) silently loses its breakpoint instead of moving it to an adjacent message — probably acceptable, but it's a silent cache miss.

seto-sama:
Thank you for the thorough and helpful review! Regarding the two notes:

1. Users can manually set prompt_cache_key through the additional parameters in a custom model or the chat bot settings. I’ll use the chat ID as the automatic default, while allowing a manually configured key to override it.
However, additional parameters do not appear to be applied on the native Responses API path. This seems separate from the scope of this PR. Would it be better to address it separately?
2. I’ll document that a cache point is also dropped when its empty message is filtered out.

PR #1550, opened 2026-07-24
refactor: extract database change accumulator from saveDb

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Extract database mutation classification and pending-save tracking from `saveDb()` into a dedicated `DatabaseChangeAccumulator`.

The accumulator converts existing `DatabaseUpdateInfo` events into the existing `toSaveType` representation, supports atomic take/reset behavior, deduplicates affected character IDs, and restores pending changes after failed saves.

## Related Issues

None.

## Changes

- Add a Svelte-independent `DatabaseChangeAccumulator` module.
- Preserve the existing classification rules for characters, presets, modules, loadouts, plugins, and plugin custom storage.
- Preserve full encoder reload handling for database root replacements.
- Move pending-change snapshot and failed-save restoration logic out of `saveDb()`.
- Keep a separate encoder input copy because `RisuSaveEncoder.set()` mutates the character list it receives.
- Add focused tests for classification, deduplication, root replacement, take/reset behavior, independent snapshots, in-flight changes, and failed-save restoration.

## Impact

This is intended to be a behavior-preserving refactor. It does not change the database proxy, save format, encoder behavior, debounce timing, storage writes, retry policy, or backup behavior.

The extraction gives save change tracking a focused boundary while retaining pending mutations that occur during an active or failed save.

## Additional Notes

Validation completed:

- `pnpm exec vitest run src/ts/storage/databaseChangeAccumulator.test.ts src/ts/storage/databaseState.svelte.test.ts` - 48 tests passed
- `pnpm check` - 0 errors and 0 warnings
- `git diff --check` - passed

### DIFF
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -25,7 +25,8 @@ import { hasher } from "./parser/parser.svelte";
 import { characterURLImport, hubURL } from "./characterCards";
 import { defaultJailbreak, defaultMainPrompt, oldJailbreak, oldMainPrompt } from "./storage/defaultPrompts";
 import { loadRisuAccountData } from "./drive/accounter";
-import { decodeRisuSave, encodeRisuSaveLegacy, RisuSaveEncoder, type toSaveType } from "./storage/risuSave";
+import { decodeRisuSave, encodeRisuSaveLegacy, RisuSaveEncoder } from "./storage/risuSave";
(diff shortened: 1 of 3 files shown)


SameDesu123:
Superseded after the DB proxy implementation was reverted in #1594. A new PR based on the current snapshot/effect implementation will follow.

PR #1555, opened 2026-07-25
chore: update Svelte to 5.56.8

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions? (No new type definitions are required.)

## Summary

Update Svelte from 5.55.1 to 5.56.8.

## Related Issues

None.

## Changes

- Bump the Svelte development dependency to 5.56.8.
- Refresh the lockfile with the matching Svelte compiler and runtime dependencies.

## Impact

This update brings recent Svelte runtime, compiler, and reactivity fixes without changing application source code. Existing Svelte plugins remain within their supported peer dependency ranges.

## Additional Notes

Validated with:

- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm check`
- `pnpm test`
- `git diff --check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "internal-ip": "^7.0.0",
     "postcss": "^8.5.6",
     "source-map": "^0.7.6",
-    "svelte": "^5.55.1",
+    "svelte": "^5.56.8",
     "svelte-check": "^4.3.5",
     "tailwindcss": "^4.2.2",
     "tslib": "^2.8.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -31,7 +31,7 @@ importers:
         version: 3.8.1
       '@lucide/svelte':
         specifier: ^0.562.0
-        version: 0.562.0(svelte@5.55.1)
+        version: 0.562.0(svelte@5.56.8(@typescript-eslint/types@8.58.0))
       '@mlc-ai/web-llm':
         specifier: ^0.2.80
         version: 0.2.80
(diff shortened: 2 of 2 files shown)


cubicj (approved):
I went through b4787f155 with the full lockfile delta. Svelte's own dependencies move with it, esrap to 2.3.0, devalue to 5.8.2, and acorn-typescript to 1.0.11, all inside the ranges 5.56.8 declares, and nothing else changes version. The `libc` lines on the platform binaries are the same serializer churn as in #1535. `@typescript-eslint/types` turns into an optional peer of esrap and is only referenced from its JSDoc, and neither esrap nor devalue reaches the client bundle.

I also compiled all 179 components with both versions. After template normalization 175 match, and the remaining differences are factory names, equivalent parentheses on an optional chain, and an exclusion set for an unused props object. The hoisted template deduplication from 5.56.0 removes 404 template declarations, and the production JS comes out about 15 KB smaller overall. On the runtime side I traced the lazy prop fallback change, and in our own components only `rawStreamingText = message` in `Chat.svelte` reads another prop. Every output it feeds is gated by `isOptimizedStreamingMessage`, which only `Chats.svelte` sets, and that caller supplies the value on both mount and update, so the sites that omit it keep those paths disabled.

The bump also clears the five open Dependabot alerts on the lockfile, the four Svelte advisories fixed in 5.55.7 and the devalue one, and I found no matching usage in the app or its compiled dependencies. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head and on a merge with current main.

PR #1556, opened 2026-07-26
feat: add sticky chat toolbar option

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Adds an optional setting that keeps each message's chat toolbar and request information visible while scrolling through long responses.

## Related Issues

None.

## Changes

- Added a migration-safe `stickyChatToolbar` database setting and exposed it under Accessibility settings.
- Added localized setting labels for all currently supported languages.
- Separated the message title from the toolbar so only the action buttons and request information become sticky.
- Added theme-aware spacing and an opaque background so message content and interactive Lua buttons scroll cleanly behind the toolbar instead of showing through it.

## Impact

The setting is disabled by default, so existing chat behavior remains unchanged. When enabled, message actions and request details remain accessible while reading long messages.

## Preview

https://github.com/user-attachments/assets/1fa7328f-9e01-4036-90da-ab0c80afdb10
+ Also works well with Request Info option

## Additional Notes

Validated with:

- `pnpm check`
- `git diff --check`

### DIFF
--- a/src/lang/cn.ts
+++ b/src/lang/cn.ts
@@ -981,6 +981,7 @@ export const languageChinese = {
     "enable": "激活",
     "postFile": "上传文件",
     "requestInfoInsideChat": "在聊天中显示请求数据",
+    "stickyChatToolbar": "滚动时固定聊天工具栏和请求信息",
     "inputTokens": "输入 Tokens",
     "outputTokens": "输出 Tokens",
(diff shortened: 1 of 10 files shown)


SameDesu123:
Since this can already be implemented sufficiently with custom css, I don't think there is much benefit to adding it as a built-in feature.

PR #1558, opened 2026-07-27
Fix module charx conversion: preserve moduleNamespace, hideChatIcon, and Replace Global Note

tammy3211:
## PR Checklist

- Required Checks

## Summary

add `moduleNamespace` and `hideChatIcon` to save charX correctly

## Related Issues

when we save module with charx namespace and hideChatIcon missed.

## Changes

add type `moduleNamespace` and `hideChatIcon` on convertModuleToCharacter and charctercard function

## Additional Notes

~~@@indicator phi is not working because `postHistoryInstructions` is legacy field. I'm not sure about this work, so in this pr I do not address this error.~~

**Update by @sub-hub (2026.08.07):** the issue above is now addressed in the commits below:
- `@@indicator` content now maps to `replaceGlobalNote` (the live field) instead of the legacy `postHistoryInstructions`
- the marker was renamed `@@indicator phi` > `@@indicator replace_global_note`
- legacy `@@indicator phi` markers are still read for backward compatibility


sub-hub (approved):
Hi @tammy3211, the Additional Notes mention that `@@indicator phi` doesn't work because `postHistoryInstructions` is a legacy field. I've added a few commits on top of your branch that address this:

- `2b0bfae` — maps the module's global-note entry to `replaceGlobalNote` (the live field) instead of the legacy `postHistoryInstructions`. The old field was being migrated into the chat note by `characterFormatUpdate`, so module global notes never reached the actual Replace Global Note field.
- `3fcd808` and the following commit — rename the marker `@@indicator phi` > `@@indicator replace_global_note` so it matches the `replaceGlobalNote` field name and can't be confused with the user-level Global Note setting.
- `72c53b8` — keeps reading the legacy `@@indicator phi` marker for backward compatibility with modules created before the rename.

All changes are limited to the module<>character conversion path in `src/ts/interchangeability.ts`. I also updated the PR title and description to reflect the wider scope. If you'd prefer these in a separate PR, I can split them out.

Anyway, since this patch involves minimal code modifications and only fixes clearly missing information, I think it is fine to merge it.
Thanks for the PR.

cubicj (commented):
I traced the three channels on 6d0f4e743 against current main at f3f0242fb, and all three losses are real. The namespace never survives the conversion, `createBaseV3` never writes `hideChatIcon`, and the global note dies in both directions because the conversions still use the legacy `postHistoryInstructions` while the card layer already uses `replaceGlobalNote`. The round trip works at this head, legacy `@@indicator phi` entries are still read, and marker text never reaches an exported card, so older clients lose nothing they keep today. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, and the merge against main has no overlapping hunks.

One thing I noticed while testing. `convertModuleToCharacter` splices the marker entries out of `m.lorebook` through a shared array reference, so converting or exporting a module deletes the global note entry from the module stored in the database. The aliasing is pre-existing and used to be mostly dormant, but after this PR every charx-imported module with a global note carries such an entry, so a plain import-then-export flow silently loses the note from the live module. I reproduced it at this head.

Smaller note, not blocking. Re-importing the same module charx now leaves both copies active for namespace references, since ids are regenerated on import and dedup is by id only. That was already reachable through `.risum`, just disclosing.

Could you take a look at the mutation side when you get a chance?

sub-hub:
@cubicj You're right, thanks for catching this. I've applied the fix: `convertModuleToCharacter` now clones the module's lorebook with `safeStructuredClone` before splicing the marker entries, so converting or exporting no longer mutates the module stored in the database. I missed the aliasing because that code predates this PR, but your analysis was spot on.

On the namespace dedup note, agreed it's worth addressing. I'll follow up on that separately.

cubicj (approved):
I went through 3c8b2f4f1. The clone lands at the assignment, before the marker loop, so the splices work on a copy and the stored module keeps its entries. The repro from my last comment no longer loses the entry at this head, and repeated conversions now stay consistent, where previously the first convert or export stripped the note from the module and every later one produced a charx without it. The convert button, `exportModule`, and the persona path all go through `convertModuleToCharacter`, so the single clone covers all three, and since the function already returned a deep clone, the conversion output is unchanged, only the input mutation is gone. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, and the merge against main is still conflict-free.

Approving — thanks for the fix!

PR #1571, opened 2026-07-29
fix: validate message source in Plugin V3 guest iframe

rsyumi:
## PR Checklist

- Required Checks

## Summary

Adds source validation for messages received in the Plugin V3 iframe. This aligns the guest bridge's behavior with the parent bridge and prevents unintended messages from other plugins from being processed.

## Related Issues

None.

## Changes

In `GUEST_BRIDGE_SCRIPT` (injected into the Plugin V3 iframe), the `message` event handler now verifies that `event.source` matches `window.parent` before processing the message.

## Impact

Cross-plugin messages, which were previously accepted by default, are now blocked. Since sending messages to another plugin's iframe is not a normal usage pattern, this is expected to have no practical effect on existing functionality.

## Additional Notes

This is a defense-in-depth change, not a fix for a confirmed exploit. Meaningful message forgery would require a valid pending `reqId` or `abortId`, with no known way for another plugin to observe them. `EXECUTE_CODE` is legacy code that cannot run under the iframe's current CSP, so it is not exploitable through this path.

The parent bridge already validates that an incoming message originates from the corresponding iframe; the guest bridge was missing the equivalent check. This change simply brings the two sides in line.

### DIFF
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -271,6 +271,8 @@ await (async function() {
     
     
     window.addEventListener('message', async (event) => {
+        if (event.source !== window.parent) return;
+
         const data = event.data;
         if (!data) return;
 


cubicj (approved):
I went through every sender that can reach the guest listener before approving. The host's five postMessage sites all pass the new check, plugin channel IPC is relayed entirely through the host so it never touches the window path, and stream traffic rides MessagePorts that this listener never sees. The only thing the guard blocks is raw postMessage from a sibling plugin frame, which was also the one route that skipped the mutual `allowed-ipc` checks the channel API enforces, so the change closes a real permission gap on top of matching the host-side checks. `pnpm check`, `pnpm test`, and `pnpm build` are clean on 49d08fdab, and there is no main-side drift on this file.

PR #1572, opened 2026-07-29
fix: block V3 plugin list changes via setDatabaseLite

rsyumi:
## PR Checklist

- Required Checks

## Summary

Fixes an issue where Plugin V3 could register arbitrary Plugin V2.1 code through `setDatabaseLite`, closing a path to arbitrary code execution.

## Related Issues

None.

## Changes

Adds plugin list validation to the Plugin V3 `setDatabaseLite` API, consistent with the restrictions already applied by `setDatabase`.

Modified plugin lists are ignored with a console warning, while other database changes continue normally.

## Impact

Plugin V3 can no longer add, remove, or modify plugins through `setDatabaseLite`. Attempts are ignored without throwing an error.

Existing functionality is unaffected when the plugin list remains unchanged.

## Additional Notes

Plugin V3 can still install other Plugin V3 plugins through the existing validated installation path.

A console warning is emitted because blocked changes otherwise fail silently, allowing users and plugin developers to identify the attempt. If this warning is unnecessary, please let me know; removing it would simplify the implementation.


cubicj (commented):
I went through every V3 entry that can reach the plugin list on de29b0646. `setDatabase` is the only other one that can write it and it already validates, `pluginV2` is in `allowedDbKeys` but nothing reads it back, and call arguments are serialized before they reach the host, so the shallow copy and `delete` cover the only `plugins` key that can arrive. It also takes the plugin list out of the wholesale replace that `setDatabaseLite` does on every allowed key, so a plugin writing back a stale `getDatabase()` snapshot no longer reverts the installed list the way it does on main. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One change I would like to ask for. Production builds strip `console.*` but only the call, so the `isEqual` and `$state.snapshot` pair stays behind as a bare expression whose result is discarded. `getDatabase` defaults to every allowed key, so `plugins` comes back on ordinary read-modify-write calls, and those calls deep snapshot and compare the whole installed list in production for a warning that is not there. The `delete` runs either way, so the comparison is not part of the block, and I would rather it not run in production.

Smaller note, not blocking. `risuai.d.ts` still documents `plugins` on what `setDatabaseLite` accepts, and it is ignored there now, while `setDatabase` still takes it.

rsyumi:
> I went through every V3 entry that can reach the plugin list on [de29b06](https://github.com/kwaroran/Risuai/commit/de29b06465f7342ae37f9ae18eb0bf1efd719ac1). `setDatabase` is the only other one that can write it and it already validates, `pluginV2` is in `allowedDbKeys` but nothing reads it back, and call arguments are serialized before they reach the host, so the shallow copy and `delete` cover the only `plugins` key that can arrive. It also takes the plugin list out of the wholesale replace that `setDatabaseLite` does on every allowed key, so a plugin writing back a stale `getDatabase()` snapshot no longer reverts the installed list the way it does on main. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.
> 
> One change I would like to ask for. Production builds strip `console.*` but only the call, so the `isEqual` and `$state.snapshot` pair stays behind as a bare expression whose result is discarded. `getDatabase` defaults to every allowed key, so `plugins` comes back on ordinary read-modify-write calls, and those calls deep snapshot and compare the whole installed list in production for a warning that is not there. The `delete` runs either way, so the comparison is not part of the block, and I would rather it not run in production.
> 
> Smaller note, not blocking. `risuai.d.ts` still documents `plugins` on what `setDatabaseLite` accepts, and it is ignored there now, while `setDatabase` still takes it.

Thank you.

I made the two suggested changes, the validation now runs only in development, and the type definition has been updated.

When you have a time, please check whether these changes looks ok.

cubicj (approved):
I checked the production output first. The build folds `import.meta.env.DEV` to a constant, so the comparison block is gone from the emitted chunk entirely, the warning string survives only in the sourcemap, and the strip and `delete` still run unconditionally in both modes. `pnpm check`, `pnpm test`, and `pnpm build` are clean on 9d95cb209.

The type change lands well too. Using `Omit` on the signature keeps `DatabaseSubset` itself intact, so `setDatabase` still documents `plugins`, a literal `plugins` argument now fails to compile, and a stored `getDatabase()` result still passes, matching what the runtime does.

One note, nothing to change here. The new doc line points authors to `setDatabase()` for plugin updates, and on current main that round trip clears the installed list because the install path only returns newly approved entries. That is pre-existing behavior, and #1598 fixes it separately.

rsyumi:
> I checked the production output first. The build folds to a constant, so the comparison block is gone from the emitted chunk entirely, the warning string survives only in the sourcemap, and the strip and still run unconditionally in both modes. , , and are clean on [9d95cb2](https://github.com/kwaroran/Risuai/commit/9d95cb2091c940f1230a64f737186ebc09342b79).`import.meta.env.DEV``delete``pnpm check``pnpm test``pnpm build`
> 
> The type change lands well too. Using on the signature keeps itself intact, so still documents , a literal argument now fails to compile, and a stored result still passes, matching what the runtime does.`Omit``DatabaseSubset``setDatabase``plugins``plugins``getDatabase()`
> 
> One note, nothing to change here. The new doc line points authors to for plugin updates, and on current main that round trip clears the installed list because the install path only returns newly approved entries. That is pre-existing behavior, and #1598 fixes it separately.`setDatabase()`
> 
> Approving — thanks for the update!

Thanks for checking. :)

I ran into the same issue as #1598 myself, however I kept meaning to split it into a separate PR but never got around to fixing it. Not that you did it on my behalf or anything, but thanks for taking care of it!

PR #1573, opened 2026-07-29
fix: scope and enforce plugin permissions

SameDesu123:
## PR Checklist

- Required Checks
    - [x] Have you checked that it will not break existing features?

## Summary

Scope cached plugin permission decisions to both the plugin script and the requested permission, and ensure a denied provider permission prevents provider invocation.

## Related Issues

None.

## Changes

- Replace plugin-name-wide session allow and deny sets with a permission-aware session cache.
- Key session decisions by the loaded plugin script hash and permission name.
- Lazily compute each loaded plugin instance's script hash once and reuse it on cache hits.
- Stop V3 provider callbacks from running when provider permission is denied.
- Preserve the existing persistent permission key format.
- Add focused tests for permission isolation, denied permissions, script changes, persistent key compatibility, lazy hash reuse, and provider permission enforcement.

## Impact

Granting or denying one plugin permission no longer implicitly grants or denies every other permission requested by that plugin. Updating a plugin script also invalidates its in-memory permission decisions for the session. Existing persisted grants remain compatible.

A denied provider permission now returns a failed provider response without invoking the plugin callback. Allowed provider requests are unchanged.

## Additional Notes

This changes model request behavior only when the user explicitly denies provider permission; previously, the denied result was ignored and the provider callback still ran.

Validation:

- `pnpm check`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/ts/plugins/apiV3/pluginPermissionCache.test.ts` (7 tests)
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test` (235 passed, 3 skipped)

### DIFF
--- a/src/lang/cn.ts
+++ b/src/lang/cn.ts
@@ -1428,6 +1428,7 @@ export const languageChinese = {
     "mainDomAccessConsent": "插件 {} 正在请求访问主文档，这可能会泄露敏感信息。是否允许？",
     "getFullDatabaseConsent": "插件 {} 正在请求访问完整数据库，这可能会泄露敏感信息。是否允许？",
     "fetchLogConsent": "插件 {} 正在请求获取日志，这可能会泄露敏感信息。是否允许？",
+    "providerPermissionDenied": "用户拒绝了插件的服务商访问权限。",
     "enableScrollToActiveChar": "启用滚动至当前角色",
     "trimStartNewChat": "修剪“开始新对话”消息",
     "hamburgerButtonBottom": "将菜单按钮移至侧边栏底部",
--- a/src/lang/de.ts
+++ b/src/lang/de.ts
@@ -1427,6 +1427,7 @@ export const languageGerman = {
     "fetchLogConsent": "Plugin {} möchte Protokolle abrufen, was sensible Informationen offenlegen könnte. Möchten Sie dies zulassen?",
     "getFullDatabaseConsent": "Das Plugin {} möchte auf die gesamte Datenbank zugreifen, was zur Offenlegung sensibler Informationen führen kann. Möchten Sie dies zulassen?",
     "mainDomAccessConsent": "Plugin {} möchte auf das Hauptdokument zugreifen, wodurch sensible Informationen offengelegt werden könnten. Möchten Sie dies zulassen?",
+    "providerPermissionDenied": "Der Benutzer hat dem Plugin den Zugriff auf den Anbieter verweigert.",
     "enableScrollToActiveChar": "Scrollen zum aktiven Charakter aktivieren",
     "newMessageButtonRightCenter": "Mitte rechts",
     "trimStartNewChat": "„Neuen Chat starten“-Nachrichten kürzen",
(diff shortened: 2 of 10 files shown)


cubicj (commented):
I checked the persistent key path first, since existing grants hang on it. The new session key matches the format `permissionForage` already uses, so per-permission grants from before this PR keep working, and denials stay session-only as before. The name-wide short circuit on main reproduces, one decision covering all six permission types for the session. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, with the new tests running in the default suite.

What gets worse is the cache hit path. On main it was a synchronous `Set` lookup, but now every call finds the plugin entry, encodes the whole script, and awaits a SHA-256 before the cache can answer. That is around 0.12ms per call for a 50KB script and 2ms for a 2MB one, and it runs on per-call APIs like `getDatabase` and provider invocations.

What do you think about this?

SameDesu123:
Good catch. I addressed this in `016d423b` by memoizing a lazy script-hash promise per loaded V3 plugin instance and passing the resolved hash into `getPluginPermission`.

This keeps decisions scoped by script hash and permission, preserves the existing persistent key format, and removes the database lookup, script encoding, and SHA-256 work from cache-hit calls. Because the hash getter closes over the script for the loaded plugin instance, a reload with updated code creates a new hash while plugins that never request a protected permission pay no hashing cost.

I also added coverage confirming that repeated and concurrent requests share one hash computation. Validation passed:

- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/ts/plugins/apiV3/pluginPermissionCache.test.ts` (5 tests)
- `pnpm check`
- `git diff --check`

cubicj (approved):
I looked at the cache hit path first since that was the concern I raised. Each loaded plugin instance now hashes its script at most once, lazily, so the plugin lookup, the script encoding, and the SHA-256 are gone from the hit path, and a hit costs an await on an already resolved promise plus a map lookup. A plugin that never touches a gated API pays no hashing at all, the persistent key format is unchanged, and an unknown permission string still returns false uncached.

The capture also fixes something I had not raised. The getter closes over `plugin.script` at API construction, the same value `host.run` executes, so decisions are keyed to the code that is actually running. On main, a permission check that lands during a plugin reload hashes the database-current script, and for up to a second a stale instance can read and write decisions under the updated script's hash. That window is gone here.

`pnpm check`, `pnpm test`, and `pnpm build` are clean at 016d423b, with the new concurrency test running in the default suite, and the branch merges clean against current main.

SameDesu123:
Thanks for the thorough follow-up and approval. While re-checking the provider invocation path mentioned in the review, I found a pre-existing enforcement gap: the `provider` permission result was awaited but its boolean value was ignored, so a denied provider callback still ran.

Because this PR is specifically correcting per-permission decisions and already owns that provider gate, I kept the fix here as a separate commit, `27ea429f`, rather than opening an unrelated follow-up. The provider callback now runs only after permission is granted; denial returns a failed provider response. Allowed requests are unchanged.

I added focused allowed/denied tests, including an assertion that the protected callback is not invoked after denial. Validation passed:

- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/ts/plugins/apiV3/pluginPermissionCache.test.ts` (7 tests)
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test` (235 passed, 3 skipped)
- `pnpm check`

This commit was added after your approval, so please treat it as a new review surface.

cubicj (approved):
I went through 27ea429f. The discarded result was something I had noticed while mapping the gates last round — the provider wrapper was the only one of the six permission types that didn't act on its decision, so this closes a pre-existing gap. Keeping the fix here makes sense, since the PR already owns that gate.

I traced the denied path end to end. `requestPlugin` is the sole consumer of `pluginV2.providers`, and its failure branch turns the denied result into a normal failed request. The denial is stored in the session cache only, so retries fail fast without another dialog and a fresh app session asks again. The allowed path is unchanged, and the per-instance hash capture and key format from the last round survive untouched.

`pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, and it merges clean against current main.

PR #1574, opened 2026-07-30
fix: keep <style> blocks whose CSS contains markup-like text after resanitization

G4tsby:
## PR Checklist

- Required Checks

## Summary

The resanitization pass added in #1569 (`cbacd7b`) silently deletes `<style>` blocks whose CSS contains `<` followed by a word character, breaking common CSS in both chat messages and `backgroundHTML` — most notably svg data-uri backgrounds and `content` strings with markup-like text. This PR keeps those styles working while preserving the sanitizer-bypass fix from #1569.

## Related Issues

- #1569 introduced the resanitization pass this PR reworks.

## Changes

Root cause: after the first sanitize + `decodeStyle`, #1569 runs a second `DOMPurify.sanitize` over the decoded `<style>` tags. DOMPurify's `SAFE_FOR_XML` check removes any node whose text content matches `/<[/\w!]/`, so a decoded style whose CSS contains `<`+word is dropped entirely.

The fix stops sending processed CSS back through the HTML sanitizer as `<style>` text. Instead the decoded CSS is held in a call-scoped array and only an index marker (a reserved PUA character pair, same range `risuEscape` already uses) passes through the second sanitize; the marker is expanded back into a `<style>` tag afterwards.

- `decodeStyle` is split into `decodeStyleContent(hex)`, returning the scoped CSS **body** (or the `CSS ERROR:` / empty fallback) rather than a `<style>…</style>` string.
- `trimMarkdown` parses the first-pass output with `DOMParser` and only replaces actual `<risu-style>` **elements** with index markers, after stripping any stray marker characters from every text node and attribute value. So restored `<style>` markup can only ever land in a text-node context, and only CSS that went through `decodeStyleRule` in the same call can be restored.
- The closing-tag escape is narrowed to `</style` only (`<\/style`, still valid CSS since `\/` is `/` in a string), leaving unrelated CSS like `content: "</div>"` untouched.
- No-style messages take a string fast path and never reparse.

Two scoping/injection paths this closes that a purely string-based restore would leave open:

- A literal `</style>` inside a CSS string splits the block in `encodeStyle`, so an attacker-authored `<risu-style>` could fall outside the block, skip `decodeStyleRule`, and be restored as unscoped global CSS. Restoring only from the parsed elements makes that impossible.
- A `<risu-style>` (or a forged marker) surviving inside an **attribute value** — e.g. `<div title='<style>a::after{content:">"}</style>'>` — is not an element, so it is never turned into a marker and stays inert; a string-based pass would have restored `<style>…` there, let the CSS quote close the attribute, and inject markup after the final sanitize (XSS).

Added regression tests (`src/ts/parser/tests/trimMarkdownStyle.test.ts`, DOM-parsed assertions): scoped decoding, svg data-uri CSS, markup-like `content`, the `</style` escape, `</div>` left intact, sanitizer-bypass attempts smuggled through `<risu-style>` hex, the nested-`<risu-style>` scoping bypass, the two attribute-context XSS reproductions, and forged index markers in the input — all in both `normal` and `back` modes where relevant. Every new test fails against the pre-fix implementation and passes after.

## Impact

Reproduction on current `main` (`ParseMarkdown(..., 'back')`, same path as chat messages):

```
input:  <style>.bg { background-image: url("data:image/svg+xml,<svg ...><circle r='4'/></svg>"); }</style><div class="bg">x</div>
before: <div class="x-risu-bg">x</div>                          (style block silently removed)
after:  <style>.chattext .x-risu-bg{background-image:url("data:image/svg+xml,<svg ...>...")}</style><div class="x-risu-bg">x</div>
```

- Fixes the "styles stopped applying" regression reported since the #1569 deploy for any character/module/background CSS containing `<` (svg data-uris, `content: "<...>"`, comments), without reopening the sanitizer bypass #1569 fixed.
- No behavior change when the input contains no styles (the fast path returns the first-pass result unchanged).

## Additional Notes

`vitest` (243 passed) and `svelte-check` (0 errors) are green. `DOMParser` is already used throughout this codebase (`characters.ts`, `translator.ts`, …), so it adds no new environment requirement.

### DIFF
--- a/src/ts/parser/parser.svelte.ts
+++ b/src/ts/parser/parser.svelte.ts
@@ -776,26 +776,53 @@ export async function ParseMarkdown(
     return trimMarkdown(data)
 }
 
+const trimPurifyConfig = {
+    ADD_TAGS: ["iframe", "style", "risu-style", "x-em", 'annotation', 'semantics', 'mrow', 'mi', 'mo', 'mn', 'msup', 'msub', 'mfrac', 'msqrt'],
+    ADD_ATTR: ["allow", "allowfullscreen", "frameborder", "scrolling", "risu-ctrl" ,"risu-btn", 'risu-trigger', 'risu-mark', 'risu-id', 'x-hl-lang', 'x-hl-text'],
+}
+
 export function trimMarkdown(data:string){
-    let sant = DOMPurify.sanitize(data, {
-        ADD_TAGS: ["iframe", "style", "risu-style", "x-em", 'annotation', 'semantics', 'mrow', 'mi', 'mo', 'mn', 'msup', 'msub', 'mfrac', 'msqrt'],
-        ADD_ATTR: ["allow", "allowfullscreen", "frameborder", "scrolling", "risu-ctrl" ,"risu-btn", 'risu-trigger', 'risu-mark', 'risu-id', 'x-hl-lang', 'x-hl-text'],
-    })
-
-    const decoded = decodeStyle(sant)
-
-    if(decoded !== sant){
-        sant = DOMPurify.sanitize(decoded, {
-            ADD_TAGS: ["iframe", "style", "risu-style", "x-em", 'annotation', 'semantics', 'mrow', 'mi', 'mo', 'mn', 'msup', 'msub', 'mfrac', 'msqrt'],
-            ADD_ATTR: ["allow", "allowfullscreen", "frameborder", "scrolling", "risu-ctrl" ,"risu-btn", 'risu-trigger', 'risu-mark', 'risu-id', 'x-hl-lang', 'x-hl-text'],
-            FORCE_BODY: true
-        })
-    }
-    else{
-        sant = decoded
+    // Without a <risu-style> there is nothing to decode, so the plain string
+    // result is already final. Note the tag itself is not trusted input:
+    // risu-style is in ADD_TAGS, so cards, model output and user scripts can
+    // author one directly. Only its position in the parsed tree is relied on.
+    if(!data.includes('<risu-style')){
+        return DOMPurify.sanitize(data, trimPurifyConfig)
     }
 
-    return sant
+    // Decoded CSS must never be handed back to the HTML sanitizer as <style>
+    // text: DOMPurify drops style nodes whose CSS contains '<' (SAFE_FOR_XML),
+    // which silently deletes svg data-uris and markup-like content strings.
+    // Take the DOM out of the sanitize we already run and swap the elements in
+    // place instead, so the CSS never re-enters the HTML parser at all.
+    // RETURN_DOM hands back the sanitized <body>, typed as Node. It is null only
+    // for empty input, which the guard above already excludes.
+    const root = DOMPurify.sanitize(data, {
+        ...trimPurifyConfig,
+        RETURN_DOM: true,
+    }) as HTMLElement | null
+    if(!root){
+        return ''
+    }
+
+    // Only real <risu-style> elements are decoded. A <risu-style> that survived
+    // inside an attribute value is not an element, so it is left as inert text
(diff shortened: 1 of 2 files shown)


G4tsby:
@kwaroran Sorry for the direct ping — flagging this for urgent review because it's a **regression currently affecting many users on the live web deploy (main)**. Reports of "styles/backgrounds not applying" are already coming in across the community in volume.

Since the #1569 resanitization landed (2026-07-29), DOMPurify's `SAFE_FOR_XML` check **silently removes any `<style>` block whose CSS text contains `<` followed by a word character**. That covers svg data-uri backgrounds, `content: "<...>"` strings, and CSS comments containing `<` — all standard patterns in styled character cards and backgrounds, so a wide range of users are hit immediately. Chat messages and `backgroundHTML` go through the same path (`trimMarkdown`), so both break, and because the failure is silent (no error, the node is just dropped) neither users nor card authors can trace the cause.

This is live on main right now. Desktop v2026.6.215 is not affected (its tag predates the parser changes, at 846c897), but the **next desktop release will pick it up as-is**. Since the web build is already exposed, it'd be much appreciated if this could be treated as a hotfix.

This PR keeps the sanitizer-bypass fix from #1569 intact and only restores the styles: instead of sending decoded CSS back through the sanitizer as a `<style>` string, it holds the CSS in a call-scoped array, replaces only the actual `<risu-style>` elements in the parsed DOM with index markers, and expands them back afterwards.

While iterating I also closed three related issues on this branch:
- a scoping bypass where a literal `</style>` inside a CSS string could inject unscoped global CSS
- an XSS where a marker landing in an attribute-value context let a string-based restore break out of the attribute and inject markup
- a mismatch between the fast-path guard and the decode query

Each case is pinned by a regression test that fails against the vulnerable implementation (`src/ts/parser/tests/trimMarkdownStyle.test.ts`, DOM-parsed assertions). vitest 243 passed, svelte-check 0 errors, CI green.

Happy to adjust the approach if you'd prefer to handle it differently.

kwaroran:
Although this PR solves the problem, it seems quite over-engineered. It modifies lots of nodes and makes fairly resource-heavy changes to them. If this function weren't called so often, it might be acceptable, but considering its high call frequency, I don't think this PR is suitable in current form

G4tsby:
You're right, and thanks for the push-back.

The cause is that I decided to keep #1569's two-sanitize structure so I'd touch your fix as little as possible. That forced style restoration to happen after the final sanitize, which opened a couple of secondary holes (forged markers, attribute-context restore) and the defenses for those are what made this heavy, including the DOM walk you're reacting to. I got stuck working around a constraint I should have questioned instead.

So I'd like to drop the second sanitize and take the DOM out of the first one (`RETURN_DOM: true`), replacing `risu-style` elements with `<style>` in place.

I prototyped it: it passes this branch's tests (minus the one that only exists for the marker design), and on the style path it comes out faster than main, not just faster than this PR.

Does that direction look acceptable to you?

kwaroran:
It seems like a good direction, but I might I need to review the code directly. Could you please update this PR or create a new one?

G4tsby:
Sure. I'll push a commit reflecting the changes to this PR around 7pm KST.

G4tsby:
Reworked the branch as described above — second sanitize gone, `RETURN_DOM` from the first one, `<risu-style>` elements swapped for `<style>` in place. No reparse, no markers, no tree walk; the only hardening left is the `</style` escape.

One thing I left out of scope while checking it: `trimMarkdown` isn't idempotent for CSS containing `<`, and the default auto-translate path runs it twice (`ParseMarkdown` with `'pretranslate'` trims, then the render template trims again), so those styles are still dropped for auto-translate users. That predates this PR and a proper fix has to touch the translator, so I've left it alone rather than widening this again.

PR #1575, opened 2026-07-30
feat(apiV3): add allow-clipboard-write to iframe sandbox

nevaeh5379:
Added missing semicolons and `allow-clipboard-write` permission to sandbox to enable clipboard access in sandboxed iframe.

## PR Checklist

- Required Checks

## Summary

I added the relevant code to fix the issue where clipboard copying was not possible in the plugin because `allow-clipboard-write` was not included in the iframe sandbox allowlist.

## Related Issues

None

## Changes

By adding the `this.iframe.sandbox.add('allow-clipboard-write');` code, we enabled clipboard copying within the plugin.

### DIFF
--- a/src/ts/plugins/apiV3/factory.ts
+++ b/src/ts/plugins/apiV3/factory.ts
@@ -782,9 +782,10 @@ export class SandboxHost {
         this.iframe.setAttribute('allowTransparency', 'true');
 
         this.iframe.sandbox.add('allow-scripts');
-        this.iframe.sandbox.add('allow-modals')
-        this.iframe.sandbox.add('allow-downloads')
-
+        this.iframe.sandbox.add('allow-modals');
+        this.iframe.sandbox.add('allow-downloads');
+        this.iframe.sandbox.add('allow-clipboard-write');
+        
         this.iframe.setAttribute('csp', this.csp);
 
         const messageHandler = async (event: MessageEvent) => {


cubicj (commented):
I went through 99ff809d7 on a Vite dev server first, since its only intended behavior change is the added sandbox token. `allow-clipboard-write` is not in the HTML spec's sandbox keyword set, so browsers ignore it. On this head `iframe.sandbox.supports('allow-clipboard-write')` returns false, Chromium logs an invalid sandbox flag error each time a plugin iframe is created, and `navigator.clipboard.writeText` from the plugin sandbox still rejects with NotAllowedError. That error log ships to production too, since it comes from the browser's attribute parser and not from a `console` call the build could strip. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, but none of them validates whether the browser recognizes the token or enables clipboard writes.

The write is blocked by two separate checks that the added token does not control. For this sandboxed plugin iframe, clipboard access has to be delegated through the permissions policy `allow` attribute, which nothing currently sets, and Chromium also requires the calling document to be focused, which cannot happen while the iframe stays hidden. In my testing the write went through once the iframe carried a `clipboard-write` delegation in `allow`, was made visible with the same display change as `showContainer`, and the call ran from a real click inside the frame. The hidden case and the visible case without `allow` both failed, for focus and permissions-policy reasons respectively.

Today the token has no effect, but that can change if a browser later defines it as a sandbox keyword. Every V3 plugin iframe created through `SandboxHost` would then opt out of that restriction automatically.

Could you take another look at the change with these results in mind?

PR #1577, opened 2026-08-01
feat(display): add option to allow textarea resizing

nevaeh5379:
## PR Checklist

- Required Checks

## Summary

This is a simple PR adding functionality to make the textarea resizable. I added this because I thought it would be convenient to be able to adjust the size of the textarea when editing prompts.

## Related Issues

None

## Changes

Added a new variable 'resizeTextarea' to the database.
Added the name 'resizeTextarea' to the en.ts and ko.ts files.
Added a checkbox to displaySettingsData.svelite.ts to determine whether the textarea should be resizable.
Updated TextAreaInput.svelite so that if the resizeTextarea value is true, the classes 'resize-y', 'shrink-0', and 'overflow-hidden' are added to the parent div.

## Impact

If the relevant option is enabled, users will be able to resize the textarea area.

## Additional Notes
스크린샷_20260801_102002

[스크린캐스트_20260801_102045.webm](https://github.com/user-attachments/assets/5cc30292-9d03-4834-b57a-5f8dd2e93e8f)

### DIFF
--- a/src/etc/resizeIcon.svg
+++ b/src/etc/resizeIcon.svg
@@ -1,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+     width="24"
+     height="24"
+     viewBox="0 0 24 24"
+     fill="none"
+     aria-hidden="true">
+  <path d="M3 21h16a2 2 0 0 0 2-2V3"
+        stroke="currentColor"
+        stroke-width="2"
+        stroke-linecap="round"
+        stroke-linejoin="round"/>
+</svg>
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -1707,6 +1707,7 @@ export const languageEnglish = {
     convertToModule: "Convert to Module",
     skipSavingAssetsOnWebSync: "Skip Saving Assets on Web Sync",
     applyAdditionalParamsToAll: "Apply Additional Parameters to All Models",
+    resizeTextarea: "Allow Textarea Resizing"
 } satisfies I18nTranslation;
 
 type I18nTranslationFunction = (...args: any[]) => string;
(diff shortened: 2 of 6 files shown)


sub-hub:
It's a simple and good patch. In my environment (Windows Chrome), resizing works well.

However, there is one problem. Have you tried running it on a mobile device? The resize button is so small on mobile that it's practically invisible, making it difficult to press. Lol.

I know it's not your fault and that you used Tailwind's built-in resizeable, but the size is just too small.
image

 I'm curious if this is the case for others on mobile as well.
Also, I would appreciate it if someone could test the mobile and PC Safari environments.

nevaeh5379:
@sub-hub
I have observed that the HTML native resize is too small on mobile. To resolve this issue, I replaced it with a custom resize button.

I tested the related functionality briefly in Safari on macOS, and it doesn't seem to have any problems.

sub-hub:
Thanks for switching to the custom handle, it's now much easier to tap on mobile.

One last suggestion: in light themes, `darkborderc` doesn't have enough contrast, so the icon is barely visible. What do you think about using `var(--risu-theme-textcolor2)` instead? It's already used for secondary text/icons across the app and is readable in both light and dark themes.

```diff
- style:color="var(--risu-theme-darkborderc)"
+ style:color="var(--risu-theme-textcolor2)"
```

Before
image
image


After
image
image

nevaeh5379:
@sub-hub 
image
Considering the mobile environment, that approach is quite good. However, I have one concern.
As seen in the screenshots I sent, the resize icons in some themes are too vivid and distracting.
I will look for a solution, such as setting the color to --risu-theme-textcolor2 and adjusting it with opacity.

PR #1579, opened 2026-08-03
Fix: show asset style setting

nevaeh5379:
## PR Checklist

- Required Checks

## Summary

Since the Insert Asset Prompt option and the Asset Style option are separate, I don't think the Asset Style option needs to be hidden when the Insert Asset Prompt is disabled.

## Related Issues

None

## Changes

Remove {#if DBState.db.characters[$selectedCharID].prebuiltAssetCommand}.

### DIFF
--- a/src/lib/SideBars/CharConfig.svelte
+++ b/src/lib/SideBars/CharConfig.svelte
@@ -581,14 +581,11 @@
             {#if DBState.db.newImageHandlingBeta}
             <CheckInput bind:check={DBState.db.characters[$selectedCharID].prebuiltAssetCommand} name={language.insertAssetPrompt}/>
 
-            {#if DBState.db.characters[$selectedCharID].prebuiltAssetCommand}
-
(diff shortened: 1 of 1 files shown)


cubicj (approved):
I went through e524547a3 on this. `prebuiltAssetStyle` only drives the image handling path in `ChatBody`, which runs with the new image handling beta on and never checks `prebuiltAssetCommand`. Insert Asset Prompt gates the inserted instruction and the `{{chardisplayasset}}` output, not that path. Asset images can still show up with the option off, from a card's own image tags or from messages sent while it was on, and the style applies to them either way, so hiding the select did not match how the setting behaves.

The select stays inside the `newImageHandlingBeta` gate, the same flag the display path checks, and the exclude buttons in this tab were already independent of the checkbox. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

sub-hub (approved):

PR #1582, opened 2026-08-04
fix(backup): include non-png assets in local backup and drive sync

sub-hub:
## PR Checklist

- Required Checks

## Summary
The existing backup only included PNG files, but this PR removes that restriction. Instead, Tauri only checks the isDirectory in case it is a folder.


cubicj (commented):
I went through ee87384de on this. `saveAsset` keeps the extension it is given, so non-png assets exist on every backend, and both restore paths were already extension-agnostic, so the save-side filters were the only thing dropping them. The directory skip on Tauri and the `assets/` guard in the full web loops match what those loops iterate, and Drive restore stops spending three one-second retries on every non-png reference it could never find. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

One change I would like to ask for. The web loop in `SavePartialLocalBackup` can now report preset icons as missing assets and, in account mode, stop the backup with the wait dialog still up. Its keys are database references rather than a storage listing, and the only guard left is the empty-key check, so any value that is not a storage key reaches `forageStorage.getItem` directly. That input is ordinary data, since the preset icon UI stores a jpeg data URL straight into `botPresets[].image` and this map picks it up, and imported preset JSON keeps arbitrary fields as well. On main the suffix check happened to skip these. I would like the `startsWith('assets/')` guard the other two web loops got here too, unless the curated set makes it unnecessary in a way I am missing.

sub-hub:
@cubicj You're right, I misunderstood how preset images work. I assumed they'd be stored as normal assets like character images, but I didn't realize the preset icon UI writes a base64 data URL straight into `botPresets[].image`, so it's embedded inline in the database rather than referenced as an `assets/` file. That also explains why `getUncleanablesSync` never treats it as a server asset.

I've applied the `startsWith('assets/')` guard to the `SavePartialLocalBackup` so non-storage keys (like that data URL) are skipped, matching the other web loops.

cubicj (approved):
I went through 1b8fc9975. The `assets/` guard runs before both per-key storage reads, so the preset data URLs never reach `forageStorage.getItem` or the missing-asset report, and the other PR hunks are unchanged from ee87384de. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

Approving — thanks for the careful follow-up!

PR #1583, opened 2026-08-05
fix(hypav3): derive summary progress from all linked chats

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This fixes an issue in HypaMemory V3's contextual embedding path where chunks with the same text could end up sharing the wrong vector.

Contextual embedding models can return different vectors for the same sentence depending on the surrounding summary. However, the results were temporarily stored using only the chunk text as the key. Because of that, when the same sentence appeared in multiple summaries, the later result could overwrite the earlier one.

The vectors are now tied directly to the chunk objects they were generated for, so identical text from different contexts remains separate.

## Related Issues

None.

## Changes

The temporary vector mapping now uses the actual `SummaryChunk` instead of its text.

I also updated the persistent cache key to include the chunk's position within the context. This handles another similar case where the same sentence appears more than once inside a single summary.

The cache key format has been versioned as `chunk-cache-v2`, so older entries created with the ambiguous format will not be reused. They will be regenerated normally when needed.

There is also an additional check for embedding provider responses. If the provider returns a different number of vectors than the number of input chunks, the operation now fails with an explicit error instead of silently connecting the wrong vectors.

Regression tests were added for duplicate text across different summaries, repeated text within one summary, context separation, invalid indices, and mismatched embedding response lengths.

## Impact

This should improve memory retrieval accuracy when using contextual embedding providers such as Voyage Context 3.

Previously, a chunk could receive a vector generated for the same text in a different context. That could change its similarity score and, in some cases, cause HypaMemory V3 to retrieve an unrelated summary.

The non-contextual embedding paths are not changed by this PR.

Existing contextual cache entries are not deleted. Since the new cache keys no longer match the previous format, affected embeddings will simply be regenerated when they are used again. The old entries may remain until the vector cache is cleared through the existing cleanup process.

The change is limited to shared TypeScript logic and does not add separate behavior for web, local, or node-hosted environments.

## Additional Notes

The regression tests are located in:

`src/ts/process/memory/contextualEmbeddingCache.test.ts`

### DIFF
--- a/src/ts/process/memory/contextualEmbeddingCache.test.ts
+++ b/src/ts/process/memory/contextualEmbeddingCache.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+import {
+    createContextualChunkCacheKey,
+    mapContextualChunkEmbeddings,
+} from "./contextualEmbeddingCache";
+
+describe("mapContextualChunkEmbeddings", () => {
+    it("keeps identical text associated with distinct chunk objects", () => {
+        const firstChunk = { text: "He closed the door.", summaryId: "first" };
+        const secondChunk = { text: "He closed the door.", summaryId: "second" };
+
+        const vectors = mapContextualChunkEmbeddings(
+            [firstChunk, secondChunk],
+            [[1, 0], [0, 1]]
+        );
+
+        expect(vectors.get(firstChunk)?.embedding).toEqual([1, 0]);
+        expect(vectors.get(secondChunk)?.embedding).toEqual([0, 1]);
+        expect(vectors.size).toBe(2);
+    });
+
+    it("rejects provider responses with a mismatched embedding count", () => {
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I went through b246f4c9e on this. The text-keyed maps in that path came from my Voyage Context 3 work, so thanks for catching this. I ran the old and new method bodies side by side against a mock provider. With the same sentence in two summaries, main hands one of the chunks the other summary's vector, and with a sentence repeated inside one summary both chunks come back with the last position's vector on the cold pass and again from the cache on the next request. On the new head each chunk keeps its own vector in both cases and the second request serves them from the cache without another provider call.

The count check also closes a worse path than the description suggests. When the provider comes back short, main stores an `undefined` embedding in the cache, and later requests that hit that cached group fail in `similarity` instead of embedding it again. The new head throws before writing anything for that group, so the next request simply retries.

PR #1587, opened 2026-08-08
perf: avoid full database snapshot for plugin storage reads

SameDesu123:
## PR Checklist

- Required Checks

## Summary

`pluginStorage.getItem()` was snapshotting the entire database before reading a single value from `pluginCustomStorage`. This changes the read path so only the requested value is snapshotted.

For users with large databases, plugin storage reads no longer need to copy unrelated characters, chats, presets, and other database state just to return one plugin value.

## Related Issues

None

## Changes

- Read the requested `pluginCustomStorage` key directly from the live database.
- Apply `$state.snapshot()` only to the requested value instead of the complete database.
- Keep the existing return behavior for falsy or missing values.
- Keep object and array results detached from the live database state, so plugins do not receive the database proxy itself.

## Impact

There is no database format change and no intended plugin API behavior change. The main difference is the amount of data copied during `pluginStorage.getItem()`: the work now scales with the requested stored value instead of the size of the whole database.

This should mainly benefit long-term users with large character and chat histories, especially when plugins read their storage frequently.

## Additional Notes

This does not change model prompting or response handling. The type-definition checklist item is not applicable because the public API signature is unchanged.

Tested with dummy plugin. [plugin-storage-value-snapshot-test.js](https://github.com/user-attachments/files/30848058/plugin-storage-value-snapshot-test.js)

### DIFF
--- a/src/ts/plugins/plugins.svelte.ts
+++ b/src/ts/plugins/plugins.svelte.ts
@@ -727,9 +727,8 @@ export const getV2PluginAPIs = () => {
         },
         pluginStorage: {
             getItem: (key: string) => {
-                const db = getDatabase({ snapshot: true });
-                db.pluginCustomStorage ??= {}
-                return db.pluginCustomStorage[key] || null;
+                const value = getDatabase().pluginCustomStorage?.[key];
+                return value ? $state.snapshot(value) : null;
             },
             setItem: (key: string, value: string) => {
                 const db = getDatabase();


cubicj (approved):
I went through d36fccc6a on this. I ran the old and new `getItem` side by side against the same seeded database and the semantics hold. Falsy and missing values still come back as `null`, plain object and array values still arrive detached from the live state, and the dropped `??=` initialization only ever wrote to the snapshot copy, so removing it changes nothing observable.

With a large synthetic database the old path spent hundreds of milliseconds per read cloning the whole database, and the value-scoped snapshot also stops each read from walking unrelated fields, so getters and `toJSON` on data the plugin never asked for no longer run.

PR #1590, opened 2026-08-08
perf: avoid cloning full chat history for reroll snapshots

SameDesu123:
## PR Checklist

- Required Checks

## Summary

This changes the reroll snapshot path so it slices out only the newly appended messages before deep-cloning them.

Previously, a successful generation cloned the entire current chat and then discarded everything before `previousLength`. On long-running chats, that meant reroll bookkeeping could temporarily copy a large amount of old message data even though only the new messages were kept.

## Related Issues

None

## Changes

Changed the reroll snapshot from `clone(all messages) -> slice(new messages)` to `slice(new messages) -> clone(new messages)`.

The stored reroll data is unchanged in shape and still contains the messages appended after `previousLength`; the difference is that older messages are no longer deep-cloned just to be discarded immediately afterward.

## Impact

This should reduce temporary CPU and memory work when generating in long chats, especially when the chat contains many messages or messages with larger nested metadata.

There is no save-format change and no intended change to reroll behavior. Short chats should see little practical difference.

## Additional Notes

This is intentionally kept as a small, isolated optimization rather than introducing a new helper or changing the broader reroll flow.

Validation: `pnpm check` (0 errors and 0 warnings), plus `git diff --check`.

### DIFF
--- a/src/lib/ChatScreens/DefaultChatScreen.svelte
+++ b/src/lib/ChatScreens/DefaultChatScreen.svelte
@@ -313,7 +313,7 @@
                 continue:continued
             })
             if(previousLength < DBState.db.characters[$selectedCharID].chats[DBState.db.characters[$selectedCharID].chatPage].message.length){
-                rerolls.push(safeStructuredClone(DBState.db.characters[$selectedCharID].chats[DBState.db.characters[$selectedCharID].chatPage].message).slice(previousLength))
+                rerolls.push(safeStructuredClone(DBState.db.characters[$selectedCharID].chats[DBState.db.characters[$selectedCharID].chatPage].message.slice(previousLength)))
                 rerollid = rerolls.length - 1
             }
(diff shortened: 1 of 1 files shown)


cubicj (approved):
I went through 0e56b763c on this. Both shapes take the same clone path, since `structuredClone` rejects the `$state` proxies whether they sit at the array root or in its elements, so `safeStructuredClone` falls back to rfdc on both sides. I ran the old and new expression side by side against real proxied chats and the stored entries come out identical and detached from the live chat state, and `reroll` and `unReroll` clone the entry again before writing it back, so nothing downstream can tell the difference.

On a synthetic 4000-message chat the old path spent over a hundred milliseconds per successful generation cloning history it was about to discard, while the new one stays under a tenth of a millisecond. It also stops the bookkeeping from walking every old message's nested metadata through the proxy on each generation.

PR #1591, opened 2026-08-08
perf: optimize checkCharOrder lookups with Set

SameDesu123:
## PR Checklist

- Required Checks

## Summary

`checkCharOrder()` currently uses repeated `Array.includes()` calls while matching character IDs against the current order. This is fine for small collections, but the repeated linear scans become unnecessary work as the character list grows.

This PR changes those temporary membership lists to `Set`s so the same checks can use `Set.has()` instead.

## Related Issues

None

## Changes

The temporary `ordered` and character ID lookup collections in `checkCharOrder()` are now `Set<string>` instances. The existing ordering, folder cleanup, trash handling, and special handling for `§temp` / `§playground` are left unchanged.

The current behavior where IDs appended to `characterOrder` during the function are not added back to the initial `ordered` lookup set is also preserved. This keeps the change focused on lookup cost rather than changing any edge-case behavior.

## Impact

There are no database format or API changes. Normal users should see no behavioral difference. The main benefit is avoiding repeated linear membership scans for users with larger character collections or folder structures.

## Additional Notes

The original and updated logic were compared over 50,000 randomized input cases, including duplicate IDs, trashed characters, folders, empty/null order entries, and the special temporary character IDs. The resulting database structure matched in every case.

`pnpm check` completes with no errors or warnings. No model/request behavior is touched by this PR, and no new type definitions are required.

### DIFF
--- a/src/ts/globalApi.svelte.ts
+++ b/src/ts/globalApi.svelte.ts
@@ -1124,28 +1124,28 @@ export function replaceDbResources(db: Database, replacer: { [key: string]: stri
  */
 export function checkCharOrder() {
     DBState.db.characterOrder = DBState.db.characterOrder ?? []
-    let ordered = []
+    const ordered = new Set<string>()
     for (let i = 0; i < DBState.db.characterOrder.length; i++) {
         const folder = DBState.db.characterOrder[i]
         if (typeof (folder) !== 'string' && folder) {
             for (const f of folder.data) {
-                ordered.push(f)
+                ordered.add(f)
             }
         }
         if (typeof (folder) === 'string') {
-            ordered.push(folder)
+            ordered.add(folder)
         }
     }
 
-    let charIdList: string[] = []
+    const charIdSet = new Set<string>()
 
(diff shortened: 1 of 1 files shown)


cubicj (approved):
I went through f14b4b369 on this. Both converted collections are local to the function and only ever answer membership checks, so the swap cannot change what gets pushed or spliced. I ran the old and new function side by side over 50,000 randomized inputs with duplicate ids, folders, trashed characters, and null order entries, and the resulting databases matched in every case. The edge case the PR body calls out holds as well. Neither side adds IDs appended during the scan back to the initial lookup, so two untrashed characters sharing an id still push it twice on both sides.

The lookup cost drops as described. On small lists the Set allocation makes it a wash, and the gain shows up once character counts grow, which is where the per-file calls in the import loop were paying the repeated linear scans. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

sub-hub (approved):

PR #1592, opened 2026-08-09
feat: add live WCAG contrast feedback to custom color schemes

SameDesu123:
## PR Checklist

- Required Checks

## Summary

Adds live text readability feedback to the custom color scheme editor using WCAG contrast ratios.

Instead of showing only WCAG terms, the editor uses simpler ratings such as `Excellent`, `Good`, `Limited`, and `Poor` together with the actual contrast ratio and AA/AAA level, so the result should still be understandable without prior knowledge of WCAG.

## Related Issues

None

## Changes

- Added WCAG relative luminance and contrast ratio calculation helpers.
- Added tests for contrast calculation, invalid colors, and rating thresholds.
- Added a `Text Readability` section to custom color schemes.
- Checks the following combinations in real time:
  - Text Color / Background
  - Text Color / Dark Background
  - Text Color 2 / Background
  - Text Color 2 / Dark Background
- Shows an `Aa` preview using the actual foreground and background colors.
- Uses simple readability ratings alongside the WCAG result:
  - `Excellent (AAA)`
  - `Good (AA)`
  - `Limited (Large text only)`
  - `Poor (Low contrast)`
- Added a short explanation that the values are calculated using WCAG contrast guidelines.
- Added translations for all currently supported languages: English, Korean, Simplified Chinese, Traditional Chinese, German, Vietnamese, and Spanish.
- Keeps the import/export actions directly below the color palette, before the readability information.
- Improved the layout for narrow containers. The contrast ratio stays on the first row, while the readability rating moves to a second right-aligned row when there is not enough horizontal space.

## Impact

This only adds informational feedback to the custom color editor.

It does not modify the selected colors automatically, block low-contrast combinations, change saved color schemes, or affect existing presets.

The calculation is local and only evaluates four color pairs while the custom color editor is visible, so the runtime cost should be negligible.

The responsive layout is based on the width of the contrast section itself rather than the viewport, so it also behaves correctly when the settings panel is narrow on desktop.

## Additional Notes

| Preview |
|---------|
| cnwoofiwnfiqjehsickqpxief |

The simpler readability labels are intentionally shown alongside WCAG levels instead of replacing them. This keeps the UI approachable while still exposing the standardized result for users who are familiar with WCAG.

`AA` and `AAA` are left untranslated because they are WCAG conformance level names.


kwaroran:
Personally, I’m against adding this feature or similar ones. I believe GUI should be 'keep default information minimal, and only expose more details in advanced settings or if the user explicitly opts in.' While these kind of feature might be nice for some users, I'm concerned that adding features like this will make the UI overly complex, by adding features that most users doesn't care much

SameDesu123:
That makes sense. Then I'll revise this so the feature is opt-in, with a toggle under the "Other" tab in Display Settings.

PR #1593, opened 2026-08-09
fix: keep database proxy bookkeeping out of Svelte reaction tracking

### LINKED ISSUE #1586: Infinite loop (effect_update_depth_exceeded) when opening "Other Bots" menu, even on fresh install with no model connected
**Environment**
- RisuAI version: v2026.6.215 (self-built from source, main branch tag)
- Deployment: Docker (self-hosted), Node 24-slim base image
- Browser: Chrome (Windows)

**Steps to reproduce**
1. Fresh install, no data, no API/model connected
2. Click "Other Bots" in the left sidebar
3. The default tab that opens is "Long-term Memory" (HypaMemory V3)

**Expected behavior**
Settings screen opens normally.

**Actual behavior**
The app freezes immediately. Console shows:
Uncaught Error: https://svelte.dev/e/effect_update_depth_exceeded

This happens with:
- No model/API connected
- Any preset (including default) or memory type selected
- Completely fresh browser storage and fresh server volume

This suggests the max memory token ratio calculation for HypaMemory V3 enters an infinite reactive loop when it cannot resolve a max context value, and this loop is not guarded against.

**Additional notes**
The UI also shows: "Unable to calculate max memory token ratio" right before freezing.

cubicj:
## PR Checklist

- Required Checks

## Summary

Fixes the freeze reported in #1586: since #1233, opening the Other Bots screen crashes with `effect_update_depth_exceeded` on every install, including fresh ones.

The database tracking proxy introduced by #1233 performs reactive reads inside its own traps — old/new value reads in `set`, eager child-subtree connection, synchronous update-listener dispatch — and the engine's post-trap proxy invariant checks additionally read the Svelte `$state` target. Any `$effect` that writes through `DBState.db` therefore subscribes itself to the property it writes. The wavespeed LoRA sync effect writes a fresh array on every run, so it never converges and hits Svelte's 100-iteration flush guard. The same mechanism also caused quieter bugs: settings-writing effects inherited dependencies from the save change tracker (a character switch could rerun the NanoGPT reset effect and clear `nanogptRequestModel`), descriptor-based writes mutated state without triggering saves, and the first read of a retained proxy inside a `$derived` threw `state_unsafe_mutation`.

## Related Issues

Fixes #1586. The commits suspected in the issue thread are not on `main`; the trigger effect predates #1233 and was harmless under plain `$state` (verified by running the same reproduction against the pre-#1233 tree).

## Changes

**`src/ts/storage/databaseState.svelte.ts`**
- The tracking proxy now uses a detached, non-reactive shape target, so ECMAScript proxy invariant checks validate against a mirror instead of reading the reactive Svelte target
- All trap bookkeeping (`set`/`deleteProperty` value reads, child reconnection and eager subtree walks, update-listener dispatch, extensibility sync) runs inside `untrack()`; the actual property read in `get` stays tracked, preserving normal reactivity
- The retained-proxy fix-up write is untracked, so first reads inside deriveds/templates no longer throw `state_unsafe_mutation` while still keeping tracking proxies out of persisted state
- `defineProperty` with a plain data descriptor now routes through the normal notified set pathway (it previously mutated state without emitting `onDatabaseUpdate`, so those writes were never saved); other descriptor shapes keep Svelte's own `$state` rules

**`src/ts/storage/databaseState.svelte.test.ts`**
- Regression tests for each fix: the write-loop case, subtree over-tracking (asserting both no-rerun on unrelated nested writes and rerun on actual replacement), descriptor writes emitting assignment-equivalent events, and derived-time retained-proxy reads staying safe and snapshot-clean

## Impact

- Opening Other Bots no longer freezes; the same fix covers every effect that writes a fresh object through `DBState.db`, not just the wavespeed screen
- `onDatabaseUpdate` payloads, paths, and ordering are unchanged, so the incremental-save performance benefit of #1233 is preserved (the existing proxy test suite locks these semantics)
- Legacy V2 plugins keep working unchanged; data-descriptor writes that previously bypassed the save tracker now persist correctly
- No component files changed; no overlap with open #1550 or #1587

## Additional Notes

Validated with:

- `pnpm test` (238 passed / 3 skipped, including 34 storage-proxy tests; each fix was written red first against the unfixed tree)
- `pnpm check` (0 errors, 0 warnings)
- Manual smoke on the dev build: Other Bots opens normally, character switching no longer clears NanoGPT fields, and root- plus character-level edits persist across a reload

### DIFF
--- a/src/ts/storage/databaseState.svelte.test.ts
+++ b/src/ts/storage/databaseState.svelte.test.ts
@@ -48,6 +48,53 @@ describe('database proxy layering', () => {
         unsubscribe()
     })
 
+    it('treats plain data descriptors like assignments and connects retained children', () => {
+        setDatabaseLite(database({ pluginCustomStorage: { retained: { value: 0 } } }))
+        const storage = DBState.db.pluginCustomStorage
+        const retained = storage.retained
+        delete storage.retained
+        const listener = vi.fn()
+        const unsubscribe = onDatabaseUpdate(listener)
+
+        try {
+            storage.defined = retained
+            const assignmentEvent = listener.mock.calls[0][0]
+            delete storage.defined
+            listener.mockClear()
+
+            Object.defineProperty(storage, 'defined', { value: retained })
+
+            expect(listener).toHaveBeenCalledTimes(1)
+            expect(listener.mock.calls[0][0]).toEqual(assignmentEvent)
+            listener.mockClear()
+            retained.value = 1
+            expect(listener).toHaveBeenCalledWith(expect.objectContaining({
+                path: ['pluginCustomStorage', 'defined', 'value'],
+                value: 1,
+                oldValue: 0,
+                type: 'set',
+            }))
+        } finally {
+            unsubscribe()
+        }
+    })
+
+    it('rejects accessor descriptors without changing database state', () => {
+        setDatabaseLite(database({ pluginCustomStorage: { value: 1 } }))
+        const storage = DBState.db.pluginCustomStorage
+        const listener = vi.fn()
+        const unsubscribe = onDatabaseUpdate(listener)
+
+        try {
+            expect(() => Object.defineProperty(storage, 'value', { get: () => 2 }))
+                .toThrowError(/state_descriptors_fixed/)
+            expect(storage.value).toBe(1)
+            expect(listener).not.toHaveBeenCalled()
+        } finally {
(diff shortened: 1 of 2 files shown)


kwaroran:
closing due to merge of #1594  I decided that reverting was better than adding a fix that requires more review, which might delay the update more.

PR #1596, opened 2026-08-12
Update retired Claude model ids in onboarding and preset defaults

kasparovabi:
## PR Checklist

- Required Checks
    - [x] Have you added type definitions? No new types. This is five string literals inside structures that are already typed.
    - [x] Have you tested your changes? `pnpm check` reports 0 errors and 0 warnings. See Additional Notes for what I did not test.
- [x] If your PR uses models:
    - [x] Have you checked if it works normally in all models? The new id is already an entry in `AnthropicModels`, so it resolves the same way any other listed model does.
    - [x] Have you checked if it works normally in all web, local, and node-hosted versions? The touched values are shared constants with no platform branch.
- [x] If your PR is highly AI generated:
    - [x] Is it not a huge change? Five lines.

## Summary

The Claude defaults in onboarding and in the OAI2 prebuilt preset point at model ids that Anthropic retired on 28 October 2025. A new user who picks Claude in the welcome flow gets `claude-3-5-sonnet-20241022` written straight into `DBState.db.aiModel`.

## Related Issues

None.

## Changes

| File | Line | Field | Was | Now |
| --- | --- | --- | --- | --- |
| `src/lib/Others/WelcomeRisu.svelte` | 104 | `aiModel` | `claude-3-5-sonnet-20241022` | `claude-sonnet-4-6` |
| `src/lib/Others/WelcomeRisu.svelte` | 105 | `subModel` | `claude-3-5-sonnet-20241022` | `claude-sonnet-4-6` |
| `src/ts/process/templates/templates.ts` | 246 | `aiModel` | `claude-3-5-sonnet-20240620` | `claude-sonnet-4-6` |
| `src/ts/process/templates/templates.ts` | 301 | `proxyRequestModel` | `claude-3-5-sonnet-20240620` | `claude-sonnet-4-6` |
| `src/ts/process/templates/templates.ts` | 388 | `customProxyRequestModel` | `claude-3-5-sonnet-20240620` | `claude-sonnet-4-6` |

Retirement dates come from Anthropic's deprecations page, https://platform.claude.com/docs/en/about-claude/model-deprecations

I picked `claude-sonnet-4-6` because it is already in your catalogue at `src/ts/model/providers/anthropic.ts:64` with `recommended: true`, and because the values being replaced were Sonnet, so this keeps the tier the same. Swap it for whichever entry you would rather ship as the default and I will amend.

Four places that a careless pass would also have rewritten are left alone.

`src/ts/model/providers/anthropic.ts` holds sixteen retired ids and should keep every one of them. That file is the selectable list, each id is paired with its own display name, and a user whose save points at `claude-3-opus-20240229` needs that entry to stay resolvable.

`src/ts/model/modellist.ts` is the Bedrock catalogue and the same reasoning applies.

`src/ts/cbs.ts:657` mentions `claude-3-opus` inside a docstring example of what `{{model}}` returns.

`templates.ts:302` sets `openrouterRequestModel` to `anthropic/claude-2`. That is OpenRouter's namespace rather than Anthropic's and I have not checked what OpenRouter serves today, so I left it for you.

## Impact

Existing saves are untouched. `setPreset` writes `aiModel`, `proxyRequestModel` and `customProxyRequestModel` only when a preset is loaded, so this reaches new users going through the welcome flow and anyone who loads OAI2 from the preset menu after it lands.

## Additional Notes

I read this diff line by line before opening it, and I am happy to adjust or drop any part of it. What brought me here was a checker I maintain that flags retired Claude model ids, so I am saying that up front rather than leaving you to wonder. I have not run the app against the Anthropic API to reproduce a failed request, so the claim rests on the published retirement dates rather than on a status code I watched come back. What I did run is `pnpm check`, which reports 0 errors and 0 warnings on this branch. If you would rather handle this your own way, or would rather not get this kind of contribution at all, tell me and I will not send another.

### DIFF
--- a/src/lib/Others/WelcomeRisu.svelte
+++ b/src/lib/Others/WelcomeRisu.svelte
@@ -101,8 +101,8 @@
                 }
 
                 if(provider === 'claude'){
-                    DBState.db.aiModel = 'claude-3-5-sonnet-20241022'
-                    DBState.db.subModel = 'claude-3-5-sonnet-20241022'
+                    DBState.db.aiModel = 'claude-sonnet-4-6'
+                    DBState.db.subModel = 'claude-sonnet-4-6'
                 }
 
                 if(provider === 'openai'){
--- a/src/ts/process/templates/templates.ts
+++ b/src/ts/process/templates/templates.ts
@@ -243,7 +243,7 @@ export const prebuiltPresets = {
       "globalNote",
       "authorNote"
     ],
-    "aiModel": "claude-3-5-sonnet-20240620",
+    "aiModel": "claude-sonnet-4-6",
     "subModel": "gemini-3-flash-preview",
     "currentPluginProvider": "",
     "textgenWebUIStreamURL": "",
@@ -298,7 +298,7 @@ export const prebuiltPresets = {
       "stoptokens": "",
       "top_k": 140
     },
-    "proxyRequestModel": "claude-3-5-sonnet-20240620",
+    "proxyRequestModel": "claude-sonnet-4-6",
     "openrouterRequestModel": "anthropic/claude-2",
     "NAISettings": {
       "topK": 12,
@@ -385,7 +385,7 @@ export const prebuiltPresets = {
     "NAIadventure": false,
     "NAIappendName": true,
     "autoSuggestPrompt": "",
-    "customProxyRequestModel": "claude-3-5-sonnet-20240620",
+    "customProxyRequestModel": "claude-sonnet-4-6",
     "reverseProxyOobaArgs": {
       "mode": "instruct"
     },


cubicj (approved):
I went through 5b1885103 on this. Both replaced ids were retired on 2025-10-28 per Anthropic's deprecations page, and `claude-sonnet-4-6` is the replacement that page names for both, so no amendment needed on the pick. The diff is exactly the five values, `setPreset` carries all three preset fields into the DB, and the welcome flow and the quick preset button are the only consumers of OAI2, so the reach matches your Impact section. Leaving the catalogue and the Bedrock list alone was the right call, and I found no other code comparing against the old ids.

Separate from this PR, the default preset still cannot complete a request against `claude-sonnet-4-6`. Anthropic dropped assistant prefill on 4.6 and rejects `temperature` together with `top_p` from 4.5 on, while the default prompt template still ends in a prefill turn and the request builder still sends both parameters. Every active Anthropic entry on current main trips at least one of these with the default prompt, so it is out of this PR's scope and belongs in a separate one.

kasparovabi:
Both halves of the separate issue you raised hold up when I trace them.

On sampling, every Anthropic entry in `src/ts/model/providers/anthropic.ts` declares `parameters: [...ClaudeParameters, ...]`, and `ClaudeParameters` is `['temperature', 'top_k', 'top_p']` (`src/ts/model/types.ts:146`). `applyParameters` in `src/ts/process/request/shared.ts` writes each of those onto the body unless the stored value is the `-1000` sentinel, and the stock values are `temperature: 80` and `top_p: 1` (`src/ts/storage/database.svelte.ts:1998` and `:2019`). A default install therefore sends `temperature: 0.8` and `top_p: 1` in the same request. The only place either one is dropped is the Bedrock branch with thinking on, `src/ts/process/request/anthropic.ts:439-441`, which never runs on the direct API path.

The prefill part is narrower than the flag table suggests. `claude-sonnet-4-6` carries no `LLMFlags.hasPrefill` (`src/ts/model/providers/anthropic.ts:68-74`), and neither do `claude-opus-4-6`, `claude-opus-4-7` or `claude-opus-4-8`. The template gate does not read that flag. `prefillsupported` in `src/ts/cbs.ts:1357-1364` returns `'1'` whenever `db.aiModel.startsWith('claude')`, so the trailing `role: "bot"` block that closes the default template (`src/ts/process/templates/templates.ts:380`) still goes out for all four. Resolving the flag off the model entry instead of prefix-matching the id would fix that and would give `hasPrefill` something to do.

One thing I left out of this PR that belongs to the same family. OAI2 also sets `"customProxyRequestModel": "claude-3-5-sonnet-20240620"` at `src/ts/process/templates/templates.ts:388`, which is a retired id, though it only reaches a request through `requests.ts:209` when the user is on the custom proxy path. I skipped it because a custom proxy may deliberately be serving an old id, and I did not want to change behaviour for people pointing at their own endpoint. Say the word if you would rather have it updated.

I have not run any of this against the API, so the sampling and prefill claims rest on the published restrictions plus the code paths above rather than on a response I have seen.

I can send the sampling and prefill work as its own PR once this one is settled, or leave it with you if you would rather keep it in-house. Either suits me.

sub-hub (approved):

cubicj:
Both traces match what I found, and the PR is merged now, so the follow-ups are yours if you want them. Either shape works for me.

Just one thing to flag on the prefill side. `hasPrefill` is also set on the Bedrock 4.6 entry, the DeepSeek entries, and every Response API variant, so a flag-only check would flip them to `'1'`. And templates.ts:388 is already part of the merged diff.

kasparovabi:
Your caveat holds. `hasPrefill` is on the Bedrock 4.6 entry with the compatibility comment (`src/ts/model/modellist.ts:54`), on all four DeepSeek entries (`:465`, `:477`, `:489`, `:500`), and it gets appended to every Response API variant at `:591` and `:756`. So I gated the flag read on `LLMFormat.Anthropic` or `LLMFormat.AnthropicLegacy` and left everything else on the prefix check it has today. That is #1612, one file, `pnpm run check` clean.

Confirmed on templates.ts:388 as well, the merged diff already has `claude-sonnet-4-6` there.

The sampling half I got wrong, and it is worth correcting here rather than only in the PR. `claude-opus-4-7` and `claude-opus-4-8` declare `parameters: []` (`src/ts/model/providers/anthropic.ts:19` and `:37`), and `applyParameters` receives `arg.modelInfo.parameters` (`src/ts/process/request/anthropic.ts:351-357`), so neither `temperature` nor `top_p` is sent for the two models that would reject them. The entries carrying `ClaudeParameters` are 4.6 and older, which take those fields. My claim that every Anthropic entry declares them was too broad, and there is no sampling PR to send.

PR #1597, opened 2026-08-12
refactor: reuse messageLargePortrait in chat props

nevaeh5379:
## PR Checklist

- Required Checks

## Summary

Modify it so that two operations are combined into one.

## Related Issues

none

## Impact

very, very slight improvement in performance

### DIFF
--- a/src/lib/ChatScreens/Chats.svelte
+++ b/src/lib/ChatScreens/Chats.svelte
@@ -116,7 +116,7 @@
                         unReroll: unReroll,
                         rerollIcon: 'dynamic',
                         character: simpleChar,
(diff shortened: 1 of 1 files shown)


cubicj (approved):
I went through d402b66a0 on this. `messageLargePortrait` is computed from the same expression a few lines up and nothing between the two reads touches its inputs, so the value handed to `Chat` is unchanged. The hash key and the prop now read from the same variable, which is the part I like, since on main the two could drift apart if either expression got edited alone.

The duplication came from my own 723c3da24, so thanks for cleaning up after it. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

sub-hub (approved):

PR #1607, opened 2026-08-25
Allow Lua image generation to set NovelAI orientation

enzi221:
## PR Checklist

- Required Checks

## Summary

Lua `generateImage()` may set its own `orientation`.

To prevent troll scripts draining NAI credits, the script can't modify dimension but only its orientation.

- If w/h set to 1000/500 (landscape)
  - No orientation (default, backward compat) => 1000/500
  - Portrait => 500/1000
  - Landscape => 1000/500

## Related Issues

None

## Changes

Lua `generateImage()` now expects up to 4 parameters.

NAI branch of `stableDiff.ts` calls a new function, `applyImageOrientation()`.

## Impact

None, backward compatible.

### DIFF
--- a/src/ts/process/scriptings.ts
+++ b/src/ts/process/scriptings.ts
@@ -7,7 +7,7 @@ import { get } from "svelte/store";
 import { DBState, ReloadChatPointer, ReloadGUIPointer, selectedCharID } from "../stores.svelte";
 import { alertSelect, alertError, alertInput, alertNormal, alertConfirm } from "../alert";
 import { HypaProcesser } from "./memory/hypamemory";
-import { generateAIImage } from "./stableDiff";
+import { generateAIImage, type ImageGenerationOptions } from "./stableDiff";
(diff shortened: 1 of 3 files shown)


cubicj (approved):
I went through 1dbdede59 on this. `applyImageOrientation` only reorders the two configured edges, so the pixel count a script can request stays exactly what the settings allow, and the Variety+ sigma comes out identical since it is computed from the product. Only the NovelAI branch reads `options`, and the other five `generateAIImage` callers still pass four arguments, so their request bodies are unchanged from main. I also checked the Lua side with wasmoon directly, and a `{orientation='portrait'}` table arrives as a plain object while an omitted or nil fourth argument takes the default path as described.

Just one thing to flag, not blocking. A non-table fourth argument such as a bare `'portrait'` string has no `orientation` field, so it skips the check and generates at the configured orientation without the error string coming back. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

PR #1608, opened 2026-08-25
Fix chat variable test mocks

enzi221:
## PR Checklist

- Required Checks

## Summary

Fix tests exploding caused by missing mocks.

## Related Issues

None.

## Changes

Adds `getCurrentChat()` mocks.

## Impact

CI should pass.

### DIFF
--- a/src/ts/parser/tests/cbs/conditionals.test.ts
+++ b/src/ts/parser/tests/cbs/conditionals.test.ts
@@ -12,6 +12,7 @@ vi.mock(
     ({
       appVer: '1234.5.67',
       getCurrentCharacter: () => ({}),
+      getCurrentChat: () => ({}),
       getDatabase: () => ({}),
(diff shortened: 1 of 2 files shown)


cubicj (approved):
I went through e46e88c6 on this. The four failures come from f4290c6cc routing `getGlobalChatVar` through `getGLChatVar`, which reads `getCurrentChat` from `database.svelte`, while both test files mock that module without the export, so the global variable tests and the `#when` toggle condition tests explode on it.

The empty object keeps the original assertion semantics, since `getGLChatVar` comes back undefined and the lookup falls through to `globalChatVariables`, the path both files' fixtures were built around. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head, with the full suite back to green.

sub-hub (approved):

PR #1610, opened 2026-08-28
fix: make SafeElement.addEventListener element-scoped as documented

poroyo:
## PR Checklist

- Required Checks

## Summary

Contrary to the documented behavior, `SafeElement.addEventListener` has attached every listener to `document`.

As a result, matching events that reached `document` triggered listeners even when they occurred outside the element's scope: a close button's `click` handler ran on every click, while an element-level `keydown` listener received every keystroke across the app, with each delivery crossing the plugin iframe boundary as an RPC.

This PR makes the implementation match the documented contract: listeners attach to their own element.

## Related Issues

None.

## Changes

- `SafeElement.addEventListener` attaches to the wrapped element. `SafeDocument` attaches to the actual `Document` node, so document-wide listening continues to behave exactly as before (including viewport `scroll` events that target the `Document` itself).
- Each registration is tracked as a `{ eventTarget, type, listener, options }` record shared by the per-instance ID map and the reload cleanup array. Removal and reload cleanup detach using the stored record.
  - This also fixes a pre-existing defect: removing with mismatched `type`/`capture` used to no-op natively while the bookkeeping was deleted anyway, leaving an untracked live listener until page refresh.
- Docs: state the element scope, point document-wide use to `getRootDocument()`, and remove the `event.target` access from the example because `target` is not included in the delivered payload.

### Intentional Changes

- Plugins that intentionally use this behavior (most commonly global hotkeys) must **explicitly** migrate to `getRootDocument()` instead of receiving global keyboard events through this indirect behavior (added to the documentation).
- Listeners are bound to a specific DOM node and do not survive the node being replaced (native semantics).

### Improvements

- The most immediately visible misfire is fixed: plugin UI no longer reacts when the user clicks outside the registered element.
- `capture`/`once` apply to the element.
- Element listeners stop receiving app-wide keystrokes. This fixes the unexpectedly broad collection surface created when `keydown` was registered on a non-focusable element.
- Out-of-scope events no longer cross the iframe boundary (an element `mousemove` listener previously produced one RPC per pointer move anywhere in the app)
- Non-bubbling event types now work correctly on elements for the first time.

### Unchanged (pre-existing)

Listener IDs remain usable only on the wrapper instance that issued them.

---

## Considered and Rejected

1. **Keep the `document` attachment, add a containment filter**: `documentElement.contains(document)` is false, so this would silently break SafeDocument viewport `scroll`; native `{once:true}` gets consumed by filtered-out events; non-bubbling whitelist entries (`mouseleave`, element `scroll`, ...) never reach a bubble-phase `document` listener at all, so they would stay broken. This approach would require both a SafeDocument exception and manual emulation of native event behavior.

2. **Deliver `event.target` and let plugins filter**: leaves the keystroke exposure and per-event RPC cost in place, and adds one host-side remote reference per delivered event; the registry-growth problem class addressed in #1536.

## Tested

Verified with four local test plugins on the running app: inside/outside delivery for `click`/`keydown`/`mousemove`; non-bubbling delivery (`mouseleave`, element `scroll`); `{once:true}` consumption; `capture` scoping in both the object and boolean forms; the mismatched-removal orphan (via `getEventListeners`); SafeDocument `click`/`keydown`/`scroll` identical before and after; individual removal and reload cleanup both return listener counts to baseline.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -400,19 +400,25 @@ const rects = await element.getClientRects();
 
 #### Event Listeners
 
-Event listeners have security restrictions and return unique IDs:
+Event listeners have security restrictions and return unique IDs. Listeners receive events that reach the element via normal DOM propagation, while non-bubbling events such as scroll fire only on the element itself:
 
 ```javascript
 // Add event listener - returns ID for later removal
 const listenerId = await element.addEventListener('click', async (event) => {
   console.log('Element clicked!');
-
-  // Do something with the event
-  const target = event.target;
 }, { capture: false });
 
 // Remove event listener using the ID
 await element.removeEventListener('click', listenerId);
+```
+
+To listen document-wide, use the `SafeDocument` from `getRootDocument()`:
+
+```javascript
+const rootDoc = await Risuai.getRootDocument();
+const id = await rootDoc.addEventListener('keydown', async (event) => {
+  console.log('Key pressed anywhere!');
+});
 ```
 
 **Allowed Events (Unlimited):**
--- a/src/ts/plugins/apiV3/risuai.d.ts
+++ b/src/ts/plugins/apiV3/risuai.d.ts
@@ -808,6 +808,7 @@ interface SafeElement {
      * - Keyboard: keydown, keyup, keypress
      *
      * listener function receives trimmed event object with common properties only.
+     * listeners receive events that reach the element via normal DOM propagation; for document-wide listening, use the SafeDocument from getRootDocument().
      * 
      * @example
      * ```typescript
@@ -827,9 +828,9 @@ interface SafeElement {
 
     /**
      * Removes an event listener using its ID
-     * @param type - Event type
+     * @param type - Event type (for compatibility; removal is resolved by ID)
      * @param id - Listener ID returned by addEventListener
-     * @param options - Event listener options
+     * @param options - Event listener options (for compatibility)
      */
     removeEventListener(
         type: string,
--- a/src/ts/plugins/apiV3/v3.svelte.ts
+++ b/src/ts/plugins/apiV3/v3.svelte.ts
@@ -360,6 +360,8 @@ class SafeElement {
     }
 
     public removeEventListener(type:string, id:string, options?: boolean | EventListenerOptions) {
+        // type/options are unused, but are kept in the signature for API compatibility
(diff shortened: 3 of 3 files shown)


cubicj (commented):
I went through 9d21f229e and the implementation does what the docs describe. Listeners land on the wrapped element, `SafeDocument` keeps the `Document` target so viewport `scroll` still arrives, and replaying the stored registration on removal closes the mismatched capture case that used to leave a live listener behind. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

Where I stop is compatibility. `SafeElement.addEventListener` has attached to `document` since 680903d7b, I have no way to tell how many plugins rely on that, and whoever installed one of them sees it change after an update with nothing pointing at the cause. I know the `SafeDocument` docs reserve breaking restrictions for user safety, but this repo has kept old behavior reachable or shipped a migration when fixing something that was built wrong, and a docs note alone does not get there.

Personally I would keep `addEventListener` on its current attachment, mark it deprecated, and add an element-scoped method next to it. The tracked registration already carries its own `eventTarget`, so the removal and reload cleanup fixes apply to both paths as they are. How does that sound to you?

cubicj (approved):
Sorry, my last comment had the baseline the wrong way round. After your reply I went looking for plugins that actually built on the `document` attachment. Among the public V3 plugins I could find, the element listeners that meet app-wide events guard themselves with a hit-test against the element's rect or register on `body` to begin with, so their authors had already noticed the delivery and worked around it. The one dependency I found, a drag with `pointermove` and `pointerup` registered on the draggable element, sits next to a `scroll` listener on the chat container that has never fired on main because the event does not bubble, and another plugin's `pointerleave` handler is in the same position. That, plus the keystroke exposure staying open for as long as the old path exists, means you were right to push back on the deprecation route.

I re-ran the checks on 9d21f229e, including a happy-dom pass over the element scoping, the `Document` target for viewport `scroll`, and the mismatched-type removal case. `pnpm check`, `pnpm test`, and `pnpm build` are clean on this head.

Just one thing to flag, not blocking. A drag written that way has no pointer capture to fall back on, so a short note under the document-wide example that `pointermove` and `pointerup` belong on `body` or the `SafeDocument` would point that class of plugin at the fix.

Sorry again for the deprecation ask, that was a bad call on my part and it cost you a round.

